PR #249, opened 2013-01-18
fixing support for Unix sockets in native binding (rebased)

booo:
Binding natively connections to Unix sockets failed due to DNS lookups applied on pathname to Unix socket's folder.

I rebased the commit to the master branch. You should be able to merge this now.

Belongs to https://github.com/brianc/node-postgres/pull/179

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -75,15 +75,19 @@ var getLibpgConString = function(config, callback) {
       params.push("dbname='" + config.database + "'");
     }
     if(config.host) {
-      if(config.host != 'localhost' && config.host != '127.0.0.1') {
-        //do dns lookup
-        return require('dns').lookup(config.host, function(err, address) {
-          if(err) return callback(err, null);
-          params.push("hostaddr="+address)
-          callback(null, params.join(" "))
-        })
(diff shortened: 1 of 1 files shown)


redaktor:
@booo [and cc. brianc] This looks anything else than "good" - sorry, but the code will never be fired:
```js
if (!config.host.indexOf("/"))
```

You can't compare an indexOf with a simple **'!'** -
If the slash is not found indexOf returns `` -1 ```which is a negative integer but not false.
You need an additional tilde symbol !!!

I will do a pull request to fix it in 2-10 minutes ;)
```js
if (!~config.host.indexOf("/"))
```

Meanwhile @booo could read http://stackoverflow.com/questions/18347033/how-to-shorten-my-conditional-statements/18347047#answer-18347047

brianc:
Oh nice catch! sorry 'bout that

redaktor:
Sorry, 
new to this: It's a minor issue for me now:
I overlooked the ```pg.connect``` *object* notation doc.
Seems more elegant to me anyway and everything is fine when I connect to 
```js
{
  host: '/foos_uberspacesocket'
}
```

PR #353, opened 2013-05-20
Handle bigint as string to prevent precision loss

sevastos:
TIL dealing with bigints in javascript is a bit of a pain.

I had a weird issue of inserting correctly bigints on the db but getting off by a bit numbers on the response.

Apparently all the numbers in javascript are floating-point numbers<sup>[1](http://www.irt.org/script/1031.htm)</sup> and the accuracy is only assured for integers ±2<sup>53</sup> between -9007199254740992 and 9007199254740992. 
However bigint ranges ±2<sup>63</sup>: -9223372036854775808 to 9223372036854775807 <sup>[2](http://www.postgresql.org/docs/9.2/static/datatype-numeric.html)</sup>.

The most common solution for handling bigint data on javascript is carry them with strings and use special libs for math operations. So strings it is.

However, I wonder if it may cause any troubles to anyone that is treating bigint as number on javascript-level and do some checks around that. Though they shouldn't since that only works till the ±2<sup>53</sup> range and is not the right way.

Anyway let me know what you think.
Thanks

### DIFF
--- a/lib/types/binaryParsers.js
+++ b/lib/types/binaryParsers.js
@@ -1,4 +1,5 @@
 var ref = require('ref');
+var endian = (ref.endianness === 'LE') ? 'BE' : 'LE';
 
 var parseBits = function(data, bits, offset, invert, callback) {
   offset = offset || 0;
@@ -108,7 +109,7 @@ var parseInt32 = function(value) {
 };
 
 var parseInt64 = function(value) {
-  return String(ref.readInt64BE(value, 0));
+  return String(ref['readInt64' + endian](value, 0));
 };
 
 var parseFloat32 = function(value) {
--- a/lib/types/textParsers.js
+++ b/lib/types/textParsers.js
@@ -169,10 +169,13 @@ var parseBigInteger = function(val) {
 };
 
 var init = function(register) {
-  register(20, parseBigInteger);
-  register(21, parseInteger);
-  register(23, parseInteger);
-  register(26, parseInteger);
+  register(20, parseBigInteger); // int8
+  register(21, parseInteger); // int2
+  register(23, parseInteger); // int4
+  register(26, parseInteger); // oid
+  register(700, parseFloat); // float4/real
+  register(701, parseFloat); // float8/double
+  //register(1700, parseString); // numeric/decimal
   register(16, parseBool);
   register(1082, parseDate); // date
   register(1114, parseDate); // timestamp without timezone
--- a/test/integration/client/type-coercion-tests.js
+++ b/test/integration/client/type-coercion-tests.js
@@ -23,7 +23,9 @@ var testForTypeCoercion = function(type){
           });
 
           assert.emits(query, 'row', function(row) {
-            assert.strictEqual(row.col, val, "expected " + type.name + " of " + val + " but got " + row.col);
+            var expected = val + " (" + typeof val + ")";
+            var returned = row.col + " (" + typeof row.col + ")";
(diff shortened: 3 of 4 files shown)


mariusa:
Would it be possible to please add a preference/setting, so that all bigint are parsed with parseInt(), knowing there will be data loss on really big numbers?

We constantly find bugs because in some place we forget data is transformed to string.

Thanks!

eric-brechemier:
> Apparently all the numbers in javascript are floating-point numbers1 and the accuracy is only assured for integers ±2<sup>53</sup> between -9007199254740992 and 9007199254740992.
However bigint ranges ±2<sup>63</sup>: -9223372036854775808 to 9223372036854775807 2.
>
> The most common solution for handling bigint data on javascript is carry them with strings and use special libs for math operations. So strings it is.

Why not keep BIGINT values as numbers up to ±2<sup>53</sup>, and only use strings for values between ±2<sup>53</sup> and ±2<sup>63</sup>?

This would be less confusing in most cases, and only confusing when truly required.

rpedela:
@mariusa Personally, I am not opposed to a config option. In the meantime, you can override the default type parser.
```js
var pg = require('pg');

// bigint
pg.types.setTypeParser(20, function (value) {
    return parseInt(value);
});

// numeric
pg.types.setTypeParser(1700, function (value) {
    return parseFloat(value);
});
```

mariusa:
Thanks, that works!

We'll trust the magic "20" number :) (saw the internal mapping, hope that won't change)

rpedela:
That magic number comes from ```SELECT oid FROM pg_type WHERE typname = 'int8';``` and is unlikely to change.

1mike12:
I guess this still hasn't really been addressed after all this time?

Here's some copy pasta based on the discussion above that finally gets count operations return as numbers. The avg operations for whatever reason doesn't return strings, so I didn't have to do the same for type `1700`

```ts
// 20 refers to BigInt, pg by default returns count subqueries as strings instead of
// numbers. We override that behavior so counts less than 2^53 
// (IE practically anything we would ever do) are returned as numbers
pg.types.setTypeParser(20, function(value) {
  if (isStringASafeInteger(value)) {
    return parseInt(value)
  } else {
    return value
  }
})

/**
 * In JavaScript, numbers are represented using the IEEE 754 double-precision floating-point format, also known as "double." This format uses 64 bits to store a number: 1 bit for the sign, 11 bits for the exponent, and 52 bits for the fraction (or mantissa), which results in JavaScript having a single Number type for both integer and floating-point values.
 *
 * Although numbers larger than 2^53 - 1 or smaller than -2^53 + 1 can be loosely represented, they are not safe integers and there can be rounding errors after operations push outside the safe bounds
 * @param str
 */
export function isStringASafeInteger(str: string) {
  const num = Number(str)
  if (isNaN(num)) return false
  return Number.isSafeInteger(num)
}

```

charmander:
@1mike12 The issue of precision loss has been addressed by this pull request. The suggestion I guess you’re referring to hasn’t been implemented because quietly changing representations based on a number’s value is a *really bad idea*. Expressions like `n + 1` will appear to work correctly until they suddenly don’t. (Even the current implementation where the JavaScript type only depends on the PostgreSQL type is a bit of a trap.)

Anyway, JavaScript has the much more efficient and better-behaved `BigInt` type now, which can be used consistently and solves the problem elegantly. The only issue is breaking backward compatibility, but it should happen in an upcoming major version.

PR #370, opened 2013-06-06
Makes client_encoding configurable and optional

badave:
Makes client_encoding configurable and optional


airzebeth:
Is it normal that even with the client_encoding set to "SQL_ASCII", I can't get to write in binary in my pg database ?

PR #456, opened 2013-10-17
Parse arrays: json[], uuid[]

albert-lacki:
I've added parsing of two additional types json[] and uuid[]. I'll be happy if you could merge it.

### DIFF
--- a/lib/types/textParsers.js
+++ b/lib/types/textParsers.js
@@ -180,11 +180,11 @@ var parseBigInteger = function(val) {
 
 var parseJsonArray = function(val) {
   var arr = parseStringArray(val);
-  
+
   if (!arr) {
     return arr;
   }
-  
+
   return arr.map(function(el) { return JSON.parse(el); });
 };
 


demisx:
Guys, I have a `jsonb[]`type column `photos` in my Postgres 9.4 table. My ORM is Bookshelf that uses `pg` as a client. When I fetch records the API responds with the array in this format:

```js
{"{\"url\": \"https://example.com/photo1.jpg\"}","{\"url\": \"https://example.com/photo2.jpg\"}"}
```

As you can see, it is still in Postgres specific format. Should the conversion to a conventional JSON array (e.g. [{...},{...}] be handled by the `pg` parser or am I wrong? I am posting under this issue, since I think it may be related to this parser. Please advise. Thank you.

brianc:
I think it should probably convert to a json structure, but I'm not sure
what additional bits of encoding are housed within `JSONB` that aren't
supported by a standard `JSON.parse` style call?  Is there a good `JSONB`
library for node we could add to [node-pg-types](
https://github.com/brianc/node-pg-types)

demisx:
Thank you for the prompt response. I believe this issue applies to both, JSON[ ] and JSONB[ ] array parsing. I am not aware of any good libraries for this and googling didn't reveal much besides just some Ruby and C libs like this one: https://code.google.com/p/pg-to-json-serializer/. Not sure if I was looking for the right stuff, though, since this is not yet my area of expertise.

Shall I create a new issue to track this effort?

bendrucker:
There's no issue with parsing jsonb with `JSON.parse` that I know of. The main upside is in storing data in a binary, typed format as opposed to just a dumb string. 

Opened https://github.com/brianc/node-pg-types/pull/21

demisx:
@bendrucker You are correct. I don't see any issues with parsing if I define the column as `json[]` instead of `jsonb[]`. My previous assumption about both was incorrect.

bendrucker:
If you want a good blog post from the guy who reviewed jsonb, check this out:

http://pgeoghegan.blogspot.com/2014/03/what-i-think-of-jsonb.html

PR #497, opened 2014-01-05
application_name

hoegaarden:
Support for `application_name` and `fallback_application_name`


ismyrnow:
Just added documentation for these to the [wiki](https://github.com/brianc/node-postgres/wiki/Client#parameters).

Never knew these options were available, but it makes monitoring connections much easier.

brianc:
thanks for adding the documentation!  I totally agree with you btw - makes figuring out who's connected much easier.

abenhamdine:
@ismyrnow Very usefull ! Thx a lot.

Issue #510, opened 2014-01-22
Question about the `date` type


jrf0110:
The Postgres `date` data type has no information about timezone. That wouldn't make any sense. However, the dateParser in node-postgres defaults to Greenwich timezone due to the way it's instantiating the object.

According to [this MDN article](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse), when creating a new Date, the constructor uses the `Date.parse` which, if passed a string without a timezone, will default to the systems timezone. This is not the behavior I'm seeing in the node.

Consider the following on my system in GMT-0600

``` javascript
new Date('2014-01-24').getDate() // => 23
new Date('2014-01-24 00:00:00').getDate() // => 24
new Date( 2014, 0, 24 ).getDate() // => 24
```

It's my feeling that when the data type is `date` then we should use one of the latter two methods so that it actually defaults to system timezone. The first method is being employed by node-postgres right now:

https://github.com/brianc/node-postgres/blob/master/lib/types/textParsers.js#L17

Thoughts?

PR #514, opened 2014-01-26
Parse date type as local time

benighted:
Resolves issue #510, but is a breaking change.  Handle with care :smile: 
Updated failing test case as well.

### DIFF
--- a/lib/types/textParsers.js
+++ b/lib/types/textParsers.js
@@ -14,7 +14,8 @@ var parseDate = function(isoDate) {
       return null;
     } else {
       //it is a date in YYYY-MM-DD format
-      return new Date(isoDate);
+      //add time portion to force js to parse as local time
+      return new Date(isoDate + ' 00:00:00');
     }
   }
   var isBC = /BC$/.test(isoDate);
--- a/test/integration/client/type-coercion-tests.js
+++ b/test/integration/client/type-coercion-tests.js
@@ -202,13 +202,14 @@ helper.pg.connect(helper.config, assert.calls(function(err, client, done) {
 if(!helper.config.binary) {
   test("postgres date type", function() {
     var client = helper.client();
+    var testDate = new Date (2010, 9, 31);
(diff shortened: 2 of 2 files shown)


alejandrolsca:
@brianc is there a way to set the timezone in node-postgres? I tried to find it in the documentation without success, i found an environment variable PGTZ here https://www.postgresql.org/docs/current/static/libpq-envars.html but i had not success trying to guess the configuration in the pool connection i.e.
`const pool = new Pool({
        connectionString: '......',
         timezone: 'America/Chihuahua'
    })`
also tried to set the timezone with a query as follows
`await client.query("SET timezone = 'America/Chihuahua';")`
but didn't worked
in fact what im trying to do is to set the timezone dynamically, so the query will be at the end as follows
`await client.query("SET timezone = $1;")`

can you give me start point to solve this issue, maybe im not looking in the right place. Thanks in advance for your support

jrf0110:
@alejandrolsca generally you'd use something like `at time zone $1` in your queries:

```sql
-- Display the timestamp in the timezone of the object
select
  orders.created_at at time zone orders.timezone
from orders
```

alejandrolsca:
@jrf0110 I was aware of that way, so its not recommended to use SET timezone.....? at this moment looks like a good solution, but in the near future could i have some problems? scalability issues? just wonder, also still wondering why SET timezone.... didnt work, even if you call that query in node postgres its override by postgres.conf?

jrf0110:
I believe `SET timezone` will simply set the timezone the database is in. Try something like:

```sql
set timezone = 'America/Chicago';
select now();
```

alejandrolsca:
@jrf0110 well the problem here is that it doesn't work in node-postgres, at least not for me, if I execute those queries in DBMS it works, but not by doing it with node-postgres library as follows:
`client.query("SET timezone = 'America/Chihuahua';")
  client.query("My query with dates here")`

PR #704, opened 2014-12-21
Use svg version of travis build status badge

eiriksm:
Hey!

Just a minor PR here. I noticed you are using the png version of the travis build status badge. They actually also offer a svg version which is nicer looking, especially on retina screens.

So if you want, this PR changes the badge to the svg version instead of png.

Thanks for making this available and have a great day!


brianc:
Ahh thanks! I dig it!

Issue #703, opened 2014-12-21
Have a way of initializing new connections in pool


monsanto:
Often I need to initialize connections before use (setting session variables, creating temporary views, etc). I don't see any way to do this with pooling; pool.connect() does not appear to have a way of communicating whether the connection returned is new or reused.

node-mysql solves this problem by having a 'connection' event on pools: https://github.com/felixge/node-mysql#connection

ss108:
@brianc I'm a fairly green programmer who's interested in taking a shot at it if you wanna walk me through it (well I'm gonna take a shot at it anyways lol).

brianc:
I would love to help you take a shot. Want to gchat me? I can walk you
through

ss108:
Awesome thanks! just sent you chat invite

ss108:
@brianc Submitted a PR, please let me know if I'm on the right track. If I am, I will run the tests. (And put something in or remove the empty `if` statement when the pool acquires a connection)

brianc:
This should become a lot easier in `pg@7.0` - I'm upgrading the pool & will expose it's factory methods in a way that makes builder customization easier.

PR #711, opened 2015-01-10
emit a connection event when a pool creates a connection

ss108:
re: https://github.com/brianc/node-postgres/issues/703

### DIFF
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -48,8 +48,9 @@ var pools = {
             }
           });
           client.poolCount = 0;
-          //Should return client object in event emission?
-          pool.emit('connection');
+
+          //Announce the new connection.
+          pool.emit('connection', client);
 
           return cb(null, client);
         });
@@ -76,11 +77,6 @@ var pools = {
           cb = domain.bind(cb);
         }
         if(err)  return cb(err, null, function() {/*NOOP*/});
-
-        if(client.poolCount === 0){
-            //If one wanted to know specifically whether the connection a pool is using
-            //is new or not, the event would go here.
-        }
 
         client.poolCount++;
         cb(null, client, function(err) {
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -48,6 +48,9 @@ var pools = {
             }
           });
           client.poolCount = 0;
+          //Should return client object in event emission?
+          pool.emit('connection');
+
           return cb(null, client);
         });
       },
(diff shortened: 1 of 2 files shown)


lib/pool.js
```diff
@@ -48,6 +48,9 @@ var pools = {
             }
           });
           client.poolCount = 0;
+          //Should return client object in event emission?
+          pool.emit('connection');
```
brianc: I think it would be good to emit the client as well, yes.

lib/pool.js
```diff
@@ -73,6 +76,12 @@ var pools = {
           cb = domain.bind(cb);
         }
         if(err)  return cb(err, null, function() {/*NOOP*/});
+
+        if(client.poolCount === 0){
```
brianc: Can we remove this? The pool only initializes a client once, so it actually will only emit the `connection` event the first time a connection is constructed. I think this wont be needed.

brianc:
@ss108 looks great so far!  I added a few small comments.  Also, and importantly, would you be able to write a couple tests for this behavior?

The tests should go in `test/integration/connection-pool` in a new file.  The file should be executable by just calling `node test/integration/connection-pool/your-file.js`.  There are some helpers to parse connection parameters and stuff, so you should use something like [this file](https://github.com/brianc/node-postgres/blob/master/test/integration/connection-pool/idle-timeout-tests.js) as a reference.

ss108:
Cool thanks

I will write those tests tmrw

ss108:
@brianc I have a couple questions--sorry I know this is pretty simple

I was looking at integration/connection/notification-tests.js as a guide, and saw that when testing for this behavior we run into the issue of not having direct access to the pool (to be able to call assert.emits on the pool). I can't just call helper.pg.connect and pass a callback to help out either, because the callback would get the newly acquired client as an argument and not the pool itself. Any advice?

I will poke around later day, but just wanted to throw this out there from now in case I was thinking about this in an obviously incorrect or overly complicated manner

Also, I noticed that there are other places in the code (ex: Connection.prototype.connect), where an event 'connected' is emitted; does it not matter that they have the same name (does Node have like a behind-the-scenes signature for events that includes both the event's name and the object doing the emitting, or something? or is because the events are not on the same 'scope' for lack of a better term)

ss108:
Hey @brianc could you take a look at the test and provide a little insight; the 'connect' event is never firing, which means the `create`  function is never getting called, which means the route I am taking to get it to execute (calling a pool's `connect` method directly) is incorrect. I wanted to call `create` directly, but this function is not exposed on pools returned via `helper.pg.pools.getOrCreate` (I guess it's intentionally private right?). 

How would you go about testing this?

Sorry the delay and stuff

kesavkolla:
Please merge this into main.
+1 vote

brianc:
hey @kesavkolla - I've split the pool out into its own module & added a `connect` event to the pool based on your initial designs: https://github.com/brianc/node-pg-pool/blob/master/README.md#connect

If you're running the most recent version of this module the most recent version of pg-pool should be installed by default & you should be getting this behavior!

Sorry it took me a while to finally get to this PR.  I really appreciate the thought & work you put in here & hope the implementation meets your needs. ❤️

Issue #701, opened 2014-12-15
New error fields in PostgreSQL v9.3


moll:
According to http://www.postgresql.org/message-id/E1U0JMj-0001zn-Sv@gemulon.postgresql.org PostgreSQL v9.3 got a few new error fields in January 2013. I don't believe these are represented in the error node-pg returns: https://github.com/brianc/node-postgres/blob/master/lib/connection.js#L523.

Particularly the table, table column, data type, or constraint fields.
The full list: http://www.postgresql.org/docs/9.3/static/protocol-error-fields.html

charmander:
Fixed in #712.

PR #712, opened 2015-01-10
#701 Expose error fields added in PG 9.3.

rpedela:
Several error fields were added in PG 9.3:
* schema
* table
* column
* data type
* constraint

This PR exposes them in the returned Error object.


brianc:
@rpedela thank you so much!  Quick question...is there any way you could write a test for this?  I know it's postgres version dependent...you could use [this function](https://github.com/brianc/node-postgres/blob/ac0d1c71c2a29fff42aac42d70e87d95ba1b595f/test/integration/test-helper.js#L17) to only run the test if the database version supports it.  Here's an [example](https://github.com/brianc/node-postgres/blob/ac0d1c71c2a29fff42aac42d70e87d95ba1b595f/test/integration/client/json-type-parsing-tests.js#L12)

rpedela:
Tests added. I am not sure how to trigger the "data type" error field, but the others are tested. I am happy to add another test for the "data type" error field if someone knows how to trigger it.

brianc:
@rpedela awesome! :clap: :dancer:  

Will push a new minor version with this change later today. :+1:

rpedela:
Thanks! :)

PR #714, opened 2015-01-14
added the option to connect without specifying a database (noDatabase option)

eventEmitter:
Added the `noDatabase` option to the connection configuration object. It makes it possible to connect to a postgres server without specifying which database to use. This doesn't change the existing API and should not introduce any issues with backward compatibility.

I am working on a db migration tool for an orm and it would be very useful for the automated setup and testing of databases to be able to create a database without the need to use an existing one.

usage:

````
var client = new pg.Client({
      user: 'someUser'
    , password: 'somePassword'
    , host: 'localhost'
    , noDatabase: true
});
````

See also #437

### DIFF
--- a/lib/connection-parameters.js
+++ b/lib/connection-parameters.js
@@ -12,7 +12,9 @@ var val = function(key, config, envVar) {
     envVar = process.env[ envVar ];
   }
 
-  return config[key] ||
+  // let the user connect without selecting a database
+  if (key === 'database' && config.noDatabase) return '';
+  else return config[key] ||
     envVar ||
     defaults[key];
 };

### CODE REVIEW

**Comment by eventEmitter** [comment]
maybe a `noDefaults` option on the configuration object would be a better option? If set, no default values and no environment variables are used.

**Comment by brianc** [comment]
I definitely prefer a `noDatabase` option to the `noDefaults` option on `pg.defaults`. The `pg.defaults` thing is actually kinda nasty & I wish I had not done that - it's effectively global state and makes things hard to test & maintain.   One question...for your use case, is there a way you can get by with the current code? something like...

```js
pg.defaults.database = ''
process.env.PGDATABASE = ''
pg.connect(function(err, client, done) {
 ...
})

```

Or does that still try to connect to some database?

eventEmitter:
`pg.defaults.database = '';` works. As long as i can set the default config to empty it's ok for me.

thanks!

brianc:
Gonna close this in favor of just blanking out the database on the defaults.

PR #719, opened 2015-01-27
Add test for using intervals as parameter in prepared statements

marekventur:
Currently node-postgres returns intervals in the form of ```{days: 12, hours: 2}```. Passing the same object back into a prepared statement causes strange behaviour and is normally not correct.

This just adds a test that will fail. The actual fix is here: https://github.com/brianc/node-pg-types/pull/16

### DIFF
--- a/test/integration/client/prepared-statement-tests.js
+++ b/test/integration/client/prepared-statement-tests.js
@@ -17,6 +17,32 @@ test("simple, unnamed prepared statement", function(){
   });
 });
 
+test("use interval in prepared statement", function(){
+  var client = helper.client();
+
+  client.query('SELECT interval \'15 days 2 months 3 years 6:12:05\' as interval', assert.success(function(result) {
+      var interval = result.rows[0].interval;
+
+      var query = client.query({
+        text: 'select cast($1 as interval) as interval',
+        values: [interval]
+      });
+
+      assert.emits(query, 'row', function(row) {
+        assert.equal(row.interval.seconds, 5);
+        assert.equal(row.interval.minutes, 12);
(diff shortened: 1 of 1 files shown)


brianc:
Hey @marekventur -

I'm really sorry about taking so long on this. :grimacing: I've started a new job and was kinda going full blast on it and not saving anytime for OSS for a bit.  Back now! :dancer: 

I'll go ahead & merge that change into `node-pg-types`.  Once that's integrated, want to bump your version you use of `node-pg-types` on your branch so the tests pass?  I actually think it might be best to _relax_ the versioning requirement for `node-pg-types` to just require a particular _major_ version, letting the minor and patch versions increment naturally.  When I'm using other people's modules I generally pin to the __exact__ semver because I've been hit by breaking changes in minor and patch versions of other peoples modules, but since we control both `node-postgres` and `node-pg-types` we know we wont put breaking changes into a patch or minor version over there! How does that sound?

brianc:
Actually since I kept you waiting so long I'll just merge this & do the rest myself! Sorry again for that.

marekventur:
Hi! 

I hope you had a good start at your job! Sorry for keeping you waiting too, I was on holiday without good internet connection. Your idea looks good to me, mildly relaxing the version requirements for bits you control sounds sensible too. Since you've already done the merge there's nothing left for me to do, thanks :) Now I can update our app to the newest version of node-postgres and remove some dirty work-around code that's no longer necessary.

Marek

brianc:
Awesome!  Please lemme know if it doesn't work or you hit any snags! :)

PR #729, opened 2015-02-15
In order for the command line execution to finish

wallymathieu:
In the example, it looks like it's missing a client end.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ pg.connect(conString, function(err, client, done) {
     }
     console.log(result.rows[0].number);
     //output: 1
+    client.end();
   });
 });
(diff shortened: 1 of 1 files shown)


brianc:
Thanks! Sorry for the delay in getting to this - been buried recently.

wallymathieu:
No problem!

dogles:
Is this correct? It seems like this would destroy the connection, which is the whole point of the connection pooling example...

wallymathieu:
Sounds like I browsed the code and didn't read the comment. As the code reads, it reads like a command line app. Perhaps an express site that answers with some info would be clearer?

PR #734, opened 2015-02-21
Fix result.rowCount being a string instead of a number via native bindings.

MayhemYDG:
Btw, once you drop support for node 0.8 you can drop the `parseInt` radix, as it should default to 10.

### DIFF
--- a/test/native/callback-api-tests.js
+++ b/test/native/callback-api-tests.js
@@ -8,6 +8,7 @@ test('fires callback with results', function() {
   client.query('SELECT 1 as num', assert.calls(function(err, result) {
     assert.isNull(err);
     assert.equal(result.rows[0].num, 1);
+    assert.equal(result.rowCount, 1);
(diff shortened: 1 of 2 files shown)


test/native/callback-api-tests.js
```diff
@@ -8,6 +8,7 @@ test('fires callback with results', function() {
   client.query('SELECT 1 as num', assert.calls(function(err, result) {
     assert.isNull(err);
     assert.equal(result.rows[0].num, 1);
+    assert.equal(result.rowCount, 1);
```
brianc: I think you _might_ want to use `assert.strictEquals` here - doing an `assert.equals` basically does ` "1" == 1` which is true. What do you think?

brianc:
Other than that one little thing, this looks great!

MayhemYDG:
You're right, I missed `strictEqual` and thought `equal` was using the strict equality operator.

brianc:
No worries! This is fabulous now! Thank you so much!! :ok_woman: :bow:

PR #735, opened 2015-02-22
Relax semver requirement of pg-types

brianc:


brianc:
This substantially reduces the version requirements from an absolute pinned version to allowing it to float against the most recent `pg-types@1.x` - this should allow `pg-types` to pull in more and more functionality without requiring an explicit bump in versions of this module. This allows people who can't or wont upgrade pg to benefit from type parsing improvements. @jrf0110 @tlhunter this one's for you. :dart:

jrf0110:
ballllllin' thanks!

PR #747, opened 2015-03-11
Correcting the library description

vitaly-t:
The library moved away from featuring the Promise library, it now supports the entire Promises/A+ family.

Happy Days! :)


brianc:
nice! :+1:

PR #756, opened 2015-03-30
Fix for issue 750

ngervasi:
The arguments array is now duplicated (using slice() that should be the faster way to do it in node.js) to avoid changing the given object and cause unexpected behaviour in external code


brianc:
Whoah @ngervasi thanks for your work on this!

ngervasi:
Brian, I'm having a little trouble with testing, as you probably have already figured out... :)
My pull request is not yet correct, I have a new one ready but I wanted to have the tests fail on your build before applying my changes. Can you give me a little help? What's wrong with this test? It should fail in your build:

```
var helper = require(__dirname + '/test-helper');
var pg = helper.pg;
var config = helper.config;

test('check that passed values types have not been changed during the query phase', function() {

	pg.connect(config, assert.success(function(client, done) {
		var originalValues = [1,2];
		var values = originalValues.slice();

		client.query('SELECT 1 WHERE 0 <> $1 AND 0 <> $2',values, assert.success(function(result) {
			console.log('result:',result);
			assert.equal(result.rowCount, 1);
			assert.equal(values.length,originalValues.length,'expecting same length as given array!');
			assert.strictEqual(isNaN(values[0]),false,'expecting a number!');
			assert.strictEqual(isNaN(values[1]),false,'expecting a number!');
			done();
		}));

	}));
});
```

brianc:
Closing this because it's quite old - if the issue still stands I'm happy to discuss more.

PR #761, opened 2015-04-04
Add node/PG version matrix to Travis-CI config

sehrope:
This PR updates the .travis.yml config file to test against various permutations of node and PostgreSQL versions. I've also removed an unused shell script for setting up PG 9.2. It wasn't referenced anywhere and Travis takes care of setting up the PG environments via the matrix config.

Previously the Travis config was only testing against node v0.10 and v0.11.

With this PR it now tests against node versions v0.10, v0.11, v0.12, and io.js v1, and PostgreSQL versions 9.1, 9.2, 9.3, and 9.4 for a total of 4 x 4 = 16 test environments (plus 1 more for whatever the defaults are).


sehrope:
Looks like there is some kind of transient error running the tests for v0.11/9.3 on Travis: https://travis-ci.org/brianc/node-postgres/jobs/57163323#L206

I suggest running that build again as it worked fine when I ran it (as well as the 16 other env combinations).

sehrope:
I've updated this PR to remove node v0.11, bump the iojs version to iojs-v2 and switch to container based builds in Travis (*by disabling sudo in `.travis.yml`*).

The switch to container based testing speeds up the automated tests quite a bit. It's enough that running the full matrix with multiple node versions (0.10, 0.12, and iojs-v2) and PostgreSQL server versions (9.1 through 9.4) takes the same amount of time as the existing tests with the non-container setup.

brianc:
This is amazing!  Thank you so much!!

PR #765, opened 2015-04-15
Updated Test Environments

vitaly-t:
* Replaced NodeJS 0.11 with 0.12, because people only care about 0.10 and 0.12 today;
* Added test support for the most recent stable version of IO.js


brianc:
:+1: thanks!

PR #772, opened 2015-05-04
Allow prepared statement names that are longer than 63 characters

bdowning:
Postgres truncates prepared statement names at 63 characters.  If you have two that are the same in the first 63 characters you'll get a "prepared statement already exists" error.  This is very noticeable if you try to use the SQL query itself as the name.

This fixes this by creating and caching unique short names for all long names, and sending the short one to Postgres.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -19,6 +19,8 @@ var Connection = function(config) {
   this.parsedStatements = {};
   this.writer = new Writer();
   this.ssl = config.ssl || false;
+  this.longNameI = 0;
+  this.longNameMap = {};
   this._ending = false;
   this._mode = TEXT_MODE;
   this._emitMessage = false;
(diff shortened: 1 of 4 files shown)


brianc:
First of all, sorry for taking a while to get to this...I really appreciate the thought & work you put into this, but I think this might be better left out of node-postgres itself.  I think it introduces some cleverness in the library I could see being a footgun if someone wasn't expecting it to work this way, and if postgres introduces a longer name cap in the future we'd have to come back and figure out a way to support both the old constraint and new constraint.  I think something that *would* be helpful though is spitting a warning out to the console if there is a query name of greater than 63 characters.  Because we don't _automatically_ attach a name & use a named query anyone who's using this feature is doing so in their code explicitly.  I think it'd be pretty easy for them to do an `md5` on the query text or something else - like what you've done - to give their queries unique names.  It might even be helpful to make a node module that generates unique query names for you and then it could be shared for anyone who's using that.  What do you think?

bdowning:
I'd be fine with doing it on my own for my use-case, I have an abstraction layer to run my own queries anyway.  Really what I was trying to fix was the WTF-factor of having two different queries with different names silently resolve to the same thing.  A warning would suffice to fix that.

brianc:
Yeah I definitely feel you that the WTF-ness on that is no bueno at all.  I'll fix that right up.

brianc:
How does this look?

https://github.com/brianc/node-postgres/pull/787

bdowning:
Closing in favor of #787 which is a lot less invasive.  Thanks.

Issue #757, opened 2015-04-02
Support for multiple result objects


irakliy81:
It would be really nice if after sending a single SQL request with 2 or more SQL statements separated by semi-colons, the query callback contained an array of result objects rather than a single object. Currently, something like this returns in a single result object with rows from both SQL statements present in result.rows

```
SELECT * FROM users WHERE id = 1;
SELECT * FROM conversations WHERE user_id = 1;
```

Thanks!

vitaly-t:
PG library always deals with one query at a time. What you are asking for is only possible through a [promise-wrapped transaction](https://github.com/vitaly-t/pg-promise#transactions).

irakliy81:
Thank you for the reply, Vitaly. What I meant was a bit different - rather than sending two separate requests, I am sending a single request with two statements separated by commas (this is already working). In my benchmarking, a single request with two semicolon-seprated statements completes 2x faster than two separate requests.

The only issue I have is parsing the results. Right now, there is a single result object - and when I make a request with two semicolon-separated statements, this single result object contains rows from both requests mixed together. I can parse through them to figure out which row belongs to which statement, but it would be much cleaner if this was already handled by the library (especially since the library maintains correct column names for each row already).

vitaly-t:
@irakliy81, This is not a standard way to do this. If you need to execute more than query at a time, you should use transactions, because this is exactly what they are for. And a transaction will execute at the same speed, and return you properly formatted results from two queries, not mixed ones.

Here's an example with the same library:

```javascript
db.tx(function (ctx) {
    return promise.all([
        ctx.query("SELECT * FROM users WHERE id = 1"),
        ctx.query("SELECT * FROM conversations WHERE user_id = 1")
    ]);
})
    .then(function (data) {
        console.log(data[0]); // print result from the first query;
        console.log(data[1]); // print result from the second query;
    }, function (reason) {
        console.log(reason); // print why failed;
    });
```

irakliy81:
I see. So two SELECT queries executed in a single transaction should run as fast as a single query with 2 semicolon-separated statements? I'll try to test that, and if so - i think the approach that you have will work very well. Thank you!

vitaly-t:
@irakliy81 
Have you made any progress on this?

irakliy81:
@vitaly-t - i haven't had a chance to get to this yet. one of the complications is that my application is written in typescript and there doesn't seem to be typescript definition files available for pg-promise library.

[...]

PR #776, opened 2015-05-06
feat(757) support returning an array of results

lukemurray:
Attempt at implementing #757.

Support for returning multiple results (array) if the query has more than one. You have to tell the query that you expect an array returned using an option. E.g.
client.query({text: query, rowMode: 'array', multiResult: true}, (err, results) => {
  // results is an Array of Result
});

I know this has no tests, could you point me in the right direction on creating tests for the project? Also this might be the completely wrong way to achieve this, let me know your thoughts, I'm no Postgres-anything expert.

Cheers

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -10,6 +10,7 @@ var Query = function(config, values, callback) {
 
   config = utils.normalizeQueryConfig(config, values, callback);
 
+  this.multiResult = config.multiResult;
   this.text = config.text;
   this.values = config.values;
   this.rows = config.rows;
@@ -23,7 +24,10 @@ var Query = function(config, values, callback) {
   if(process.domain && config.callback) {
     this.callback = process.domain.bind(config.callback);
   }
-  this._result = new Result(config.rowMode, config.types);
+  this.rowMode = config.rowMode;
+  // the active result in muti result mode
+  this._result = new Result(this.rowMode);
+  this._resultsArray = [];
   this.isPreparedStatement = false;
   this._canceledDueToError = false;
   EventEmitter.call(this);
@@ -70,6 +74,11 @@ Query.prototype.handleCommandComplete = function(msg, con) {
   if(this.isPreparedStatement) {
     con.sync();
   }
+  // new result for next query
+  if(this.multiResult) {
+    this._resultsArray.push(this._result);
+    this._result = new Result(this.rowMode);
+  }
 };
 
 Query.prototype.handleReadyForQuery = function() {
@@ -77,7 +86,7 @@ Query.prototype.handleReadyForQuery = function() {
     return this.handleError(this._canceledDueToError);
   }
   if(this.callback) {
-    this.callback(null, this._result);
+    this.callback(null, this.multiResult ? this._resultsArray : this._result);
   }
   this.emit('end', this._result);
 };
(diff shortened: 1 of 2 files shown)


brianc:
I like it!  Would you be able to include some unit tests for this?  I can't merge w/o tests!

lukemurray:
I could not figure out how to get the callback to trigger in a unit test using client.query(...) so I added a couple of simple integration tests.

Seems they run for native as well as non-native. I'm not sure how the native stuff fits in yet, I'll have to look at it tomorrow.

Also seems some unrelated tests failed that passed locally for me, one waiting for a callback that never happens.

lukemurray:
Hmm so I'm not sure how Native works, it seems to get all the results at once in native/query.js NativeQuery.prototype.submit

@brianc do think it will be possible with the way native works? Or can I tell some how that is is native mode and not return multiple results. Could you maybe point me in the right direction?

Thanks

vitaly-t:
Is this another happily forgotten merge? It was showing great potential for improving the library...

linpy:
I am out of the office until 2015/07/24.

I would on leave from 2015/1/22~1/25. Please drop me mail if anything.
Thanks.


Note: This is an automated response to your message  "Re: [node-postgres]
feat(757) support returning an array of results (#776)" sent on 20/07/2015
3:46:50.

This is the only notification you will receive while this person is away.

irakliy81:
It seems like lukemurray's solution works very well for JS code. The only change I would potentially make is to replace line 91 in the query.js like so:

```
from
---------------------------------
this.emit('end', this._result);

to
---------------------------------
this.emit('end', this.multiResult ? this._resultsArray : this._result);
```

Could this be merged into the master? I understand that the multiResult support would not work for Native bindings but I think it's better to have it for JS only rather than not to have it at all.

lukemurray:
I'd love it to be merged :) I haven't had anymore time to figure out the native side of things, but if @brianc is happy for just the JS side, that'll be awesome as I solved a problem I had with the library.

vitaly-t:
+ 1. This is definitely a must-have, since Postgres supports it.

yasonk:
This would a very helpful feature.  +1

nazar:
+1
Would really love for this to be merged :D

denshikov-vovan:
+1 yup, i need this feature too :)

apla:
Please make sure this statements will work before merging:

```sql
-- first one
select ';' as val; select ';;;' as val;

-- second one
do $$
begin
raise notice 'just notice; it is fine to ignore;';
end;
$$;
```

lukemurray:
both of those statements work. This change doesn't actually split your statement or anything, some underlying part was already streaming each result back.

abenhamdine:
@brianc this feature would be very useful, could you tell if something is still missing in the PR ?
Thx a lot.

maxcnunes:
I have tested it locally with some queries. It is working pretty well. @brianc could you please accept this pull request?

RyanMcDonald:
Would love for this feature to be merged. 👍

brianc:
This feature looks useful, but unfortunately this PR doesn't hit the nail on the head. 🔨 

Things missing
1) unit test failing
2) no support for the feature in `lib/native/query`

If someone would like to open up a new pull request with based on this work but with those things covered I'll happily merge it.

lukemurray:
Hey @brianc happy to look at this again. Last I looked I couldn't figure out the native mode. It seems to work quite differently to the other one. Do you have any tips? It seems to only get every result at once, not like the other mode. That is where I was hoping for some guidance. I'll have another look now that it may be merged.

Probably next week.

PR #787, opened 2015-05-21
Add warning on long query names

brianc:
Postgres has a 63 character limit on query names.  To avoid potential footgunning of users we'll log to their stderr if they use a longer query name.

For reference: https://github.com/brianc/node-postgres/pull/772

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -199,6 +199,11 @@ Connection.prototype.parse = function(query, more) {
 
   //normalize missing query names to allow for null
   query.name = query.name || '';
+  if (query.name.length > 63) {
+    console.error('Warning! Postgres only supports 63 characters for query names.');
+    console.error('You supplied', query.name, '(', query.name.length, ')');
+    console.error('This can cause conflicts and silent errors executing queries');
+  }
   //normalize null type array
   query.types = query.types || [];
   var len = query.types.length;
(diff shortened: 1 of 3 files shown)


bdowning:
I'll run it against my code when I get back from lunch.

brianc:
Awesome thanks!

bdowning:
Looks good to me.

```
:; ./node_modules/.bin/babel-node --stage 1 --optional bluebirdCoroutines main.js
Listening at http://[::]:3000
process 7459 handling GET /orders/1
Warning! Postgres only supports 63 characters for query names.
You supplied 
        SELECT * FROM (
        WITH matching_orders AS (
            SELECT o.id FROM tix.orders o WHERE o.id = $1
          ), matching_transactions AS (
[...]
             ON at.order_id = o.id
         WHERE o.id IN (SELECT id FROM matching_orders)
         GROUP BY o.id, e.name
    ) e
     ( 1836 )
This can cause conflicts and silent errors executing queries
```

bdowning:
If I'm not mistaken I think you need a check/warning in lib/native too though...

brianc:
Good catch!  Also: thanks so much for looking at this. :smile:

brianc:
Sorry for the delay on this...added warning for native bindings too!

PR #790, opened 2015-05-23
emit a connection event and custom validate connection

kesavkolla:


brianc:
Hey @kesavkolla - I'm a bit confused on what this PR is trying to accomplish.  There are also a _ton_ of merge commits keeping your branch up to date with master, which makes me think this is still important to you.  Care to elaborate on what you're trying to add here? I'd love to understand it in a bit more detail! 😄

kesavkolla:
Thanks for considering the commits. It's very minor change all I was doing is emitting event when a connection is retrieved from pool.  I've a need to set search_path and session timezone before making query. I've application thst deals with multiple timezones. The way I'm addressing that use case is by emitting an event on pool connection and on event handler I'm setting the timezone and search_path. The selection of timezone is based on user selection in application so can't be static string during initialization phase.  

I'd appreciate if you can either merge my pull request or provide any other mechanism to address the use case.

Regards,
-Kesav

brianc:
hey @kesavkolla - I've split the pool out into its own module & added a connect event to the pool based on your initial designs: https://github.com/brianc/node-pg-pool/blob/master/README.md#connect

If you're running the most recent version of this module the most recent version of pg-pool should be installed by default & you should be getting this behavior!

Sorry it took me a while to finally get to this PR. I really appreciate the thought & work you put in here & hope the implementation meets your needs. ❤️
 @brianc

kesavkolla:
Good to see the split of pg-pool into separate module.  I was thinking to suggest this from long time.  This allows people to write their own implementation of pooling.  I'm using ORM (loopback.io) I've to see if they upgrade their dependencies to this new version.  I hope this new version still maintains the backward compatibility in terms of API.

PR #804, opened 2015-06-14
Connection Parameters optimization

vitaly-t:
For an app that executes millions of requests through a connection pool it is absolutely unnecessary to parse the connection every single time. If the client provides its own instance of `ConnectionParameters`, it should be used instead of parsing everything all over again, on each request.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -12,7 +12,7 @@ var Connection = require(__dirname + '/connection');
 var Client = function(config) {
   EventEmitter.call(this);
 
-  this.connectionParameters = new ConnectionParameters(config);
+  this.connectionParameters = (config instanceof ConnectionParameters) ? config: new ConnectionParameters(config);
   this.user = this.connectionParameters.user;
   this.database = this.connectionParameters.database;
   this.port = this.connectionParameters.port;


brianc:
I appreciate the sentiment of this, but clients are not often created & destroyed.  If your app is executing millions of requests through a connection pool the clients are likely _never_ getting destroyed.  Therefore, the connection parameters are never re-created since that only happens in a client constructor.  Am I missing something?

vitaly-t:
Yes, you are missing the situation when the client passes a connection string, and that connection string is then passed into the connection pool where it is parsed to create a ConnectionParameters object,...ON EVERY SINGLE REQUEST! Doing all that on every request is a waste of resources, and quite pointless. If the client is smart enough to pass an existing connection object, it should be usable directly.

brianc:
Would you be able to show me in the code by providing links to the lines where the connection string is parsed every time a client is checked out from the connection pool? I can't find the code you're talking about.  AFAIK the string is only parsed when the client is first _constructed_ which only rarely happens.

vitaly-t:
The issue is quite easy to reproduce....

Your typical client works like this:

* calls `pg.connect(cn)`
* executes a query;
* calls `done()`

If you open file `connection-parameters.js`, and inject logging in the beginning of the constructor:

```javascript
var ConnectionParameters = function(config) {
  console.log("Calling parse() here...");
  config = typeof config == 'string' ? parse(config) : (config || {});
  // etc...
};
```

Then you will see that every time the client calls `pg.connect(cn)`, your code will parse the connection string every time, quite unnecessarily.

brianc:
I just tested this.  Ran 1,000 queries with the connection pool.  `ConnectionParameters` constructor was called 10 times (the size of the pool).  so I'm not sure what you're seeing, but I'm certain its only parsing the strings once per connection being __constructed__.

vitaly-t:
Ha! I had a larger pool, and it didn't occur to me to go over the pool size to see no more connections being created. You are absolutely right! Sorry for the turbulence!

brianc:
No worries! Glad its working for you!!

PR #819, opened 2015-07-29
Add reference to XeCycle/pg-template-tag to README

XeCycle:
Dead simple and super flexible query builder.  No idea about its performance, however I have been using it without trouble for quite some time (long before publishing to GitHub).


brianc:
Very nice! thanks!!

Issue #821, opened 2015-07-30
Unhandled exception when DB drops


rfshelley:
I've connected to the database, and then explicitly stop the DB, and on the next call to query the DB I get the error:

```
events.js:72
        throw er; // Unhandled 'error' event
              ^
error: terminating connection due to administrator command
    at Connection.parseE (.../node_modules/pg/lib/connection.js:539:11)
    at Connection.parseMessage (.../node_modules/pg/lib/connection.js:366:17)
    at Socket.Connection.attachListeners (.../node_modules/pg/lib/connection.js:105:22)
    at Socket.emit (events.js:95:17)
    at Socket.stream.pause.paused (_stream_readable.js:764:14)
    at Socket.emit (events.js:92:17)
    at emitReadable_ (_stream_readable.js:426:10)
    at emitReadable (_stream_readable.js:422:5)
    at readableAddChunk (_stream_readable.js:165:9)
    at Socket.Readable.push (_stream_readable.js:127:10)
```

This occurred from a simple "Client.query()" call from a pooled connection. The thrown exception kills the server and requires me to restart. Is there a better way to gracefully handle downed databases with a reconnect policy or at catch the exception?

sehrope:
I'm able to reproduce this with pg v4.4.0 on Linux. What's interesting is that you don't even have to run a query.

Here's the sample coffee-script program I used to test it (output is below). It creates a pooled connection, runs a test query, then prints that it's still alive continuously till it dies. Per the defaults the pooled connection should remain open for 30 seconds so that's enough time to manually stop the PostgreSQL server.

    pg = require 'pg'

    testQuery = (cb) ->
      pg.connect process.env.DATABASE_URL, (err, client, done) ->
        if err then return cb(err)
        client.query 'SELECT 1 AS x', (err, result) ->
          done(err)
          if err then return cb(err)
          console.log '%s - Ran a test query successfully', new Date()
          return cb(null)

    showAlive = () ->
      console.log "%s - Still alive", new Date()
      setTimeout showAlive, 100

    testQuery (err) ->
      if err
        console.error "Gracefully received error: ", err
        process.exit(1)
      showAlive()

Running with the server offline gives (proper behavior):

    Gracefully received error:  [Error: Connection terminated]

Running with the server online and then shutting it down gives (node process dies):

    Fri Jul 31 2015 19:10:38 GMT-0400 (EDT) - Ran a test query successfully
[...]

PR #824, opened 2015-08-02
Add failing connection pool test for server errors

sehrope:
As this is a failing test it really shouldn't be merged in until a fix is found for it. I'm adding it as a PR so a fix can be built atop it.

### DIFF
--- a/test/integration/connection-pool/terminate-backend-tests.js
+++ b/test/integration/connection-pool/terminate-backend-tests.js
@@ -1,0 +1,33 @@
+var helper = require(__dirname + "/../test-helper");
+var pg = require(__dirname + "/../../../lib");
+var async = require('async');
+
+// Creates a pooled connnection and returns it's backend PID
+// The pooled connection will remain in the pool for 30-seconds (default).
+function getBackendPid(cb) {
+  pg.connect(helper.config, function(err, client, done) {
+    if (err) return cb(err);
+    client.query('SELECT pg_backend_pid() AS pid', function(err, result) {
+      done(err);
+      if (err) return cb(err);
+      return cb(null, result.rows[0].pid);
+    });
+  });
+}
+
+// Create a new non-pooled connection and use it to terminate a backend.
+function terminateBackend(pid, cb) {
+  var client = helper.client();
+  client.query('SELECT pg_terminate_backend($1)', [pid], function(err, result) {
+    client.end();
+    return cb(err);
+  });
+}
+
+async.waterfall([
+  getBackendPid,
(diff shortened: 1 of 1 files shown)


brianc:
Nice! What's the expected behavior here?  The `pg` object will emit an error event when a client in the pool disconnects - should we handle that error & test for it?

sehrope:
> Nice! What's the expected behavior here?

Well not crashing the node process would be nice :)

As far as handling the error goes, it's a toss up between silently swallowing/logging it or saving it to dish out the next time that particular connection is checked out. I don't really see much value in the latter as if the server has been bounced most people would want a fresh connection (*i.e. transparent recovery*). If the server is still down (*i.e. the entire pool got flushed*), a new connection will be attempted and you'll get the error for that anyway. Probably would be more complicated too v.s. just swallowing it.

> The pg object will emit an error event when a client in the pool disconnects - should we handle that error & test for it?

Yes following up from info in #821 I'm going to test some options of adding a default handler to the `pg` object to catch these. Note that it shouldn't handle all errors, just errors for non-checked out pooled connections. The alternative is to document that everybody using pooled connections manually add their own top level error handler but that doesn't sound like a good idea.

brianc:
@sehrope I know it's taken me a while to get to it but I _think_ this should fix the error!  :smile: 

https://github.com/brianc/node-postgres/pull/843

sehrope:
Haven't had a chance to test this out yet but this looks like exactly what's need. Great idea to add/remove the error handler when the connection is idle in the pool. 

With how the rest of the error handling works, this would leave the bad connection in the pool right? So the next checkout of that connection would fail? If so, I wonder if there's a way to dispose the connection entirely. I don't see any value in having failed connections linger in the pool. Otherwise if you have a pool of size N, you'd need N checkouts to "flush" them out.

Still better than the entire node process crashing though :smile:...

sehrope:
Peeking through the code a bit, I think this fixes errors on first connect but not if the server is bounced after things have already been running. The same approach of should work for that too.

We'd need to add/remove the error handler when the connection is [checked out](https://github.com/brianc/node-postgres/blob/35e5567f86774f808c2a8518dd312b8aa3586693/lib/pool.js#L82) and [checked back in](https://github.com/brianc/node-postgres/blob/35e5567f86774f808c2a8518dd312b8aa3586693/lib/pool.js#L86).

brianc:
Closing this because it's 18 months old.  If it's still relevant I apologize: doing some house-cleaning before I start really going to town on fixing up some stuff & prepping for 7.0.

PR #828, opened 2015-08-02
Test that an active query is cancelled

kevinburke:
Test first makes a query to the active Postgres database, then issues a cancel
request, and verifies that the query was indeed canceled.

It looks like the test currently in that file only tests that the query is spliced from the queue.

### DIFF
--- a/test/integration/client/cancel-query-tests.js
+++ b/test/integration/client/cancel-query-tests.js
@@ -1,5 +1,20 @@
 var helper = require(__dirname+"/test-helper");
 
+test("cancellation of an active query", function() {
+  var client = helper.client();
+
+  var qry = "select pg_sleep(4)";
+  var query1 = client.query(qry);
+  client.cancel(client, query1);
+  var otherClient = helper.client();
+  activeQueries = "SELECT * FROM pg_stat_activity WHERE query LIKE '%pg_sleep(4)%' AND query NOT LIKE '%pg_stat_activity%'";
+  otherClient.query(activeQueries, assert.calls(function(err, result) {
+    assert.equal(result.rows.length, 0);
+    otherClient.end();
+    client.end();
+  }));
+});
+
 //before running this test make sure you run the script create-test-tables
 test("cancellation of a query", function() {
 
@@ -13,11 +28,11 @@ test("cancellation of a query", function() {
 
   var query1 = client.query(qry);
   query1.on('row', function(row) {
-    throw new Error('Should not emit a row')
+    throw new Error('Should not emit a row');
   });
   var query2 = client.query(qry);
   query2.on('row', function(row) {
-    throw new Error('Should not emit a row')
+    throw new Error('Should not emit a row');
   });
   var query3 = client.query(qry);
   query3.on('row', function(row) {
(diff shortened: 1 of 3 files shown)


test/integration/client/cancel-query-tests.js
```diff
@@ -1,5 +1,20 @@
 var helper = require(__dirname+"/test-helper");
 
+test("cancellation of an active query", function() {
+  var client = helper.client();
+
+  var qry = "select pg_sleep(4)";
+  var query1 = client.query(qry);
+  client.cancel(client, query1);
+  var otherClient = helper.client();
```
kevinburke: I was a little confused by this behavior, it seems like queries can't run concurrently on a given client? When I tried running this on the same client it would block until the `sleep` finished. If there's a way to do this without opening two connections let me know. 

I'm not 100% certain that the `cancel` will always complete before the `otherClient` issues a query against the database, mostly because I am not sure how `stream.write` works.

test/integration/client/cancel-query-tests.js
```diff
@@ -1,5 +1,20 @@
...
+  activeQueries = "SELECT * FROM pg_stat_activity WHERE query LIKE '%pg_sleep(4)%' AND query NOT LIKE '%pg_stat_activity%'";
+  otherClient.query(activeQueries, assert.calls(function(err, result) {
+    assert.equal(result.rows.length, 0);
+    otherClient.end();
+    client.end();
```
kevinburke: The process wouldn't exit without these lines; do you know why? (Obviously it's good to clean up resources, it just seemed kind of like magic)

test/integration/client/cancel-query-tests.js
brianc: Postgres only allows for & handles one executing query per connection (client) at a time.  If you try to submit two on the same client the 2nd one will return an error message.  That's been a built in constraint of postgres for as long as I've been using it & how it behaves when consumed from any programming language.  It actually makes some sense when you think about it too...how would transactions work in a semi-non deterministic environment of sending 2 queries over the network at the same time working within a single transaction?  Yikes.  So to work with this constraint node-postgres does two things:

1) Provides a client pool so you can run multiple queries "at once" because you're sending them on multiple different clients.
2) The node-postgres client internally queues up pending queries so if you send two at once it keeps them ordered and dispatches them one after another.

So its a pain to cancel a query because you have to issue one query from one client & then connect & cancel it from another client - there's no real way around it.  The client that's busy waiting on a query wont send any data to the backend, and even if you *did* the backend would ignore any data coming in on the client which is currently processing a query and simply return an error on it.

test/integration/client/cancel-query-tests.js
brianc: Each client opens a node `socket` to the database.  The sockets are still alive when the query is done & so the event loop is blocked.

brianc:
I appreciate the work here, but the tests are failing in travis - any way you could make them pass?

kevinburke:
Odd, the tests pass if I run the file individually, but fail if I run the suite. I'll try and figure out what's going on.

I assumed `cancel` was sync based on the signature, but reading through it, looks like that's not true? Might be running into a race here.

PR #837, opened 2015-08-12
forcing sydney time

forkfork:


sehrope:
Your patch changes the time zone to Sydney for all users. There's no way something like this would be merged in.

If you need this type of functionality you should either change your node server time zone or have your own wrapper that does this for your app.

forkfork:
My mistake, I had no idea I'd created a PR for the origin here, I was doing this to debug an issue on my upstream... github usage error by myself, apologies.

PR #840, opened 2015-08-20
Question: support arrays for JSON columns

hmalphettes:
Currently when a JSON array is inserted into a JSON column it is transformed into a String array suitable for an Array column.

A workaround consists of serialising the JSON array into a string before we insert its value into the database.
Reviewing the 'toPostgresql' function supported for object serialisation it looked like maybe we could do something here as well.

This PR defines an none enumerable property 'isJSON' to serialise the array as a String suitable for a JSON array.
Let me know if that is a good way to proceed.

We could also add a configuration parameter to the client that would assume that all arrays are targeting JSON columns.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -35,6 +35,9 @@ var prepareValue = function(val, seen) {
     return dateToString(val);
   }
   if(Array.isArray(val)) {
+    if (val.hasOwnProperty('isJSON') && !val.propertyIsEnumerable('isJSON')) {
+      return JSON.stringify(val);
+    }
     return arrayString(val);
   }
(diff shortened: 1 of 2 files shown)


brianc:
Sorry for the massive delay on this!  @hmalphettes I don't totally understand what you mean...could you provide a code example of the problem you're trying to fix?  If there's a way to accomplish what you want now without a code change that's probably preferable, but I'd like to get a better understanding of what you're going for.

brianc:
Closing this because it's old & I'm not understanding exactly whats being asked for.

hmalphettes:
@brianc fair enough. Thanks for your attention.

PR #843, opened 2015-08-24
Don't emit error events parsed out of data stream

### LINKED ISSUE #746: Unhandled 'error' event thrown when the connection password doesn't match
When I connect to an existing database with the connect username but the wrong password, I got an unhandled exception using the following code. The parameter err is set, but the exception is still thrown.

````
pg.connect(dbURL, function(err, client, done) {
  if(err) {
    return console.error('error fetching client from pool', err);
  }
  client.query('SELECT $1::int AS number', ['1'], function(err, result) {
    //call `done()` to release the client back to the pool
    done();

    if(err) {
      return console.error('error running query', err);
    }
    console.log(result.rows[0].number);
    //output: 1
    client.end();
  });
});
````

oliversalzburg:
Messages with the name `error` will also cause the `error` event to be raised on the stream. Thus, emitting the event here again will only cause problems.
Fixed #746

### DIFF
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -23,8 +23,13 @@ var pools = {
       log: clientConfig.poolLog || defaults.poolLog,
       create: function(cb) {
         var client = new pools.Client(clientConfig);
+        // Ignore errors on pooled clients until they are connected.
+        client.on('error', Function.prototype);
         client.connect(function(err) {
           if(err) return cb(err, null);
+
+          // Remove the noop error handler after a connection has been established.
+          client.removeListener('error', Function.prototype);
 
           //handle connected client background errors by emitting event
           //via the pg object and then removing errored client from the pool
--- a/test/unit/pool/basic-tests.js
+++ b/test/unit/pool/basic-tests.js
@@ -121,9 +121,10 @@ test('pool with connection error on connection', function() {
         process.nextTick(function() {
           cb(new Error('Could not connect'));
         });
-      }
+      },
+      on: Function.prototype
     };
-  }
+  };
   test('two parameters', function() {
     var p = pools.getOrCreate(poolId++);
     p.connect(assert.calls(function(err, client) {


brianc:
The connection parses errors out and emits them just as any other message so they can be handled higher up. Either by the client if a query is executing or emitted into the client's application.  Not emitting errors basically removes any indication that anything has gone wrong including a query failure, bad query syntax, a disconnection...anything.  Swallowing errors is a total no-go, sorry.  What is the actual problem you're running into?

oliversalzburg:
@brianc The problem I'm running into is #746. I also wrote up a short analysis of what's going wrong in there.

brianc:
Ah I see now - much more context.  Yeah so eating the error is probably not the right approach here.  In the case of a bad password __some__ installs of postgres will disconnect the socket after a bad password attempt.  This causes the socket to emit an error, which is passed up via the connection -> client like you were describing.  Problem is, which you also outlined, is that the client already received an error _message_ from postgres about the bad password.  The right way to handle this is probably something like this...

if the client is trying to connect and the stream emits an error or the client receives an error message and the client is pooled...then suppress any further error messages on the client within the pool, and dispose of the client.

I haven't been able to fix this because I don't have access to an instance of postgres that has the double-error problem, but if you wanna take a crack at fixing it there in the pool I'd 100% :heart: to merge the PR!

oliversalzburg:
Awesome. I'll try to find a proper solution. 

I'm really wondering about this double-error thing though. I was recently looking into a very similar issue at https://github.com/voxpelli/node-pg-pubsub/issues/6 and I'm assuming @voxpelli never experienced this either.

I think I saw mentioning that this might be a Windows issue, which seems reasonable, given that I'm on Windows.

oliversalzburg:
I hope I'm not causing too much trouble with the pushes, but I can't run the tests on Windows locally so far, so I'm letting travis handle that.

brianc:
No trouble at all. I only get travis email on master pass/fail or PRs I am
the submitter of.  Have at it!

oliversalzburg:
@brianc I'm a bit stuck right now. The test failure is telling me that `client.on` is not a function. But `EventEmitter.call(this);` is the first thing happening in the `Client` constructor. Any idea what's going on?

oliversalzburg:
Okay, so it's a `FakeClient` during testing. But that inherits from EventEmitter as well. I'm still confused :P

oliversalzburg:
Okay, got it, I was missing the mocked `Client` type. I opted to add a noop `on` to it, The only new thing that I added that uses `on`, only uses it to register a noop event handler. So that seemed pretty safe.

@brianc You mentioned disposing the client. From what I can tell, that should already be happening, because all the connection-related error handling is already in place. With the current approach, only the one specific case where two (or multiple) errors are generated, before a connection has been established. Is that enough?

oliversalzburg:
@brianc If you could provide some feedback on this, I'd really appreciate it :)

brianc:
@oliversalzburg sorry for taking a while to get to this - I think this actually looks really good!  Also - that's a cool trick for `noop` using function.prototype - I didn't know that one!  I'll merge this & push a new patch version right away. :)

oliversalzburg:
Awesome! :+1: :)

PR #852, opened 2015-09-03
Add lazy mode for row parsing

vickvu:
In lazy mode, raw values will be used and parsing function need to be called explicitly.
For example:
```javascript
function processRow(row) {
  return {
    name: row.obj_name, //Raw value
    createdTime: row.parse_obj_created_time() //Return the parsed value
  }
}
```
This would be useful when we want to access raw value from database, and also when we only want to access a subset of values. For example, suppose we have the following query result

obj_id | obj_name | obj_created_time | property_name 
-------- | ------------- | ----------------------- | ---------------
 1        | Object 1   |  2015-01-01         | Prop A
 1        | Object 1   |  2015-01-01         | Prop B
 1        | Object 1   |  2015-01-01         | Prop C

Ideally, we only need `obj_name` and `obj_created_time` of the first row and we don't need the parsing of `obj_created_time` to be run for more than once.

This PR will need a new version of pg-native (https://github.com/brianc/node-pg-native/pull/26)

### DIFF
--- a/lib/native/index.js
+++ b/lib/native/index.js
@@ -112,6 +112,7 @@ Client.prototype.query = function(config, values, callback) {
     query.name = config.name;
     query.callback = config.callback;
     query._arrayMode = config.rowMode == 'array';
+    query._lazyMode = config.rowMode == 'lazy';
   }
 
   //support query({...}, function() {}) style calls
--- a/lib/native/query.js
+++ b/lib/native/query.js
@@ -12,6 +12,7 @@ var NativeQuery = module.exports = function(native) {
   this.callback = null;
   this.state = 'new';
   this._arrayMode = false;
+  this._lazyMode = false;
 
   //if the 'row' event is listened for
   //then emit them as they come in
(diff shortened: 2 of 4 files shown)


linpy:
I am out of the office until 2015/09/04.

I would on leave from 2015/9/3. Please drop me mail if anything. Thanks.


Note: This is an automated response to your message  "[node-postgres] Add
lazy mode for row parsing (#852)" sent on 03/09/2015 15:51:23.

This is the only notification you will receive while this person is away.

forkfork:
@linpy any thoughts on this one?

brianc:
I really appreciate the thought you put into this & I can see cases where there would be a need for what you're describing.  However, it's already achievable without a change to the node-postgres source code.  I'd recommend if you want custom type parsing, including "lazy" type parsing by not parsing the rows, then you override the built in [type parser](https://github.com/brianc/node-pg-types)

You could even do something like this:

```
var client = new Client({
  types: { getTypeParser: () => val => val }
})
```
That will return every row as the raw text returned from the postgres protocol.  This could even be encapsulated in a separate module if you were interested! As far as adding it as a core feature to node-postgres I don't think that's the right way to go.

vickvu:
Thanks @brianc . I didn't know that I can override the built in type parsers.

PR #862, opened 2015-09-21
travis: fix node@4 compile errors

jonathanong:


brianc:
Thanks @jonathanong !

PR #866, opened 2015-09-26
Allow passwords to be provided using a function

arekinath:
This patch allows the `password` parameter, when given in a connection settings object, to be a function instead of a string. Example:

```js
var pg = require('pg');
var fs = require('fs');
var connOpts = {
  host: 'somehost',
  database: 'test',
  user: 'test',
  password: getPassword
};
pg.connect(connOpts, function (err, client, done) {
  ...
});

function getPassword(params, cb) {
  fs.readFile('some_pw_file', function(err, data) {
    if (err)
      cb(err);
    else
      cb(null, data.toString());
  });
}
```

This enables applications to provide their own mechanism for reading in (eg from a disk file) or calculating a password at the time of connection rather than in advance before setting up a pool.

Note that this works fine with pools, because `JSON.stringify()` on an object with a function-type member simply skips the function in its output.

I've also added tests for this feature, and amended a few old tests to make sure the code I changed still works.

The reason I personally want this is to be able to use one-time passwords with a postgres server, which have to be different each time you start a new connection (on the postgres side this is implemented by using LDAP auth, and the LDAP server supports the one-time password scheme), but it could also be useful:
 * in a situation where you want to change the password for a user without having to restart your application (it would always read the latest one), or
 * where you're not sure when/if you will use the postgres connection and you want to delay getting the password until you actually need it.

Let me know what you think.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -63,16 +63,27 @@ Client.prototype.connect = function(callback) {
 
   function checkPgPass(cb) {
     return function(msg) {
-      if (null !== self.password) {
+      if (self.password !== undefined && self.password !== null) {
         cb(msg);
-      } else {
-        pgPass(self.connectionParameters, function(pass){
-          if (undefined !== pass) {
-            self.connectionParameters.password = self.password = pass;
-          }
-          cb(msg);
-        });
+        return;
       }
+      self.connectionParameters.getPassword(function (err, pw) {
+        if (err) {
+          self.emit('error', err);
+          return;
+        }
+        self.password = pw;
+        if (null !== self.password) {
+          cb(msg);
+        } else {
+          pgPass(self.connectionParameters, function(pass){
+            if (undefined !== pass) {
+              self.password = pass;
+            }
+            cb(msg);
+          });
+        }
+      });
     };
   }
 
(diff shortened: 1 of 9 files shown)


arekinath:
It looks like the node 4.1 builds are failing because of issues in `nan` -- this is the same as on `master`?

brianc:
I really dig this!  I'll try to figure out why the 4.1 build is failing...it is passing for me locally. The native changes are such a pain some times!

brianc:
Alrighty - i fixed the travis errors on the master branch - could you rebase this one on top of that now? It should work great after that.  The code looks good & I'm happy to merge once the tests pass! Thanks a ton for the help here too. :+1:

arekinath:
I rebased, but still getting the nan errors. It looks like the last build of `master` failed again, too -- https://travis-ci.org/brianc/node-postgres/builds/82570813

(yay native)

brianc:
I appreciate the thought & work that went into this, but coming back after a long break & looking at it, I think the best approach to this situation is actually to read the password out of whatever system it's stored in (disk, db, whatever) *before* you initialize postgres.  I think this library already supports an overwhelming number of ways to configure clients & this adds more permutations.  I think the functionality is definitely useful, but should probably live in your code base or in a separate module.

PR #869, opened 2015-10-06
Support ending single pool

Suor:
Done with `pg.end(connString)`.

This is a draft pull request. Tell me what you think about API and/or code. Also it looks like to test this I'll need to add some sink magic to `ending-pool-tests.js`, I don't however have more than 1 config.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -17,27 +17,37 @@ var PG = function(clientConstructor) {
 
 util.inherits(PG, EventEmitter);
 
-PG.prototype.end = function() {
+PG.prototype.end = function(config) {
   var self = this;
-  var keys = Object.keys(self.pools.all);
-  var count = keys.length;
-  if(count === 0) {
-    self.emit('end');
+  var pool;
+  if (config) {
+    pool = self.pools.get(config);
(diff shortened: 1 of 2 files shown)


Suor:
Also what do you think of adding async.js as a dependency? It's common language now. For now I included `forEachOf()` implementation inplace.

Suor:
Any feedback on this?

brianc:
Hey sorry for missing this!  I like the idea, but is there any way to add a test for it?

Suor:
This is exactly my question to you :) I tried to understand how should I test connections in node-postges, but it looks to intricate for an outsider.

brianc:
In the most recent version I made it much easier (encouraged in fact) to create pool instances directly instead of relying on `pg.connect` - Check out the readme now on this repo for more info.  I plan on deprecating the singleton use of `pg.connect` soon.  In the mean time, I think the best path forward if you need to close a specific pool is instantiate the pool yourself and end it whenever you need to:

https://github.com/brianc/node-pg-pool

Thanks for this PR though!  I hope the new, easier to directly use pool makes it easier for you to end pools in your own apps!

Issue #858, opened 2015-09-17
npm install fail( node.js 4.1 pg 9.4.4)


realdog:
```

npm install pg pg-native

make: Entering directory `/application/wztj/gate1/node_modules/pg-native/node_modules/libpq/build'
  CXX(target) Release/obj.target/addon/src/connection.o
In file included from /home/kraken/.node-gyp/4.1.0/include/node/node.h:42,
                 from ../node_modules/nan/nan.h:24,
                 from ../src/addon.h:4,
                 from ../src/connection.cc:1:
/home/kraken/.node-gyp/4.1.0/include/node/v8.h:336: error: expected unqualified-id before ‘using’
/home/kraken/.node-gyp/4.1.0/include/node/v8.h: In constructor ‘v8::MaybeLocal<T>::MaybeLocal()’:
/home/kraken/.node-gyp/4.1.0/include/node/v8.h:353: error: ‘nullptr’ was not declared in this scope
/home/kraken/.node-gyp/4.1.0/include/node/v8.h: In member function ‘bool v8::MaybeLocal<T>::IsEmpty() const’:
/home/kraken/.node-gyp/4.1.0/include/node/v8.h:360: error: ‘nullptr’ was not declared in this scope
/home/kraken/.node-gyp/4.1.0/include/node/v8.h: In member function ‘bool v8::MaybeLocal<T>::ToLocal(v8::Local<S>*) const’:
/home/kraken/.node-gyp/4.1.0/include/node/v8.h:364: error: ‘nullptr’ was not declared in this scope
/home/kraken/.node-gyp/4.1.0/include/node/v8.h: In member function ‘bool v8::WeakCallbackInfo<T>::IsFirstPass() const’:
/home/kraken/.node-gyp/4.1.0/include/node/v8.h:430: error: ‘nullptr’ was not declared in this scope
/home/kraken/.node-gyp/4.1.0/include/node/v8.h: At global scope:
/home/kraken/.node-gyp/4.1.0/include/node/v8.h:469: error: expected unqualified-id before ‘using’
/home/kraken/.node-gyp/4.1.0/include/node/v8.h: In constructor ‘v8::Global<T>::Global()’:
/home/kraken/.node-gyp/4.1.0/include/node/v8.h:790: error: ‘nullptr’ was not declared in this scope
/home/kraken/.node-gyp/4.1.0/include/node/v8.h: In constructor ‘v8::Global<T>::Global(v8::Global<T>&&)’:
[...]

PR #872, opened 2015-10-07
Add clang flags to travis environment

brianc:
Trying to fix the building on travis...


abenhamdine:
Hi Brian,

Does it fix #858 ?

PR #875, opened 2015-10-10
Require statements break Browserify compatibility

joostfarla:
Using __dirname in require statements breaks compatibility with Browserify.

I've also submitted a related PR for pg-types (https://github.com/brianc/node-pg-types/pull/43).

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -4,10 +4,10 @@ var util = require('util');
 var pgPass = require('pgpass');
 var TypeOverrides = require('./type-overrides');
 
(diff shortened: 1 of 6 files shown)


brianc:
Thanks for this!  Why are you trying to browserify this library, by the way?

joostfarla:
@brianc we were trying to get the module up & running in AWS Lambda, using the JAWS framework. JAWS uses Browserify to bundle & compress the JS files. (see https://github.com/jaws-framework/JAWS/issues/201)

PR #877, opened 2015-10-12
Pass in connection string as property

calebmer:
Whereas, I would like to be able to configure some specific `pg` properties, yet at the same time also be able to use a connection string. Therefore I propose support for the following client configuration:
```js
{
  connection: 'postgres://brian:pass@host1:333/databasename',
  poolSize: 10
}
```

The connection string would override other properties (`user`, `port`, etc).

### DIFF
--- a/lib/connection-parameters.js
+++ b/lib/connection-parameters.js
@@ -35,7 +35,7 @@ var useSsl = function() {
 
 var ConnectionParameters = function(config) {
   config = typeof config == 'string' ? parse(config) :
-           typeof config.connection == 'string' ? parse(config.connection) :
+           typeof config == 'object' && typeof config.connection == 'string' ? parse(config.connection) :
            (config || {});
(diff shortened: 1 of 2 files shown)


calebmer:
I closed this issue because I couldn't remember why I opened it, but on second thought, I do still want this feature. Is there anyone willing to comment at this point?

calebmer:
This could be very helpful for CLI tools that want users to provide only a PostgreSQL URL and also wants to configure out-of-band options like pool size.

langpavel:
I'm not sere if this is useful. When `connection` string is passed in, others should be used as defaults or others overrides connection string..? It make more sense first but.. When anybody wants this, it's trivial to implement using 'pg-connection-string' module. I personally want this but in my code not in this library.

calebmer:
Yeah, that's reasonable.

PR #878, opened 2015-10-14
escapeLiteral handles more types of input

andythiv:
I added functionality to escape more kinds of values instead of only just strings similar to the way other SQL libraries do it:
https://www.npmjs.com/package/mysql#escaping-query-values

I left the existing code for handling strings in place and now also handle null, undefined, numbers, booleans, and arrays. The only strange thing might be that if the value being escaped is not one of those types, I just pass it back.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,8 +42,6 @@ test-missing-native:
 	@echo "***Testing optional native install***"
 	@rm -rf node_modules/pg-native
 	@node test/native/missing-native.js
-	@npm install pg-native@1.4.0
-	@node test/native/missing-native.js
 	@rm -rf node_modules/pg-native
 
 node_modules/pg-native/index.js:
--- a/lib/client.js
+++ b/lib/client.js
@@ -277,11 +277,11 @@ Client.prototype.escapeLiteral = function(value) {
     The unit tests cover this behavior.
     */
     return value == true ? 'TRUE' : 'FALSE'; // jshint ignore:line
-  } else if (value instanceof Date){
+  } else if (value instanceof Date) {
     return utils.dateToString(value);
   } else if (value.constructor === Array) {
     return utils.arrayString(value);
-  } else if(typeof value === 'string' || value instanceof String){
+  } else if(typeof value === 'string' || value instanceof String) {
     // Ported from PostgreSQL 9.2.4 source code in src/interfaces/libpq/fe-exec.c
     var hasBackslash = false;
     var escaped = '\'';
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pg",
-  "version": "4.4.1",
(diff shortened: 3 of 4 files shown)


andythiv:
First I had a jshint failure, but I added an ignore and an explanation for it, which I think you should find acceptable.

However, now I have a problem with a unit test that relies on checking a string output based on a Date object, so the timezone the test runs in makes a difference. I thought I already accounted for this, but I'm looking into it now.

andythiv:
Fixed it

andythiv:
Fixed merge conflict

abenhamdine:
Great !
I am very interested by this
@brianc would it be possible for you to review and merge this one ?

brianc:
First of all, let me say I'm sorry for missing this when it first came in!  :grimacing: 

@rpedela initially ported the libpq version of escape literal to JavaScript, which is what the existing code does.  The mirroring of the libpq version is intentional to provide the 'minimum' version of functionality available to those who don't have access to a compiler & libpq native bindings.  

I'm not sure extending this with extra escaping really belongs inside node-postgres itself.  Is there a reason you couldn't spin this out into a separate module...something like `pg-escape` that one could opt-in to separately with an additional dependency?  This escaping definitely seems helpful, but unless it exists in libpq itself the additions are arbitrary & I could see this growing in complexity & usefulness...which is a perfect place for a different module!

Thoughts?

rpedela:
I made a library specifically for this. It is a Javascript implementation of PG's built-in function [format()](http://www.postgresql.org/docs/9.4/static/functions-string.html#FUNCTIONS-STRING-FORMAT). It also handles all Node.js types including JSON and buffers. The library is pretty mature, but I am happy to resolve issues or take PRs.

https://github.com/datalanche/node-pg-format

brianc:
@rpedela to the rescue! :dancer:

rpedela:
LOL! :)

abenhamdine:
@rpedela thanks for this module !

brianc:
I'm going to close this in favor of using @rpedela's node-pg-format module. :smile:

rpedela:
Just an idea. The status quo is good too.

What if the source was kept as a module but it was included in the main ```pg``` as a replacement for the escape functions? I think it would help people write better raw SQL, and it isn't intrusive. For example, this is a trivial example of how I usually use the module.

```javascript
var pg = require('pg');
pg.format = require('pg-format');

var sql = pg.format('SELECT * FROM t where c = %L', 34);
var client = new pg.Client('postgres://username:password@localhost/database');

client.connect(function (err) {

    if (err) {
        return console.error(err);
    }

    client.query(sql, function (err, result) {

        client.end();

        if (err) {
            return console.error(err);
        }

        console.log(result);
    });
});
```

PR #888, opened 2015-11-14
Error message misspelling.

sibedge:
Fixing error message misspelling.


brianc:
Ah thank you! Sorry it took me a while to get to this.  Every little bit helps. :smile: :money_with_wings:

PR #899, opened 2015-12-01
readme entry for pg-gen

jray:
Adding a readme entry for [pg-gen](https://github.com/jray/pg-gen)


brianc:
Awesomeness! :boom: :dancer:

PR #907, opened 2015-12-29
doc update change username from jray to recursivefunk

recursivefunk:
github seems to be taking care of redirects for now, just wanted to get ahead of it.


brianc:
Dig the new user name!

PR #910, opened 2016-01-07
Support for defaults.connectionString

### LINKED ISSUE #384: [enhancement] Support pg.defaults.connectionString
One doesn't have to specify connectionString on each pg.connect() by using pg.defaults. 
However, there's no easy way to set the defaults using a connectionString, rather than setting pg.defaults.user, pg.defaults.database...

Please support pg.defaults.connectionString as an alternative to the current user/db/pass...  set.

bdunavant:
An attempt to resolve #384 

If a connectionString is provided in defaults it ignores anything else set by the defaults file and acts like the user passed in the connection String to the constructor.   I had considered parsing it and overlaying it on top of the 'defaults' file but felt this is probably 'more correct.'  If you disagree I will fix.

### DIFF
--- a/lib/connection-parameters.js
+++ b/lib/connection-parameters.js
@@ -34,7 +34,13 @@ var useSsl = function() {
 };
 
 var ConnectionParameters = function(config) {
+  //if a string is passed, it is a raw connection string so we parse it into a config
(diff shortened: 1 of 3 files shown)


brianc:
Thanks @bdunavant! Sorry for the delay in getting to this

PR #913, opened 2016-01-11
Fix of pool leaking

NumminorihSF:
There is some bug with connection pool. 
To reproduce it - you need connect to postgresql, start request and kill db process by `kill -9`. Client will not be free on any time. 

To stop this, we can add heartbeat to net.Socket on pg.Client. To force backward capability heartbeat on default is 0. If heartbeat is 0, no heartbeat will be used. 

I have tested it on my db, seems, that it works fine.

P.S. sorry for my English.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -25,7 +25,8 @@ var Client = function(config) {
 
   this.connection = c.connection || new Connection({
     stream: c.stream,
-    ssl: this.connectionParameters.ssl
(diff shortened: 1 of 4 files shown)


NumminorihSF:
@brianc, did you see this?) or you only read issue now?)

marcbachmann:
:+1: Would be nice if this lands in master soon

PR #918, opened 2016-01-14
Fix of pool leaking by TCP-keepalive

NumminorihSF:
There is some bug with connection pool.
To reproduce it - you need connect to postgresql, start request and kill db process by kill -9. Client will not be free on any time.

To fix this, we can add heartbeat to client connection (https://github.com/brianc/node-postgres/pull/913) or add TCP-keepalive.

I have tested it. On TCP-socket close it return error to `client#query` callback function. I think, that it is the best resolve of this problem.

P.S. Please, if you will merge one of my PR - close another one.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -48,6 +48,7 @@ Connection.prototype.connect = function(port, host) {
 
   this.stream.on('connect', function() {
     self.emit('connect');
+    self.stream.setKeepAlive(true);
   });
 
   this.stream.on('error', function(error) {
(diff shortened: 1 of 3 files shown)


brianc:
@NumminorihSF sorry for missing this!  This definitely looks like a solid approach.  I'll go ahead & merge this & close the other pull request!  In order to make this backwards compatible I'm going to add  this as a config option to the client:

```
var client = new Client({
  keepAlive: true
})
```

It will default to `false` and only alter current behavior if set to true.

PR #925, opened 2016-01-27
Fix prepare value map to postgres

mnahkies:
There was an issue where if a complex value (a value having a toPostgres function defined) was passed to prepareValue from a map operation, the index of the value would be passed as an additional parameter and prevent the internal seen variable from being correctly initialized.

Rather than try and identify all usages of this and fix the individual map operations, I thought it would be better to prevent the seen variable from being accidentally tampered with.

An example of current usage that causes problems: https://github.com/brianc/node-pg-cursor/blob/master/index.js#L6

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -103,6 +103,10 @@ function normalizeQueryConfig (config, values, callback) {
 }
 
 module.exports = {
-  prepareValue: prepareValue,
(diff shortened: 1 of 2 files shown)


brianc:
Very nice @mnahkies!  I'll get a patch version pushed to npm today! Thank you!

PR #926, opened 2016-01-27
Adding MassiveJS to Extras section of README

bradyholt:


brianc:
Awesome! Sorry again for the delay!

Issue #915, opened 2016-01-12
Request: mention pg-rxjs in Readme


jadbox:
I've created a library forked from pg-then that's for specific Rx: 
[pg-rxjs](https://github.com/jadbox/pg-rxjs)

Almost full code coverage is included.

sibedge:
@jadbox you could reduce all that code to about 10% of it, using [pg-promise](https://github.com/vitaly-t/pg-promise), because it already has [automatic transactions](https://github.com/vitaly-t/pg-promise/wiki/Learn-by-Example#transactions) in it, plus [automatic streaming through pg-query-stream](https://github.com/vitaly-t/pg-promise/wiki/Learn-by-Example#streaming).

jadbox:
@sibedge I too use QueryStream for the Stream operator. I do like pg-promise though, but it's not as thin as a Iayer I was wanting for pg-rxjs. Also, wrapping Rx over promises creates unnecessary performance issues by having the two handshake.

rpedela:
Fork, modify readme, and make a pull request.

jadbox:
^ done and done

PR #927, opened 2016-01-27
pg-rxjs mention added

jadbox:
See issue:
https://github.com/brianc/node-postgres/issues/915#issuecomment-171454974


brianc:
+1 fantastique! :dancer:

PR #928, opened 2016-01-28
Support TCP keepalives similarly to libpq

### LINKED ISSUE #909: TCP keepalives - possible discrepancy with libpq/pg-native
I'm just beginning to investigate this issue, on a hunch that it is the cause of a problem I'm having in a `pg-then`-based app (if the network fails during a query, it hangs).

First, I don't see that `node-postgres` supports the [`keepalives`](http://www.postgresql.org/docs/9.4/static/libpq-connect.html#LIBPQ-KEEPALIVES) connection option in any way. Though I don't know if that's intended.

Second, libpq (and therefore `pg-native`?) enables TCP keepalives by default, and thus should be able to detect a dropped connection; whereas AFAICT Node's `net.Socket` (and therefore `pg`) does not. So, with the same connection string that doesn't explicitly set `keepalives`, different behaviors would be observed with `pg` vs. `pg-native`.

Anyone care to comment/verify/enlighten? :)

motiz88:
This depends on https://github.com/iceddev/pg-connection-string/pull/7 (which is why the tests don't pass yet) and would fix #909. Tested in production in my project. This is a more fleshed-out solution than #918.

I'd appreciate a comment from a maintainer here in support of the `pg-connection-string` PR, so that it gets merged there, after which the tests here can be re-run and the PR merged.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -25,7 +25,8 @@ var Client = function(config) {
 
   this.connection = c.connection || new Connection({
     stream: c.stream,
-    ssl: this.connectionParameters.ssl
+    ssl: this.connectionParameters.ssl,
+    keepalives: this.connectionParameters.keepalives
   });
   this.queryQueue = [];
   this.binary = c.binary || defaults.binary;
--- a/lib/connection-parameters.js
+++ b/lib/connection-parameters.js
@@ -48,6 +48,9 @@ var ConnectionParameters = function(config) {
 
   this.application_name = val('application_name', config, 'PGAPPNAME');
   this.fallback_application_name = val('fallback_application_name', config, false);
+
+  this.keepalives = 'keepalives' in config ? config.keepalives : val('keepalives', config);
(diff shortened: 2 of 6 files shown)


motiz88:
@brianc? Anyone?

jspavlick:
+1

brianc:
Hey @motiz88 sorry for missing this - currently the tests are failing & the PR wont cleanly merge.  I really appreciate the thought you put in here...I've been busy w/ some life stuff & just now returning to computer-world.  Do you think you could get the tests to pass & rebase your changes onto current master & resolve conflicts? Afterwards i'll happily merge & push a new minor version.

motiz88:
I'd love to. If I don't get around to it in a week or so, please feel free
to ping me about it!

NumminorihSF:
@brianc, can you close  #948, #918, #913 if you will merge this?

brianc:
This looks like it'll merge cleanly, but all the tests are still failing. :( If they pass I'd happily merge it in!

brianc:
Support will be merged in with `pg@6.0.0` for this feature! Thanks for taking the time to report it & put thought & work into it!! 👏

PR #937, opened 2016-02-12
Bring native API in line with pure js version. Fixes #743 

### LINKED ISSUE #743: require('pg').native doesn't work with query row event
If I uncomment the query.on('row'...) section below, node complains that the function addRow does not exist. When I just use require('pg') it works properly.

    var pg = require('pg').native;

	pg.connect(connectionString, function(err, client, done) {
	
		if (err) {
			console.error(err);
		}

		var query = client.query(queryString, params);
		
		query.on('error', function(err) {
			console.error(err);
		});
		
		/*query.on('row', function(row, result) {
			result.addRow(row);
		});*/

		query.on('end', function(result) {
			done();
		});
	});

coagmano:
>  node-postgres abstracts over the pg-native module to provide exactly the same interface as the pure JavaScript version. **No other code changes are required.** If you find yourself having to change code other than the require statement when switching from require('pg') to require('pg').native please report an issue.

In issue #743 the code breaks using `pg-native` where the `Result.addRow` method is used. 

This method is missing from the native bindings because libpq loads all rows into memory before handing it to pg-native. The native bindings already simulate emitting row events, but does not replicate the addRow function.

I've added a test in `test/integration/client/simple-query-tests.js` using `Result.addRow`. 

And an empty function stub in `lib/native/result.js` which doesn't have to do anything since `pg-native` accumulates rows by default.

### DIFF
--- a/lib/native/result.js
+++ b/lib/native/result.js
@@ -19,3 +19,10 @@ NativeResult.prototype.addCommandComplete = function(pq) {
     });
   }
 };
+
+NativeResult.prototype.addRow = function(row) {
+  // This is empty to ensure pg code doesn't break when switching to pg-native
+  // pg-native loads all rows into the final result object by default.
+  // This is because libpg loads all rows into memory before passing the result
+  // to pg-native.
+};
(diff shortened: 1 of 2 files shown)


brianc:
Hey this is a great find & fix!  Thank you!! Merging now. :)

Issue #902, opened 2015-12-12
bug w/ generic-pool effects node-postgres


timothykimball:
See https://github.com/coopernurse/node-pool/issues/108

brianc:
Thanks for letting me know about this! Once they sort out the issue feel
free to lemme know if i need to expose something in the API here.

charmander:
pg-pool 2 dropped generic-pool.

Issue #805, opened 2015-06-17
Set a timeout on pg.connect


kevinburkeshyp:
Hi,
Thanks for this library, it's great.

Currently if I call `pg.connect` but my pool is full, node-postgres will block until a connection becomes available. Is there a way to set a timeout on a connection attempt? I would rather not block indefinitely if the pool is full. 

Something like `pg.connect({timeout: 500}, function(err) { console.log(err); }`

I can wrap it in a `setTimeout`, but do I need to do cleanup/is there a way to cancel a connection attempt?

kevinburkeshyp:
Ugh, looks like this is a known issue in the pool implementation: https://github.com/coopernurse/node-pool/issues/85

vitaly-t:
If you use the pool correctly, you won't run out of connections.

By default the library supports 10 connections in the pool. If you release the connection properly after you're done using it, then you won't run out of the connections.

kevinburkeshyp:
Hi Vitaly,
Thanks for your feedback. There are two reasons you could run into this:

- you're checking out more than 10 connections at a time, or some anomalous event causes load on your system that causes the number of connections to increase beyond the number of connections your database can handle at any one time.
- you are not releasing the connection in a place that you should be. it can be tricky, and no one's perfect! this is what actually bit me and turned me on to this issue - my tests were blocking and i couldn't figure out why.

in either situation, it's good to be defensive; rather than tying up memory and FD's waiting for a connection that may never materialize, I would rather bail early and return an error to the caller. this pattern's used by netflix, for example, see here: https://github.com/Netflix/Hystrix/wiki/How-it-Works#isolation

vitaly-t:
Most likely it is the second case:
>you are not releasing the connection in a place that you should be

Managing connections can be tricky. It is impossible to know without the code analysis.

I stopped using connections directly ever since I wrote [pg-promise](https://github.com/vitaly-t/pg-promise). Maybe it can help you too ;)

ensonik:
@vitaly-t Hi Vitaly, @kevinburkeshyp point is that a system shouldn't become unresponsive or crash when another external system is mis-behaving. By not having a connection timeout, my application will fail miserably because of something that's out of my control (for example, if the posgresql server is slow). 

With the timeout option I can simply return an error to the client after half a second if I chose. 

This concept and problem area exists wether I use pg-promise or not.

vitaly-t:
test

vitaly-t:
[...]

Issue #397, opened 2013-07-11
Add ability to set connect timeout


nfitch:
I read the docs on client options:
https://github.com/brianc/node-postgres/wiki/Client#constructor

And didn't see a way to set the connection timeout.  A simple example of what I'd like to tune:

```
var pg = require('pg');

//Assuming no machine with ip 192.168.255.255 exists
var client = new pg.Client('tcp://root:root@192.168.255.255');
var start = (new Date()).getTime();
client.connect(function (err) {
        var end = (new Date()).getTime();
        console.log({
                err: err,
                elapsedTime: end - start
        });
});
```

Results in:

```
[nate@dev madtom]$ node ./test/postgres_timeout.js
{ err:
   { [Error: connect ECONNREFUSED]
     code: 'ECONNREFUSED',
     errno: 'ECONNREFUSED',
     syscall: 'connect' },
  elapsedTime: 75818 }
```

Notice that the elapsed time is 75 seconds.

jwhitehorn:
:+1:

booo:
+1 comments are super helpful to understand problems...

jhansen-tt:
+1

brianc:
Anyone care to take a crack at this in a pull request?

zacharyrankin:
hi @nfitch,

I was looking for the same thing.  I am not using the pool feature, but pg.Client
directly.  I was playing with a couple of ideas.

Example 1:

```js
let stream = new net.Stream()
stream.setTimeout(2000) // however many seconds
stream.on('timeout', function () {
  console.log('client timed out')
  this.destroy()
})
let config = {
  // connection info
  stream: stream
}
let client = new pg.Client(config)
client.connect(err => {
  if (err) {
    console.error(err)
  }
  console.log('connected')
})
```

In Example 1, if a query takes longer than 2 seconds to return
results I think it will also trigger that timeout.  So here was another idea I had if you
just wanted to have a connection timeout.

```js
let stream = new net.Stream()
let config = {
  // connection info
  stream: stream
}
let client = new pg.Client(config)
let connTimeout = setTimeout(() => stream.destroy, 2000)
client.connect(err => {
  if (err) {
    console.error(err)
  }
  console.log('connected')
  clearTimeout(connTimeout)
})
```

These are just some ideas.  Let me know your thoughts.

charmander:
Implemented as [`connectionTimeoutMillis`][1].

  [1]: https://github.com/brianc/node-pg-pool#create

[...]

PR #939, opened 2016-02-17
POC: Time out pg.connect() call

### LINKED ISSUE #782: More documentation on pooling
Can you elaborate on how pooling works?  Right now I create a new instance like the following in every request:
```
someHttpGetRoute(req, res){
        var pgClient = new pg.Client(conString);
        pgClient.connect(function(err){
            pgClient.query('some query', [...], function(err, result){
                pgClient.end();
                res.send(result);
            });
        });
}
```
I read the documentation more carefully and it seems that this is bad because creating a connection takes some effort and this request is called very often (multiple times every second).
I want to switch to pooling, but I don't know how pooling will work after I exceed the pool size.  Let's say I set the size to 20, but 21 requests happened at the same time, what happens to the last request?  Is it not responded?  Blocked until one connection becomes available? The pool grows automatically to handle this request?  Or is doing new pg.Client(conString); the right way to go?

Thanks in advance

kevinburkeshyp:
Currently if you call pg.connect(), the call will block indefinitely until a
connection becomes available. In many cases, if a connection is not available
after some period of time, it's preferable to return an error (and call
control) to the client, instead of tying up resources forever.

Blocking on resource checkout also makes it easier for clients to deadlock -
recently at Shyp, we had a situation where a row got locked and the thread
that could unlock it was blocked waiting for a connection to become available,
leading to deadlock. In that situation, it would be better to abort the
checkout, which would have errored, but also broken the deadlock.

Add two new settings to defaults: `block`, which, if false, will immediately
return an error if the pool is full when you attempt to check out a new
connection. Also adds `acquireTimeout`, which will wait for `acquireTimeout`
milliseconds before giving up and returning an error.

This builds on two pull requests against `generic-pool`:

- Support options.block: https://github.com/coopernurse/node-pool/pull/125
- Support options.timeout: https://github.com/coopernurse/node-pool/pull/127

For the moment I'm pointing `generic-pool` at a branch that incorporates
both of these commits. I'm marking this as a proof-of-concept until those go
through, which hopefully they will soon. I'd also like feedback on the
API.

Adds semicolons in many places that omitted them and fixes several typos. I'm
happy to pull those out into a different commit.

Sets the TZ=GMT environment variable before running the tests; without this
value set, and with a Postgres server set to the America/Los_Angeles timezone,
a timezone test failed.

Fixes #782 and #805. Will help alleviate #902. May help with #397.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,10 +4,10 @@ connectionString=postgres://
 
 params := $(connectionString)
 
-node-command := xargs -n 1 -I file node file $(params)
+node-command := TZ=GMT xargs -n 1 -I file node file $(params)
 
 .PHONY : test test-connection test-integration bench test-native \
-	 jshint publish test-missing-native update-npm
+	 jshint publish test-missing-native update-npm clean
 
 all:
 	npm install
@@ -67,3 +67,6 @@ prepare-test-db:
 jshint:
 	@echo "***Starting jshint***"
 	@./node_modules/.bin/jshint lib
+
+clean:
+	@rm -rf node_modules
--- a/lib/defaults.js
+++ b/lib/defaults.js
@@ -30,6 +30,17 @@ var defaults = module.exports = {
   //from the pool and destroyed
   poolIdleTimeout: 30000,
 
+  // acquireTimeout is the amount of time in ms to wait to check out a database
+  // connection. By default, node-postgres will wait indefinitely for a
+  // connection to become available. If the acquireTimeout is negative,
+  // or NaN, pg.connect will return an error immediately. If the acquireTimeout
+  // is zero, node-postgres will return an error in the next tick if the pool
+  // is full.
+  //
+  // The error message on timeout will be "Cannot acquire resource because the
+  // pool is full".
+  acquireTimeout: undefined,
+
   //frequency to check for idle clients within the client pool
   reapIntervalMillis: 1000,
 
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   "main": "./lib",
   "dependencies": {
     "buffer-writer": "1.0.1",
-    "generic-pool": "2.1.1",
+    "generic-pool-timeout": "2.7.1",
     "packet-reader": "0.2.0",
     "pg-connection-string": "0.1.3",
     "pg-types": "1.*",
--- a/script/test-connection.js
+++ b/script/test-connection.js
@@ -9,7 +9,7 @@ pg.connect(helper.config, function(err, client, done) {
     console.error(err);
     process.exit(255);
   }
-  console.log("Checking for existance of required test table 'person'")
+  console.log("Checking for existence of required test table 'person'")
   client.query("SELECT COUNT(name) FROM person", function(err, callback) {
     if(err != null) {
       console.error("Recieved error when executing query 'SELECT COUNT(name) FROM person'")
--- a/test/unit/pool/basic-tests.js
+++ b/test/unit/pool/basic-tests.js
@@ -5,29 +5,30 @@ var libDir = __dirname + '/../../../lib';
 var poolsFactory = require(libDir + '/pool')
 var defaults = require(libDir + '/defaults');
 var poolId = 0;
+var pg = require(libDir);
 
 require(__dirname + '/../../test-helper');
 
 var FakeClient = function() {
   EventEmitter.call(this);
-}
+};
 
 util.inherits(FakeClient, EventEmitter);
 
 FakeClient.prototype.connect = function(cb) {
   process.nextTick(cb);
-}
+};
 
 FakeClient.prototype.end = function() {
   this.endCalled = true;
-}
+};
 var pools = poolsFactory(FakeClient);
 
 //Hangs the event loop until 'end' is called on client
 var HangingClient = function(config) {
   EventEmitter.call(this);
   this.config = config;
-}
+};
 
 util.inherits(HangingClient, EventEmitter);
 
@@ -36,11 +37,11 @@ HangingClient.prototype.connect = function(cb) {
     console.log('hung client...');
   }, 1000);
   process.nextTick(cb);
-}
+};
 
 HangingClient.prototype.end = function() {
   clearInterval(this.intervalId);
-}
+};
 
 test('no pools exist', function() {
   assert.empty(Object.keys(pools.all));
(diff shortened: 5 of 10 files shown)


test/unit/pool/basic-tests.js
```diff
@@ -210,6 +211,66 @@ test('pool#connect client.poolCount', function() {
...
       });
-    })
+    });
+  });
+});
+
+pg.defaults.block = false;
+pg.defaults.poolSize = 1;
+
+test('pool#connect acquire errors if the pool is full', function() {
```
kevinburkeshyp: One of these tests introduces an unnecessarily long timeout, maybe 2 seconds. I'm not super familiar with how the test runner knows when tests are done or I would address it, was hoping you'd know :(

kevinburkeshyp: Also - there are more tests around the `generic-pool` implementation then there are here, this just covers the common cases.

Makefile
```diff
@@ -4,10 +4,10 @@ connectionString=postgres://
 
 params := $(connectionString)
 
-node-command := xargs -n 1 -I file node file $(params)
+node-command := TZ=GMT xargs -n 1 -I file node file $(params)
```
brianc: +1 nice - sorry I missed that one!

lib/defaults.js
```diff
@@ -30,6 +30,24 @@ var defaults = module.exports = {
   //from the pool and destroyed
   poolIdleTimeout: 30000,
 
+  // block sets the behavior of node-postgres when the connection pool is
+  // full. Set to false to immediately return an error if no connections are
+  // available. By default, node-postgres will wait an infinite amount of time
+  // for a connection to become available.
+  block: true,
```
brianc: +1 very nice feature

lib/defaults.js
```diff
@@ -30,6 +30,24 @@ var defaults = module.exports = {
...
+  // database connection. By default, node-postgres will wait infinitely for
+  // a connection to become available. If the acquireTimeout is negative, zero,
+  // or NaN, pg.connect will return an error immediately.
+  //
+  // If block is set to false, this setting is ignored - node-postgres will
+  // error immediately if block is false and the pool is full.
+  //
+  // The timeout message will be "Cannot acquire resource because the pool is
+  // full"
+  acquireTimeout: undefined,
```
brianc: Related to `block = true` above - this _could_ be a single property - if you set `acquireTimeout === 0` then that could be the same meaning as `block = true` removing the need for the `block` property. 

So, 
```js

acquireTimeout: 0 //timeout immediately, effectively block: true
acquireTimeout: > 0 //timeout after this amount of time
acquireTimeout: <any other value> //noop - same behavior as existing which is to wait 'forever' to timeout
```

Up to you on the implementation & decision here 100% - just putting in my 2 cents.  Less properties is nice because then you don't have to explain how a `block = false && acquireTimeout = 100` will interplay with one another; however, explicitness can be a nice feature of an API as well.

package.json
```diff
@@ -19,7 +19,7 @@
   "main": "./lib",
   "dependencies": {
     "buffer-writer": "1.0.1",
-    "generic-pool": "2.1.1",
+    "generic-pool": "git://github.com/Shyp/node-pool.git#shyp-master",
```
brianc: Would you be opposed to waiting for the upstream fixes to `generic-pool` to land in npm before we merge this? Alternatively you could publish your fork at a different module name on npm?  I really don't want to rely on github for installs - this breaks some production workflows and is especially painful whenever github goes down.  What do you think?

script/test-connection.js
```diff
@@ -9,7 +9,7 @@ pg.connect(helper.config, function(err, client, done) {
     console.error(err);
     process.exit(255);
   }
-  console.log("Checking for existance of required test table 'person'")
+  console.log("Checking for existence of required test table 'person'")
```
brianc: :facepalm:

brianc:
I'm a big +1 on these features!  Thanks for putting in the time to implement these.  I'm on a cruise right now :ship: with very very bad internet speeds so apologies for taking a while to get to this.

As far as the test runner is concerned its very primitive because node-postgres was written before most of the now standard test frameworks for node existed, so I rolled my own with makefiles.  Timeouts are handled by the test suite hanging indefinitely - so you can use as short of timeouts as you want.  Once I'm back on Monday I'll have time to finish reviewing this - there are a few utility methods I monkey-patched on to the `assert` module way back in the day that make testing async code easier...I'll try pointing them out once I get to a place where I can pull up new internet pages in under 30 seconds. :grimmace:

package.json
kevinburkeshyp: I'm not opposed, and I agree that you shouldn't rely on github for installs. I'm not sure when it's going to get merged, and I don't want to put any pressure on the maintainer. I wouldn't mind publishing this on a fork, but I'm also a little wary about contributing to package bloat :(

sandfox: Hopefully I'll get it merged this weekend! I haven't really had a chance to read it all through yet. Just put out a bunch of minor bug fixes and stuff in v2.4.1 so these fixes should be in v2.5.0 and the only thing I have lined up for it

brianc:
Hey @kevinburkeshyp thanks again for putting this together!  Once the PR is merged for `node-pool` if you can, just update this PR to use the npm published version in package json, then come back here and @ mention me on this PR & I'll merge it up! :dancer:

vitaly-t:
I hope this is merged soon ;)

brianc:
Hey @kevinburkeshyp did the changes ever land upstream in generic pool allowing for a timeout before failing the connection?

kevinburkeshyp:
no, the maintainer has said for 3 months now he's going to look at it, but no progress

brianc:
Ah - sorry about that!  That maintainer sounds about as slow as me. 😬   Once its on npm I'll merge this ASAP.

vitaly-t:
Can we not finish this without that maintainer? He may as well never come back at this point...

brianc:
If you fork generic-pool and push it to a new package on npm & then add *that* package as a dependency that will unblock this - I don't think that is the ideal solution, but it would work.  This would also need to be cleaned up so it would once again cleanly merge into master.  Then we're good to go! 👏

kevinburkeshyp:
Sounds good! Will let you know!

kevinburke:
Submitted as #1006! Closing.

PR #942, opened 2016-02-23
Add pg-client to the list of extras in README.md

koistya:
https://github.com/kriasoft/node-pg-client


langpavel:
Yeah, I should submit my PR for [`pg-async`](http://langpavel.github.io/node-pg-async/) as well :-)

@brianc: I suggest to create wiki page instead.

brianc:
Yeah I think moving this to a wiki page would be a good thing to do when I have time!  in the mean time i'll add it to the readme. :smile:

Issue #945, opened 2016-02-24
Cloning the library throws error when pg-native is not installed


janmeier:
Related to https://github.com/sequelize/sequelize/pull/5438 and https://github.com/sequelize/sequelize/issues/5476.

``` js
var pg = require('pg')
_.clone(pg);
```

Throws `Error: Cannot find module 'pg-native'`

Wanting to clone the library directly doesn't make much sense, but in the issues linked above, it happens because a sequelize model, which has a reference to pg is being deeply cloned. 

This should be fixable by switching to `defineProperty` and marking the `native` property as non-enumerable. 

I can write up a PR later today if you'd like - I just need a hint as to where to put the test

brianc:
Hey @janmeier - I would definitely appreciate a PR for this issue!  Sounds like a pretty annoying bug and also not a super difficult fix!  

As far as where to put the tests - if a test actually needs a postgres instance running I put it under `test/integration` otherwise under `test/unit` - either is fine really.  And for making a test...node-postgres was written initially before any of the now defacto test frameworks existed so I "wrote my own" in a sense.  Every single file under the `test/` directory is recursively required & is expected to run to completion & cleanly exit as fast as possible.  So you can just dump any old `whatever.js` file under the `test` directory & include a test using the standard `var assert = require('assert')` module from node.  As long as the js file exits cleanly (0 exit code) the test is considered passing!

Does that make sense?

basically:

```
$: touch test/unit/clone-pg-object-tests.js
$: vim test/unit/clone-pg-object-tests.js
# ... some code editing ensues
$ node test/unit/clone-pg-object-tests.js  <--- that should exit with a 0 status 
```

If you want to run all the unit tests (including the new one you added) you can type `make test-unit` from the top level dir.  If you want to run *all* the tests you can type `make test-integration` but you need a running instance of postgres.

Hope that helps!

janmeier:
Thanks for the thorough contribution guide! ;)

jibwa:
+1

PatrickKelly:
+1

TinOo512:
@janmeier any plan of opening a PR here with your fork ? I have the same issue and your fork seems to resolve it :)

janmeier:
@TinOo512 https://github.com/brianc/node-postgres/pull/950 has been merged instead

TinOo512:
[...]

PR #950, opened 2016-02-26
pg.native – return null if pg-native is missing

### LINKED ISSUE #838: when using babel, pg-native is always required (and fails if not present)
I'm not sure if this is a Babel or a node-postgres bug, steps to reproduce:

```
export PATH=$PATH:/tmp/iojs-v3.0.0-linux-x64/bin/:./node_modules/.bin
mkdir t
cd t
npm install babel pg

# create a t.js file with this ES6 content:
import * as pg from "pg";
console.log(pg);

# now execute: 
babel t.js > t-es5.js
node t-es5.js 

Error: Cannot find module 'pg-native'
    at Function.Module._resolveFilename (module.js:338:15)
    at Function.Module._load (module.js:280:25)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
    at Object.<anonymous> (/tmp/o/node_modules/pg/lib/native/index.js:1:76)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Module.require (module.js:364:17)
```

Babel generates the following code in `t-es5.js`:

```
function _interopRequireWildcard(obj) { if (obj && obj.__esModule) { return obj; } else { var newObj = {}; if (obj != null) { for (var key in obj) { if (Object.prototype.hasOwnProperty.call(obj, key)) newObj[key] = obj[key]; } } newObj["default"] = obj; return newObj; } }

var _pg = require("pg");

var pg = _interopRequireWildcard(_pg);

console.log(pg);
```

The exception is happening in the `_interopRequireWildcard(_pg)` call. This doesn't happens with other libraries, so I'm not sure if a Babel or a node-postgres bug.

langpavel:
`require('pg').native` will be `null` and report error once to `stdout` when `pg-native` is missing.

Fixes #838

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -75,7 +75,16 @@ if(typeof process.env.NODE_PG_FORCE_NATIVE != 'undefined') {
   //lazy require native module...the native module may not have installed
   module.exports.__defineGetter__("native", function() {
     delete module.exports.native;
-    module.exports.native = new PG(require('./native'));
-    return module.exports.native;
+    var native = null;
+    try {
+      native = new PG(require('./native'));
+    } catch (err) {
+      if (err.code !== 'MODULE_NOT_FOUND') {
+        throw err;
+      }
+      console.error(err.message);
+    }
+    module.exports.native = native;
+    return native;
   });
 }
--- a/test/native/missing-native.js
+++ b/test/native/missing-native.js
@@ -3,6 +3,4 @@
 
 var assert = require('assert');
 
-assert.throws(function() {
-  require('../../lib').native;
-});
+assert.equal(require('../../lib').native, null);


langpavel:
May be solution for #945 too. @janmeiner what you think?

janmeier:
It would solve the biggest part of the issue, but would still mean that an error message is logged to the console, without the user knowing why - "But I'm not accessing the `native` property directly, why do I see this message"

langpavel:
@janmeier Because something (babel in this case?) accessing `native` property.
Message can be removed and null returned, but this is not good solution.

Message is shown only once when `native` property is accessed first time.

Any suggestions welcomed!

anthonyjspoerl:
Is there a better solution for this being worked on? I am not using pg-native and am running into this issue in CI. I would prefer this solution over having to unnecessarily configure pg-native in CI.

langpavel:
@anthonyjspoerl There is alse #952 which hide property from Babel by descriptor.
I'm not sure how tools like `eslint-plugin-import` cope with hidden property, so return `null` and report warning only once when property is very firstly required is correct way.
I can add `process.env.`… switch to suppress warning

brianc:
Dang sorry I missed this!  I've been in the weeds with other things in my life for a while, getting out now finally. 💃 

I'll go through the PRs and see which ones break backwards compatibility, clump them up, merge them, and push a new major version.  I consider this a backwards break since someone somewhere might be depending weirdly on it throwing an exception.

vitaly-t:
All libraries that rely on `node-postgres` to detect and throw when `pg-native` isn't found would have to be modified, if we allow this change.

At the moment, the exception thrown is the only pointer that `pg-native` is missing as a dependency.

gaastonsr:
Bump!

Using knex.js and Node 6 and receiving error.

Advice for a patch in the mean time?

langpavel:
I think this is good choice. Thanks!

PR #952, opened 2016-02-28
Mark the native property as non-enumerable. Closes #945

### LINKED ISSUE #945: Cloning the library throws error when pg-native is not installed

janmeier:


janmeier:
@brianc Tests pass now :)

brianc:
Very nice!  I think this probably could break backwards compatibility in some weird edge cases - what do you think?  I think it should be a major version bump.

janmeier:
Not sure I can think of any cases right now - Would users every purposefully enumerate the object? The only cases I can think of are when the pg object is cloned as a child of something else

langpavel:
@janmeier users brobably don't enumerate but babel or webpack can. This was debated in #950

brianc:
Yeah sorry for dropping the ball on this one...

I've been busy w/ life stuff.  I see there is [this pr](https://github.com/brianc/node-postgres/pull/950) as well.  Do you want me to unify the changes between these two PRs & submit a new PR that does both of these in combo?  I think both changes make sense & fix different parts of the issue.

janmeier:
@brianc Yeah that would be nice! :)

brianc:
I believe this is fixed in 5.1

PR #954, opened 2016-02-29
Catch errors thrown by type parsers and emit error

aheuermann:
Errors thrown by type parsers are not currently caught. This catches them and emits an error.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -160,7 +160,7 @@ Client.prototype.connect = function(callback) {
     self.readyForQuery = true;
     self._pulseQueryQueue();
     if(activeQuery) {
-      activeQuery.handleReadyForQuery();
+      activeQuery.handleReadyForQuery(con);
     }
   });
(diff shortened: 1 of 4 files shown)


aheuermann:
Looks like I need to update `pg-native` too.

brianc:
I appreciate the work here, but I think the best place for this logic would likely be inside the type parsers themselves and not here - most type parsers don't throw under any circumstance...if you need special try/catch in a type parser best approach is probably to override the type parser like you've done here in your tests & try/catch there.

felixfbecker:
Please note though that `query()` is an asynchronous function, and async functions should always either always throw synchronously or always throw asynchronously (emit an error event, callback with error, reject). Everything else leads to very hard to find bugs ([commonly referred to as "releasing Zalgo"](http://blog.izs.me/post/59142742143/designing-apis-for-asynchrony))

Issue #949, opened 2016-02-26
crash due to null activeQuery in lib/client.js


spollack:
We had a crash in production last night in the pg module. (we just updated to pg 4.5.1 javascript, after running pg 2.11.1 native for many many months. this is the first time i've seen this issue.)

The crash was in lib/client.js:123:
```
  self.activeQuery.handleCommandComplete(msg, con);
```
stack trace:
```
TypeError: Cannot call method 'handleCommandComplete' of null
    at null.<anonymous> (/app/node_modules/pg/lib/client.js:123:24)
    at emit (events.js:95:17)
    at Socket.<anonymous> (/app/node_modules/pg/lib/connection.js:109:12)
    at Socket.emit (events.js:95:17)
    at Socket.<anonymous> (_stream_readable.js:765:14)
    at Socket.emit (events.js:92:17)
    at emitReadable_ (_stream_readable.js:427:10)
    at emitReadable (_stream_readable.js:423:5)
    at readableAddChunk (_stream_readable.js:166:9)
    at Socket.Readable.push (_stream_readable.js:128:10)
    at TCP.onread (net.js:529:21)
```

i do not have a repro case at this time. 

obviously we could put guards in this function (and its peers) to check if self.activeQuery exists before calling the function. but not sure how we got in this state in the first place. @brianc thoughts on this? thanks

vitaly-t:
In cases like this one, it is worth trying to roll back specifically node-postgres version, to see whether the module itself caused the issue or one of its updated dependencies, which I reckon was the latter, by the look of the error log.

But then again, if you cannot reproduce the issue, your guess is as good as anyone's. You won't be able to fix it without knowing how to check that it's gone.

Figure out how to reproduce the issue and get back ;)

spollack:
Hi Vitaly, not sure i follow, the only code on the stack when the crash happens is node internals and the pg library itself, no other dependency libraries. That of course doesn't mean that a change to a pg dependency didn't lead to us getting in this state, but i'm curious what in the error log leads you to believe that? thanks.

spollack:
We have now had about 10 crashes in production due to this issue. Last week we made a local patch to pg to add guards as suggested above, and that resolved the issue. We will open a PR here for that patch. cc @jshepard

spollack:
@jshepard dug into this more, and it appears that this problem was likely caused by an issue in another library we were using on top of pg. given that, closing this.

brianc:
Glad you got it figured out!!! :dancer:

PR #961, opened 2016-03-07
add check before accessing activeQuery

spollack:
addresses #949

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -100,44 +100,58 @@ Client.prototype.connect = function(callback) {
 
     //delegate rowDescription to active query
     con.on('rowDescription', function(msg) {
-      self.activeQuery.handleRowDescription(msg);
+      if (self.activeQuery) {
+        self.activeQuery.handleRowDescription(msg);
+      }
     });
 
     //delegate dataRow to active query
     con.on('dataRow', function(msg) {
-      self.activeQuery.handleDataRow(msg);
+      if (self.activeQuery) {
+        self.activeQuery.handleDataRow(msg);
+      }
     });
 
     //delegate portalSuspended to active query
     con.on('portalSuspended', function(msg) {
-      self.activeQuery.handlePortalSuspended(con);
+      if (self.activeQuery) {
+        self.activeQuery.handlePortalSuspended(con);
+      }
     });
 
     //deletagate emptyQuery to active query
     con.on('emptyQuery', function(msg) {
-      self.activeQuery.handleEmptyQuery(con);
+      if (self.activeQuery) {
+        self.activeQuery.handleEmptyQuery(con);
+      }
     });
 
     //delegate commandComplete to active query
     con.on('commandComplete', function(msg) {
-      self.activeQuery.handleCommandComplete(msg, con);
+      if (self.activeQuery) {
+        self.activeQuery.handleCommandComplete(msg, con);
+      }
     });
 
     //if a prepared statement has a name and properly parses
     //we track that its already been executed so we don't parse
     //it again on the same client
     con.on('parseComplete', function(msg) {
-      if(self.activeQuery.name) {
+      if(self.activeQuery && self.activeQuery.name) {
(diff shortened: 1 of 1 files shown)


brianc:
@spollack I'm so sorry you were having crashes in production after upgrading. :frowning: never ever ever fun to deal with.

I want to merge this as soon as I possibly can & get this pushed as a new patch version.  Before I do that could I trouble you for a bit more information on how this happens?  Do you have any steps to reproduce it?  I can't fathom a time the activeQuery would go null before the query response comes in - and all the unit tests are still passing so...it's strange.  Ultimately I'd like to get this merged with a failing test case that passes after the code changes, because my guess is if this happens and the null check bypasses any message handling in the client without an active query your client `query end callbacks` are never going to fire - and I would *guess* your app would eventually hang waiting for callbacks which never complete.

spollack:
@brianc yes i totally agree, i'd like to understand this better too. i've tried a bit today to repro this without luck so far. Let me do some more digging.

spollack:
I reviewed the logs from all 12 cases in production where this crashed on us (pretty much all while i was away on vacation last week, doah!). In every case, it was correlated with a query being cancelled on the backend due to a statement timeout. however, the inverse was not always true -- many other statements that were cancelled due to statement timeout did not result in a crash. So it is a more complex interaction. I also tried writing a simple integration test that sets the statement timeout, then does pg_sleep to hit it, and (not surprisingly) this didn't repro it by itself. So, no smoking gun yet, but @brianc i'm wondering if that triggers any ideas for you on how we could get into this state.

spollack:
starting at the pg code, i see something suspicious. look here https://github.com/brianc/node-postgres/blob/master/lib/client.js#L170 and you'll see that we do:
```
self.activeQuery = null;
return activeQuery.handleError(error, con);
```
i.e. we null out activeQuery before doing handleError. inside handleError, we potentially do more work on the connection here https://github.com/brianc/node-postgres/blob/master/lib/query.js#L97 
```
connection.sync();
```
and perhaps this would cause the command complete handler to then fire, which is where we see the crash here https://github.com/brianc/node-postgres/blob/master/lib/client.js#L123
```
self.activeQuery.handleCommandComplete(msg, con);
```
@brianc thoughts? i don't know the insides of pg nearly as well as i would like to. Thanks!

spollack:
another possibility here is that it is a race coming from the postgres side -- that we could sometimes be receiving a commandComplete message on the connection after receiving an error message. @brianc thoughts?

brianc:
@spollack sorry for taking a while to get back to you - been really busy w/ life stuff.


> starting at the pg code, i see something suspicious. look here https://github.com/brianc/node-postgres/blob/master/lib/client.js#L170 and you'll see that we do:
> 
> self.activeQuery = null;
> return activeQuery.handleError(error, con);
> i.e. we null out activeQuery before doing handleError. inside handleError, we potentially do more work on the connection here https://github.com/brianc/node-postgres/blob/master/lib/query.js#L97
> 
> connection.sync();
> and perhaps this would cause the command complete handler to then fire, which is where we see the crash here https://github.com/brianc/node-postgres/blob/master/lib/client.js#L123
> 
> self.activeQuery.handleCommandComplete(msg, con);
> @brianc thoughts? i don't know the insides of pg nearly as well as i would like to. Thanks!

I think it might make sense to null out the active query *after* calling `activeQuery.handleError`...like so:

```js
var result = activeQuery.handleError(error, con);
self.activeQuery = null;
return result;
```

Because you're right...there might be more things going on in there? (unlikely, but worth checking) - have you tried just this modification to the code & see if it fixes your issue?

spollack:
Thanks Brian. we have not tried just that modification. I wish we had a clear repro case! We have been running with this PR in production for several weeks now, and have not had this particular issue anymore.

spollack:
@jshepard dug into this more, and it appears that this problem was likely caused by an issue in another library we were using on top of pg. given that, closing this.

PR #963, opened 2016-03-10
Fix byte representation of CancelRequest message

dorianj:
I noticed that query cancellation was not working when connecting through pgbouncer,
even though it worked fine when directly connected. This is because we're appending an
extra null byte, and pgbouncer is strict about the packet length.
(per http://www.postgresql.org/docs/9.1/static/protocol-message-formats.html)

This removes the extraneous byte, which fixes cancellation against pgbouncer.


brianc:
Ah dang nice catch!

PR #971, opened 2016-03-23
readme: fix pool size

LinusU:


brianc:
Thanks and good catch!

Issue #981, opened 2016-04-07
Native Bindings override issue


vitaly-t:
The way Native Bindings currently work, is by overriding the library's global settings:
```js
    module.exports.__defineGetter__("native", function () {
        delete module.exports.native;
        module.exports.native = new PG(require('./native'));
        return module.exports.native;
    });
```
As a result, it creates a conflict when you try using two third-party libraries - 
* one that uses `pg` via Native Bindings
* one that uses `pg` via JavaScript bindings

Just as one library activates Native Bindings, it becomes global, due to Node.js module sharing principle, so it overrides this setting for libraries that do not support Native Bindings.

### How to test it

Install `pg-native`, and declare:
```js
var pg1 = require('pg');
var pg2 = require('pg').native;
```
Just as `pg2` is initialized, any call into `pg1` for `connect` or `query` goes into the layer for the Native Bindings, thus breaking any module that supports only JavaScript bindings.

### Thoughts

I don't know yet how difficult it would be to fix this, or even  whether it is possible to make Native Bindings work in parallel with JavaScript Bindings within the same process, without running into conflicts. Any thoughts are welcome!

vitaly-t:
Below is a complete example replicating the issue:
```js
var pg1 = require('pg');
//var pg2 = require('pg').native;

var cn = "pg://postgres@localhost:5000/test"; // invalid connection

pg1.connect(cn, function (err) {
    return console.error(err);
});
```
Outputs connection error from JavaScript binding.
```
{ [Error: connect ECONNREFUSED 127.0.0.1:5000]
  code: 'ECONNREFUSED',
  errno: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 5000 }
```

Now let's un-comment that second require:
```js
var pg1 = require('pg');
//var pg2 = require('pg').native;

var cn = "pg://postgres@localhost:5000/test"; // invalid connection

pg1.connect(cn, function (err) {
    return console.error(err);
});
```
Now suddenly the error arrives formatted according to the Native Bindings:
```
error fetching client from pool { [Error: connect ECONNREFUSED 127.0.0.1:5430]
  code: 'ECONNREFUSED',
  errno: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 5430 }
```
even though we are still using `pg1`, as the second `require` overrides global binding settings, so we no longer have access to the JavaScript bindings of the library.

[...]

PR #984, opened 2016-04-08
Requiring native bindings polutes 'global'

brianc:
There was some nasty global-ish variable reference updating happening when the native module 'initializes' after its require with `require('pg').native`

This fixes the issue by making sure both `require('pg')` and `require('pg').native` each initialize their own context in isolation and no weird global-ish references are used & subsequently stomped on.

### DIFF
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -3,93 +3,95 @@ var EventEmitter = require('events').EventEmitter;
 var defaults = require('./defaults');
 var genericPool = require('generic-pool');
 
-var pools = {
-  //dictionary of all key:pool pairs
-  all: {},
-  //reference to the client constructor - can override in tests or for require('pg').native
-  Client: require('./client'),
-  getOrCreate: function(clientConfig) {
-    clientConfig = clientConfig || {};
-    var name = JSON.stringify(clientConfig);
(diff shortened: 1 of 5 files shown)


brianc:
This fixes #981 - nice find by @vitaly-t! Thanks!

lib/pool.js
```diff
@@ -3,93 +3,95 @@ var EventEmitter = require('events').EventEmitter;
 var defaults = require('./defaults');
 var genericPool = require('generic-pool');
 
-var pools = {
-  //dictionary of all key:pool pairs
-  all: {},
-  //reference to the client constructor - can override in tests or for require('pg').native
-  Client: require('./client'),
```
brianc: I was 'punching over' this which is bad since this is a singleton.  I made it not a singleton since there's no reason for it to be one, and I pass the proper constructor (native or JS) when its initialized.

brianc:
fixed - pushing new patch version to npm now

martinkuba:
@brianc This change is breaking the newrelic instrumentation.  We used to be able to wrap the client instance used by the pool, and now it's not accessible.  Can we expose the client constructor on pools again?  Or, can you please advise how to get around this?

PR #992, opened 2016-04-14
don't mutate params when preparing statement

### LINKED ISSUE #750: client.query(sqlString, params[], function) alters the given params[]
When supplying a parameters array to client.query the given array is modified (in my case it originally was an array of numbers and got transformed to an array of strings). This is not correct behaviour, I see the problem in the source code (query.js, line 143):

```
self.values[i] = utils.prepareValue(self.values[i]);
```

If transformation is needed it should be done on a clone of the passed array so that it doesn't break any logic in case the original array is reused somewhere else after the query.

I think a quick fix would be to change (query.js, line 14) from:

```
this.values = config.values;
```

to

```
this.values = config.values.slice();
```

devTristan:
Fixes #750

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -146,11 +146,8 @@ Query.prototype.prepare = function(connection) {
     }, true);
   }
 
-  //TODO is there some better way to prepare values for the database?
   if(self.values) {
-    for(var i = 0, len = self.values.length; i < len; i++) {
-      self.values[i] = utils.prepareValue(self.values[i]);
-    }
+    self.values = self.values.map(utils.prepareValue);
   }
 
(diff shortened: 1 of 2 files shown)


langpavel:
:+1:

seamusabshere:
@devTristan can you fix the test on here?

devTristan:
@seamusabshere I haven't the foggiest idea why it's failing on that one postgres+node version combination. It might be transient or something stranger. It's on my TODO list.

brianc:
Ah it's failing due to some stupid race condition in an error checking test I haven't fixed yet.  I'll merge this & look at that test later.  Thanks a lot for putting this together!

Issue #996, opened 2016-04-19
error: invalid input syntax for integer: "123.456"


devTristan:
numeric arguments in prepared statements are sent to postgres as [strings](https://github.com/brianc/node-postgres/blob/fe6f32a6e7e62b276cda100e2a82af755500bb42/lib/utils.js#L51). I'm not sure why this is important, but it causes the following case to fail:

```js
client.query('SELECT * FROM numbers WHERE n > $1', [123.456])
```

where `numbers.n` is of type `int`, `bigint`, etc. Since `123.456` is sent as a string, postgres interprets this like `WHERE n > '123.456'::int` which fails with the following error:

```
error: invalid input syntax for integer: "123.456"
```

Preparing the query manually, as follows, works without a hitch:

```js
client.query('SELECT * FROM numbers WHERE n > 123.456')
```

A test case is incoming.

vitaly-t:
Use `SELECT * FROM numbers WHERE n > $1::decimal` in the query, and you will be fine.

joskuijpers:
What vitaly set. Also, an integer does not have a decimal value. An integer is a whole value, like 1 or 123.

Sidenote: I _always_ use the `::type` on input values to have some error checking.

@brianc Also not a pg problem. Never had any troubles myself.

brianc:
:godmode:

PR #997, opened 2016-04-19
Add test case for issue #996

devTristan:
test fails


seamusabshere:
@devTristan why not merge this into #996 ?

devTristan:
> 996-tests.js

I wanted to name my test file based on the issue number which I couldn't get until making the issue. Sure, I could've chosen [latest issue number + 1], but that introduces a race condition. ¯\_(ツ)_/¯

charmander:
The related issue appears to be closed as working as intended.

PR #998, opened 2016-04-20
enable instrumentation of Client used in pool

martinkuba:
This commit broke the newrelic instrumentation since the Client constructor used in pool is now hidden and cannot be instrumented.
https://github.com/brianc/node-postgres/commit/a8bd44a6ec89b941bcfcd2ef7f386eca34a2fdb4

This is a quick fix to unblock any users using pg native and pooling with newrelic.  I am open to working on a better way to instrument as a follow up, if this is not optimal.

### DIFF
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -6,6 +6,7 @@ var genericPool = require('generic-pool');
 
 module.exports = function(Client) {
   var pools = {
+    Client: Client,
     //dictionary of all key:pool pairs
     all: {},
     //reference to the client constructor - can override in tests or for require('pg').native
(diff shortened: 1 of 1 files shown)


brianc:
I know what you're saying - I'm really sorry that the newrelic instrumentation was broken!  I'll merge this for now.  Would you be able to put together a unit test exercising this functionality so I don't break it again in the future?  A separate PR for the test is fine. :)

I'll push a patch version out to npm today.

martinkuba:
@brianc Yes, I will open a new PR for adding a test.  Thanks for merging this in!

PR #1002, opened 2016-04-27
Change test matrix in .travis.yml

brianc:
Add tests for node @ v6. Remove node & postgres test permutations for older versions of node.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,37 +16,25 @@ matrix:
   include:
     - node_js: "0.10"
       addons:
-        postgresql: "9.1"
-    - node_js: "0.10"
-      addons:
-        postgresql: "9.2"
-    - node_js: "0.10"
-      addons:
-        postgresql: "9.3"
-    - node_js: "0.10"
-      addons:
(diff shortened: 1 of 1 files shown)


vitaly-t:
You really should Node.js versions as a single number everywhere: `node_js: "4"`, `node_js: "5"`, `node_js: "6"`. That will guarantee the use of the latest sub-version.

brianc:
Ah that's good to know, thanks. I'll do that.

vitaly-t:
You missed one for the 4 ;) Current is 4.4, and you are testing under 4.1 ;)

vitaly-t:
Cheers! Once you're all done, please release the new one as 4.6.0 ;) So we will know it is the one that's supposed to work with Node.js .6.x properly ;)

brianc:
4.5.5 is supposed to work with node 6 properly - fixing a bug with how buffers were used does not warrant a new minor version bump & the minor version of this library never has indicated in any way which versions of node it was compatible with.

PR #1007, opened 2016-05-02
Adding returnToHead in clientConfig param

twisha16:
Making returnToHead param for creating a Pool() configurable https://github.com/coopernurse/node-pool
Right now it defaults to false and there is no way to override it.


brianc:
+1 looks good! Thanks!

PR #1011, opened 2016-05-05
don't return clients that have errored to the connection pool

calebTomlinson:
The test for this one mostly speaks for itself.

We're using some constructs to make sure connections are always returned to the pool. Unfortunately after a client has errored, we still call `done()`. This was returning bad connections to the pool, that would error when we tried to use them. We've fixed our code to call `done(err)` if a connection has errored, but it's complicated and is uglying up our code ;). I believe the library should behave intelligently for this case.

### DIFF
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -84,6 +84,8 @@ module.exports = function(Client) {
           cb(null, client, function(err) {
             if(err) {
               pool.destroy(client);
+            } else if (client._destroying) {
+              return;
             } else {
               pool.release(client);
             }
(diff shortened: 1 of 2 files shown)


calebTomlinson:
Is there any feedback that would help this get merged, or an opinion on it?

brianc:
Sorry I missed this - I don't actually understand how the code fixes the issue you're describing - where does `_destroying` get set on the client? Also, the test doesn't ever call `done()` so I don't understand how the test is checking for the expected behavior either.

brianc:
I've changed things a bit with the release of [pg-pool](https://github.com/brianc/node-pg-pool) in release `pg@6.0` - if this problem still exists it's a pool problem & we should probably restart the discussion over there.

PR #1017, opened 2016-05-12
new Buffer(str) is deprecated, use Buffer.from(str) instead

ignitenet-martynas:
https://nodejs.org/dist/latest-v6.x/docs/api/buffer.html#buffer_new_buffer_str_encoding says new Buffer(str) is deprecated and Buffer.from(str) should be used instead. Sorry I missed that in my first pull request.


brianc:
I appreciate the thought here but this breaks backwards compatibility with a lot of versions of node.  If `Buffer(str)` is fully removed *or* there is a suitable alternative which is not deprecated which works on all node environments I'll happily merge something, but breaking compat w/ so many versions of node is a no-go right now.

PR #1022, opened 2016-05-16
Proper error message with undefined where options

mdrmuhaimin:
When user provides a knex select statement with an undefined options in where clause it is not properly handled an give an ambiguous error message telling `Unhandled rejection TypeError: Cannot read property 'toString' of undefined.` This PR will helpful to users at it will tell them the exact problem.


brianc:
Nice! Thanks!

PR #1025, opened 2016-05-19
added .npmignore in order to not publish tests to npm

henriksjostrom:
Added an .npmignore so that pg doesn't publish the test code to npm, thereby shrinking the dependency size for anyone using it.


brianc:
Thank you!

PR #1030, opened 2016-05-28
client.end: add .onDone option

juliangruber:
I was hitting a code path where I needed to check the "private" property `.activeQuery` in order to properly end the client. This patch adds a `.onDone` option to `client#end()` that only ends when it's not busy.

Context: I was using `client.on('drain', () => client.end())` before, but the problem with that is if you have that as the last line of a program that isn't supposed to run forever, like a server, and if the there's nothing in the client's queue, there will never be a `drain` event fired.

I didn't add tests or docs as I wasn't sure whether you're ok with this api and approach.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -327,8 +327,14 @@ Client.prototype.query = function(config, values, callback) {
   return query;
 };
 
-Client.prototype.end = function() {
-  this.connection.end();
+Client.prototype.end = function(config) {
+  if (config && config.onDone && this.activeQuery) {
+    this.once('drain', function() {
+      this.end();
+    });
+  } else {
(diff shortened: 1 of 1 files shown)


ekristen:
+1

brianc:
Thanks for putting this PR together - would you mind writing some test for it?

brianc:
Closing this because it's gone stale & I never received tests.  I apologize for not being more supportive before - sometimes I get pretty busy & so my attention levels vary.  I've recently (as of last week) quit my job so I can focus a *lot* more of my time here.  If you'd like to re-open I'll give it another go, but for now closing so I can prep for 7.0.

PR #1031, opened 2016-05-29
use latest generic-pool

devTristan:
generic-pool has had a few bug fixes since 2.1.1 and it'd be good to get those in


brianc:
You got it! Thanks for putting this together & sorry I let this one sit out a bit - I'll merge this & push a new patch version out asap.

devTristan:
Thanks! You're a life saver, @brianc.

brianc:
💃

Issue #848, opened 2015-08-31
Connection using SSL even though I have SSL set to false in options


ym-mqual:
Stems from line 44 in the connection-parameters.js.  I corrected it as follows:

OLD: config.ssl || useSsl(); -this is a problem if you send in false.
CORRECTED: this.ssl = typeof config.ssl === 'undefined' ? useSsl() : config.ssl;

ym-mqual:
Actually thought about it again and this is probably better:

this.ssl = typeof config.ssl === 'boolean' ?  config.ssl : useSsl();

simonv3:
:+1: I can confirm that this is an issue, and that fixes it.

joskuijpers:
Can you submit a Pull Request?

joskuijpers:
@brianc This can be closed now

brianc:
@joskuijpers you are so helpful! Thank you!!

PR #1055, opened 2016-06-20
Fix SSL configuration error and add tests. #848

joskuijpers:
Fixes a small configuration bug. When the user explicitly passes `false`, it is still overridden by the environment. This fixes that error.


brianc:
💃  fantastic! thank you!!

PR #1063, opened 2016-06-23
Update README.md

unkani:
Fixed minor typo "exampe" -> "example"


brianc:
Thanks for that!! I don't have any spell checking in my editor, and sometimes things like that slip through! 😬

PR #1068, opened 2016-07-01
Accept connection parameters in Pool constructor

daliwali:
It seems that `Pool` doesn't accept connection parameters like `connect` does, this fixes that.

### DIFF
--- a/lib/pool-factory.js
+++ b/lib/pool-factory.js
@@ -1,11 +1,13 @@
 var Client = require('./client');
+var ConnectionParameters = require('./connection-parameters');
 var util = require('util');
 var Pool = require('pg-pool');
 
 module.exports = function(Client) {
 
   var BoundPool = function(options) {
-    var config = { Client: Client };
+    var config = new ConnectionParameters(options);
+    config.Client = Client;
     for (var key in options) {
       config[key] = options[key];
     }
--- a/lib/pool-factory.js
+++ b/lib/pool-factory.js
@@ -8,9 +8,6 @@ module.exports = function(Client) {
   var BoundPool = function(options) {
     var config = new ConnectionParameters(options);
     config.Client = Client;
-    for (var key in options) {
-      config[key] = options[key];
-    }
     Pool.call(this, config);
   };
 


0x8890:
Test is failing due to a temporal error, I think.

brianc:
Hey @0x8890 thanks for the PR, I really appreciate the work! Question: is there a particular use case that is disabled by not using connection parameters? I'd like to understand a bit better what's missing & hopefully get a test up around it so a regression is not introduced in another change in the future.

0x8890:
@brianc yes, I would like to use a connection string instead of a configuration object to instantiate a Pool. There does not seem to be any public API for parsing a connection string exposed by this library, and it would also be convenient to accept a connection string here.

I have written an [interface](https://github.com/fortunejs/fortune-postgres) that makes use of the connection string, and currently it's not possible to use a Pool with a connection string, as noted [here](https://github.com/fortunejs/fortune-postgres/issues/14).

I could also put together a test case for this if you want.

Voles:
@brianc I think in general it makes sense to be able to pass a connection string. I've used this approach as well for other applications.

allspiritseve:
@0x8890 if all you're wanting to do is pass a connection string to `pg.Pool`, I believe you can already do so as options passed when creating a pool are [used to create clients](https://github.com/brianc/node-pg-pool/blob/master/index.js#L33) and `pg.Client` [accepts a `connectionString` parameter](https://github.com/brianc/node-postgres/blob/master/lib/connection-parameters.js#L49-L51).

@brianc please correct me if I'm wrong.

AlJohri:
👍

vvo:
```js
const db = new PgPool({
  connectionString: process.env.DATABASE_URL,
});
```

Works, this is a documentation issue

brianc:
I'd like to add connection string support back in to pg@7.0 in some form.  It gets weird because you're configuring the pool and the client in a single string, but I think we can make it work in a way that makes everyone more productive & happy, which is the point!  I'll add this to the 7.0 milestone.

vvo:
@brianc Awesome that you are working on a 7.0, any way we can see the planned "roadmap"? thanks

brianc:
@vvo - you bet! [here you go](https://github.com/brianc/node-postgres/milestone/3) right now just organizing issues into there, and I'll be adding more.  First I want to overhaul the documentation to be better - it's old & a lot of issues I get are opened that should be answered in docs. I also get a lot of word of mouth reports about things that have been confusing in the past.  Please feel free to open any issues around things you'd like to see added/changed in 7.0 as well as anything you think is missing from the docs - I'll add 'em to the appropriate milestones & work on them.

brianc:
as @vvo pointed out you can pass a connection string as the `connectionString` parameter to the pool.  I just added a regression test to this repo [here](https://github.com/brianc/node-postgres/pull/1310) to make sure this remains covered forevermore.  This makes the most sense because the pool also accepts other config parameters that don't apply to the client so passing the string as one of many options is the way to go.  I've documented this [here](https://node-postgres.com/features/connecting#connection-uri).

PR #1072, opened 2016-07-09
changed for self signed ssl support

guoxiangyang:
support ssl connect to postgresql server with server cert and client cert & key file


brianc:
Hey sorry for taking a bit to get to this! I appreciate the change! I'm going to merge this & push out a new patch version right now, and I'm going to move your documentation to the wiki, because I try to keep some docs there to they can be edited w/o pull requests.  Thanks a million!

PR #1076, opened 2016-07-12
Simplify conditions

charmander:
Not very important, might be counter to the preferred style, but may as well ask?

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -257,7 +257,7 @@ Connection.prototype.bind = function(config, more) {
   buffer.addInt16(len);
   for(var i = 0; i < len; i++) {
     var val = values[i];
-    if(val === null || typeof val === "undefined") {
+    if (val == null) {
       buffer.addInt32(-1);
     } else if (val instanceof Buffer) {
       buffer.addInt32(val.length);
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,5 +1,6 @@
 {
   "trailing": true,
   "indent": 2,
+  "eqnull": true,
   "evil": true
 }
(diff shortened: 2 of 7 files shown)


lib/connection.js
```diff
@@ -257,7 +257,7 @@ Connection.prototype.bind = function(config, more) {
   buffer.addInt16(len);
   for(var i = 0; i < len; i++) {
     var val = values[i];
-    if(val === null || typeof val === "undefined") {
+    if (val == null) {
```
brianc: I think this will cause '0', '', and other falsy values to be coerced into null used as a parameter? 🙅

brianc:
Hey @charmander I appreciate the thought & effort, but I'm reticent to change any of the if conditions comparing strictly to `null` and `undefined` to a more lax double equals falsy check. That can subtly change semantics & behavior that might not be properly covered in tests, and start causing weird unexpected breakages.  I think the CBA on this change is pointing towards potentially higher negative impact & cost than the minor stylistic changes.

charmander:
`x == null` is different from typical `==` checks in that it’s *exactly* the same as `x === null || x === undefined`, which is why JSHint has an option to allow it specifically. But it is a style choice, so I can take those changes out if you want.

lib/connection.js
charmander: It won’t, don’t worry. (Putting it on the line note too for visibility.)

brianc:
I appreciate the thought here, but I'm gonna close this as it doesn't really add anything specific.  There are a lot of open bugs that need my attention & doing logic refactorings is additional overhead I unfortuantely don't have cycles for right now!  If you'd like to pop some issues off the list & PR for those that'd be greatly appreciated!  Any logic fixed during those w/ supporting test coverage ofcourse is totally fine. 😄

PR #1082, opened 2016-07-19
node-postgres blocks with faulty SQL

brianc:
I'm using ES6 with async/await and node-postgres with connection pooling.

Let there is a table with a column which has UNIQUE constraint and a primary key column, the following code block blocks on `client.query`:

````
export async function someFunction () {
    const client = Postgres.connect()

    try {
        //  First insertion
        const result = await client.query('INSERT INTO table (column) VALUES ($1::text) RETURNING *', ['test'])

        //  Second insertion, which will fail
        const secondResult = await client.query('INSERT INTO table (column) VALUES ($1::text) RETURNING *', ['test'])

        console.log(secondResult.rows)    //  never called
    } catch (err) {
        console.error(err)
    } finally {
        client.release()
    }
}
````

However, I'm expecting to see this in somewhere:

     ERROR:  duplicate key value violates unique constraint "table_column_key"
     DETAIL:  Key (column)=(test) already exists.

Am I doing something wrong?

### DIFF
--- a/test/integration/connection-pool/yield-support-body.js
+++ b/test/integration/connection-pool/yield-support-body.js
@@ -1,0 +1,28 @@
+var helper = require('./test-helper')
+var co = require('co')
+
+var tid = setTimeout(function() {
+  throw new Error('Tests did not complete in tme')
+}, 1000)
+
+co(function * () {
+  var client = yield helper.pg.connect()
+  var res = yield client.query('SELECT $1::text as name', ['foo'])
+  assert.equal(res.rows[0].name, 'foo')
+
+  var threw = false
+  try {
+    yield client.query('SELECT LKDSJDSLKFJ')
+  } catch(e) {
+    threw = true
+  }
+  assert(threw)
+  client.release()
+  helper.pg.end()
+  clearTimeout(tid)
+})
+.catch(function(e) {
+  setImmediate(function() {
+    throw e
+  })
+})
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by Chatatata** [comment]
I'm using babel with async-to-bluebird-promises transformation module.

Buğra Ekuklu

Github, Bitbucket, Stack Overflow: @Chatatata

SO Ready To Help

Please information if this is considered spam.

19 Tem 2016 tarihinde 17:19 saatinde, Brian C <notifications@github.com> şunları yazdı:

**Comment by brianc** [comment]
@brikou - I've pushed a fix up here; once the tests pass I'll merge & push out a new patch version - thanks for reporting this and the information you supplied in helping me identify the bug! 💃

**Comment by brikou** [comment]
@brianc Thank you, can't wait for the release :)

File: test/integration/connection-pool/yield-support-body.js [hunk 1]
```diff
@@ -0,0 +1,28 @@
+var helper = require('./test-helper')
+var co = require('co')
+
+var tid = setTimeout(function() {
+  throw new Error('Tests did not complete in tme')
```
brikou: typo here

brianc: ah thanks - i'll fix on master when I push the new version.

brianc:
Published version with fix: `pg@6.0.3`

PR #1094, opened 2016-07-26
Add badges for npm version and number of dl on npm

abenhamdine:
Adding those two badges in the Readme for the following reasons  : 
- having a badge for npm version allow to see at a glance the current version
- the badge for dl/month allow to see immediatly that this module is very used in the ecosystem, which is an useful information when you are a newcomer to node and databases drivers.


brianc:
Merci @abenhamdine!  Greatly appreciated!! 💃 🍔

PR #1096, opened 2016-07-26
Add details about differences between js and native bindings

abenhamdine:
For a new comer, native bindings seems to be a drop in replacement, but there's some sligthly differences between the two API, especially regarding the error object .
Theses differences have raised several questions and issues.
Thus, providing some details about these actual differences should avoid some difficulties for the user who wants to use the native bindings.

I'm not a native english speaker :disappointed: so feel free to reformulate the text of the PR, I won't be upset :wink: .


brianc:
Thanks for the documentation updates @abenhamdine! Always appreciated!

PR #1098, opened 2016-07-30
pgpass: Upgrade to 1.x

hoegaarden:
pgpass is using semver versioning now, thus a dependency on
version 1.x should be safe.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "pg-connection-string": "0.1.3",
     "pg-pool": "1.*",
     "pg-types": "1.*",
-    "pgpass": "0.0.6",
+    "pgpass": "1.x",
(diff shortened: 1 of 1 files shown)


mthrok:
@brianc 
I am not sure if you are aware of the context of this update, so let me explain it here.
In the latest `pgpass`, file handler issue in connection pool, (`.pgpass` files are left open forever and will consume all the file handler allowed by OS) is fixed.
This PR is to incorporate it in `node-postgres`.

I really appreciate if you can merge and release this.
Thanks

brianc:
Sorry for missing this.  I appreciate the heads up from y'all!  Merging & pushing a new patch version now!

mthrok:
Thanks!

PR #1107, opened 2016-08-11
Update example: client.end() does not accept callback parameter

pihvi:
No callback: https://github.com/brianc/node-postgres/blob/master/lib/client.js#L339


brianc:
Thanks for the doc updates!  Always appreciated!  In this case I went ahead & fixed the bug so the current docs are actually right now....still appreciate the effort though! 💃

pihvi:
Well that was fast! And a better outcome :) Thanks!

PR #1110, opened 2016-08-13
Multiple-statement Support

abacigil:
# Adjustment of result object for multiple statements

```
    - If you query multiple statements, current lib does not support and it returns wrong results in result object.
- Now it supports multiple statements and it returns "result" object as an array and this array has Result object for each statement that requested in query.
- Example for multiple statement : SELECT * FROM users;select * from events;update users set id = 1 where id = 1
- If you set "multipleStatementResult" to "true" in connection configuration object, this feature will be enabled, otherwise nothing will change, same functionality of         original one will be executed.
- Tested with SELECT, UPDATE, DELETE and INSERT
```
# Simple example

```
- An example for multiple statements is available in "sample" folder.
```

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -323,7 +323,7 @@ Client.prototype.copyTo = function (text) {
 Client.prototype.query = function(config, values, callback) {
   //can take in strings, config object or query object
   var query = (typeof config.submit == 'function') ? config :
-     new Query(config, values, callback);
+     new Query(config, values, callback,this.connectionParameters.multipleStatementResult);
   if(this.binary && !query.binary) {
     query.binary = true;
   }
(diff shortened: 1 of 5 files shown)


joskuijpers:
Hi! Thanks!
Could you add tests, fix the linting errors and remove all those changed lines where only a space was added? It makes the diff unreadable 😄 

Also, what if I have opened a connection with multi-result, but I only send a single query? What is the result then?

abacigil:
I'll try to add tests as soon as possible ( I'm busy recently ), and fix the lines.

It'll work as usual and it'll return "result" object as an array with just 1 response for executed query.

shilch:
Any updates on this PR?

rightaway:
I'm happy to test and provide feedback.

charmander:
This was implemented differently in pg 7: a query that produces multiple result sets resolves with an array.

PR #1120, opened 2016-09-01
Add Flow Type definition.

iamchenxin:
Using .js.flow. 
Flow will auto pickup `.js.flow` as a type def, when it appears in packages.

This is cooperated with [node-pg-pool PR#26](https://github.com/brianc/node-pg-pool/pull/26)  type def.

---

I am new to `Postgresql` and `pg`. So maybe there are some misunderstanding in my type def.
In this type def, i hide those function be used  internally.
And only exposed the public API.(Did i miss some API?)

### DIFF
--- a/lib/index.js.flow
+++ b/lib/index.js.flow
@@ -1,0 +1,188 @@
+/* @flow
+ *
+**/
+const EventEmitter = require('events').EventEmitter;
+const Pool = require('pg-pool');
+import type {
+  PgPoolConfig,
+  PoolConnectCallback,
+  DoneCallback,
+  PoolClient
+} from 'pg-pool';
+
+// error
+type PG_ERROR = {
+  name: string,
+  length: number,
+  severity: string,
+  code: string,
+  detail: string|void,
+  hint: string|void,
+  position: string|void,
+  internalPosition: string|void,
+  internalQuery: string|void,
+  where: string|void,
+  schema: string|void,
+  table: string|void,
+  column: string|void,
+  dataType: string|void,
+  constraint: string|void,
+  file: string|void,
+  line: string|void,
+  routine: string|void
+};
+
+type ClientConfig = {
+  //database user's name
+  user: string,
+  //name of database to connect
+  database: string,
+  //database user's password
+  password: string,
+  //database port
+  port: number,
+  // database host. defaults to localhost
+  host: string,
+  // whether to try SSL/TLS to connect to server. default value: false
+  ssl: boolean,
+  // name displayed in the pg_stat_activity view and included in CSV log entries
+  // default value: process.env.PGAPPNAME
+  application_name: string,
+  // fallback value for the application_name configuration parameter
+  // default value: false
+  fallback_application_name: string,
+}
+
+type Row = {
+  [key: string]: any,
+};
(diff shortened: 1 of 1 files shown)


brianc:
First and foremost: I really appreciate the time you took to do this!!! Thank you! :pray: I don't know anything about flow yet, so I can't really review it unfortunately.  I'll leave it open for a week or so to give anyone else an opportitunty to review it.  Once all (or any!) feedback has been incorporated I'll merge it in!! :dancer:

joskuijpers:
Could you give a tiny code example with Flow? Then I will try it out.

Have you looked at the TypeScript definitions?

iamchenxin:
@brianc Thank you, hope someone can see this, and review my code.

iamchenxin:
@joskuijpers just fork both [node-pg-pool PR#26](https://github.com/brianc/node-pg-pool/pull/26)  and [`node-postgres PR#1120`](https://github.com/brianc/node-postgres/pull/1120) , and install them locally.
Then the type def should work, if there was it is not work, just check if in the `lib` directory was ignored in `.flowconfig `.

lib/index.js.flow
```diff
@@ -0,0 +1,188 @@
...
+    types: Types;
+    Client: Class<Client>;
+    Pool: Class<Pool>;
+    Connection: any;
+  };
+// The end(),connect(),cancel() in PG is abandoned ?
+}
+
+// module export, keep same structure with index.js
+declare var pg:PG ;
```
joskuijpers: Space is in wrong location, it should be `pg: PG;`

joskuijpers:
I would say we just merge and see what problems come up.
I am also in favor of moving the TS definiton to this repo and using the package.json typings attribute.

Does Flow support this attribute? If so, you might want to update it.

lib/index.js.flow
iamchenxin: marked, will amend them tomorrow, when the review finished, will also delete some comments only used for review purpose.
Sleeping, see you tomorrow. Thanks for review.

iamchenxin:
@joskuijpers Flow will auto pickup the `.js.flow` file. do not need any additional config in packages.
ex: assume there is a `src/a.js`, when add a `src/a.js.flow`, Flow will auto  pickup it, and use it as the `src/a.js`'s type definition.

iamchenxin:
@brianc will rebase them into one, when review finish.

joskuijpers:
Looks good to me. When a user of this finds a problem it can easily be fixed. It is only type definitions afterall,

iamchenxin:
@brianc Im still a bit confused with Object which extends from 'EventEmitter' (like client, query, pool).
Should them allow users emit and receive custom event on them? or those object should be restrict to their defined event like 'on','end', 'error'?

iamchenxin:
@brianc  squash into one commit.
A known issue is current Flow version, can not express overloading correctly between string literal and string, this will cause some problem in event overloading( event.on() ). But i think this will be fixed later.

iamchenxin:
@brianc Will move this to [flowtype/flow-typed](https://github.com/flowtype/flow-typed). Seems now Facebook encourage to place typedefs there.

PR #1133, opened 2016-09-20
when define a model without primary key, it will add automatically

JuneDeng2014:
``` js
var Sys = sequelize.define('name_sysid', {
    system_id:Sequelize.BIGINT,
        // primaryKey: true

    system_name: Sequelize.STRING(100),
    system_type: Sequelize.BIGINT,
}, {
        timestamps: false,
        freezeTableName: true
    });


Sys.findAll()
   .then(function (sys) {
       for (let i in sys) {
           console.log(`${i.type}: ${i.name}`);
       }
   }).catch(function (err) {
       // error
       console.log(err);
   });
```
# 

error information

> ```
> Executing (default): SELECT "id", "system_id", "system_name", "system_type" FROM "name_sysid" AS "name_sysid";
> { SequelizeDatabaseError: column "id" does not exist
> .....}
> ```

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -73,9 +73,12 @@ if(typeof process.env.NODE_PG_FORCE_NATIVE != 'undefined') {
   module.exports = new PG(Client);
 
   //lazy require native module...the native module may not have installed
-  module.exports.__defineGetter__("native", function() {
-    delete module.exports.native;
-    module.exports.native = new PG(require('./native'));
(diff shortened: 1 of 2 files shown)


joskuijpers:
@JuneDeng2014 Could you elaborate on how the problem relates to this pull request?

vitaly-t:
This PR is quite nonsensical. Should be just closed.

PR #1146, opened 2016-10-10
Add client_encoding parameter for connect

jonasrodrigeri:
So you can be informed the database encoding in connect


joskuijpers:
Hi! Thank you.
Could you add tests and make sure the tests succeed? (See TravisCI).

PR #1148, opened 2016-10-11
Fix newlines in string array

### LINKED ISSUE #1147: Newline in text array loses backslash
### Issue:

When inserting/updating a column of type `text[]`, newlines get converted to a character `n`.
### How to reproduce:

Run the following code:

```
var pg = require('pg');
var client = new pg.Client();
client.connect(function (err) {
  client.query('CREATE TABLE test (array_test text[])', function(err, result) {
    client.query('INSERT INTO test (array_test) VALUES ($1)', [['hello\nworld']], function (err, result) {
      client.query('SELECT * FROM test', function (err, result) {
        console.log(result.rows[0]);
        client.end();
      })
    })
  })
});
```
#### Behaviour:

The console.log is missing a backslash:

```
{ array_test: [ 'hellonworld' ] }
```
#### Expected behaviour

The console.log should output a string including the backslash:

```
{ array_test: [ 'hello\nworld' ] }
```
### Quick fix

Quick fix for this issue is to change `/lib/utils.js:28` to:

```
result = result + JSON.stringify(prepareValue(val[i])).replace('\\n', '\n');
```
### Other notes:

Not related to #895 (which seems to be fixed atm)

jehna:
Fixes #1147

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -25,7 +25,7 @@ function arrayString(val) {
     }
     else
     {
-      result = result + JSON.stringify(prepareValue(val[i]));
+      result = result + JSON.stringify(prepareValue(val[i])).replace('\\n', '\n');
     }
   }
   result = result + '}';
@@ -104,7 +104,7 @@ function dateToString(date) {
 }
 
 function dateToStringUTC(date) {
-  
+
   var ret = pad(date.getUTCFullYear(), 4) + '-' +
       pad(date.getUTCMonth() + 1, 2) + '-' +
       pad(date.getUTCDate(), 2) + 'T' +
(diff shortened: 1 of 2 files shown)


jehna:
Checks are failing for the error:

```
Message: role "sfalsdkf" does not exist
```

...but I'm getting that in the master also. Is this a broader problem?

swarthy:
Same here #1152

charmander:
Added a pull request for some other things that shouldn’t be escaped.

Also, failures are fixed by brianc/node-pg-pool#31.

brianc:
I merged https://github.com/brianc/node-postgres/pull/1177 to fix this! Will be pushing a patch version to npm shortly.

PR #1154, opened 2016-10-22
Use `buffer-alloc-unsafe` to create emptyBuffer

JLHwung:
Deprecation Warning when testing against node 7.0.0 rc1

> Using Buffer without `new` will soon stop working from node v7.0.0. Use `new Buffer()`, or preferably `Buffer.from()`, `Buffer.allocUnsafe()` or `Buffer.alloc()` instead.

I know we shall support node prior to 6.x, so I use `new Buffer()` to replace `Buffer()`.

### DIFF
--- a/test/test-buffers.js
+++ b/test/test-buffers.js
@@ -1,10 +1,11 @@
 require(__dirname+'/test-helper');
+var bufferFrom = require('buffer-from');
 //http://developer.postgresql.org/pgdocs/postgres/protocol-message-formats.html
 
 var buffers = {};
 buffers.readyForQuery = function() {
   return new BufferList()
-    .add(new Buffer('I'))
+    .add(bufferFrom('I'))
     .join(true,'Z');
 };
 
@@ -23,7 +24,7 @@ buffers.authenticationCleartextPassword = function() {
 buffers.authenticationMD5Password = function() {
   return new BufferList()
     .addInt32(5)
-    .add(new Buffer([1, 2, 3, 4]))
+    .add(bufferFrom([1, 2, 3, 4]))
     .join(true, 'R');
 };
 
@@ -71,7 +72,7 @@ buffers.dataRow = function(columns) {
     if(col == null) {
       buf.addInt32(-1);
     } else {
-      var strBuf = new Buffer(col, 'utf8');
+      var strBuf = bufferFrom(col, 'utf8');
       buf.addInt32(strBuf.length);
       buf.add(strBuf);
     }
@@ -94,7 +95,7 @@ var errorOrNotice = function(fields) {
     buf.addChar(field.type);
     buf.addCString(field.value);
   });
-  return buf.add(new Buffer([0]));//terminator
+  return buf.add(bufferFrom([0]));//terminator
 }
 
 buffers.parseComplete = function() {
--- a/test/test-buffers.js
+++ b/test/test-buffers.js
@@ -4,7 +4,7 @@ require(__dirname+'/test-helper');
 var buffers = {};
 buffers.readyForQuery = function() {
   return new BufferList()
-    .add(Buffer('I'))
+    .add(new Buffer('I'))
     .join(true,'Z');
 };
 
@@ -23,7 +23,7 @@ buffers.authenticationCleartextPassword = function() {
 buffers.authenticationMD5Password = function() {
   return new BufferList()
     .addInt32(5)
-    .add(Buffer([1,2,3,4]))
+    .add(new Buffer([1, 2, 3, 4]))
     .join(true, 'R');
 };
 
@@ -94,7 +94,7 @@ var errorOrNotice = function(fields) {
     buf.addChar(field.type);
     buf.addCString(field.value);
   });
-  return buf.add(Buffer([0]));//terminator
+  return buf.add(new Buffer([0]));//terminator
 }
 
 buffers.parseComplete = function() {
--- a/test/unit/client/md5-password-tests.js
+++ b/test/unit/client/md5-password-tests.js
@@ -3,7 +3,7 @@ require(__dirname + '/test-helper');
 test('md5 authentication', function() {
   var client = createClient();
   client.password = "!";
-  var salt = Buffer([1, 2, 3, 4]);
+  var salt = new Buffer([1, 2, 3, 4]);
   client.connection.emit('authenticationMD5Password', {salt: salt});
 
   test('responds', function() {
--- a/test/unit/client/md5-password-tests.js
+++ b/test/unit/client/md5-password-tests.js
@@ -1,9 +1,10 @@
 require(__dirname + '/test-helper');
+var bufferFrom = require('buffer-from');
 
 test('md5 authentication', function() {
   var client = createClient();
   client.password = "!";
-  var salt = new Buffer([1, 2, 3, 4]);
+  var salt = bufferFrom([1, 2, 3, 4]);
   client.connection.emit('authenticationMD5Password', {salt: salt});
 
   test('responds', function() {
--- a/test/unit/connection/inbound-parser-tests.js
+++ b/test/unit/connection/inbound-parser-tests.js
@@ -158,7 +158,7 @@ test('Connection', function() {
   testForMessage(plainPasswordBuffer, expectedPlainPasswordMessage);
   var msg = testForMessage(md5PasswordBuffer, expectedMD5PasswordMessage);
   test('md5 has right salt', function() {
-    assert.equalBuffers(msg.salt, Buffer([1,2,3,4]));
+    assert.equalBuffers(msg.salt, new Buffer([1, 2, 3, 4]));
   });
   testForMessage(paramStatusBuffer, expectedParameterStatusMessage);
   testForMessage(backendKeyDataBuffer, expectedBackendKeyDataMessage);
@@ -173,7 +173,7 @@ test('Connection', function() {
   });
 
   test("no data message", function() {
-    testForMessage(Buffer([0x6e, 0, 0, 0, 4]), {
+    testForMessage(new Buffer([0x6e, 0, 0, 0, 4]), {
       name: 'noData'
     });
   });
--- a/test/unit/connection/inbound-parser-tests.js
+++ b/test/unit/connection/inbound-parser-tests.js
@@ -1,6 +1,7 @@
 require(__dirname+'/test-helper');
 return false;
 var allocUnsafe = require('buffer-alloc-unsafe');
+var bufferFrom = require('buffer-from');
 var Connection = require(__dirname + '/../../../lib/connection');
 var buffers = require(__dirname + '/../../test-buffers');
 var PARSE = function(buffer) {
@@ -160,7 +161,7 @@ test('Connection', function() {
   testForMessage(plainPasswordBuffer, expectedPlainPasswordMessage);
   var msg = testForMessage(md5PasswordBuffer, expectedMD5PasswordMessage);
   test('md5 has right salt', function() {
-    assert.equalBuffers(msg.salt, new Buffer([1, 2, 3, 4]));
+    assert.equalBuffers(msg.salt, bufferFrom([1, 2, 3, 4]));
   });
   testForMessage(paramStatusBuffer, expectedParameterStatusMessage);
   testForMessage(backendKeyDataBuffer, expectedBackendKeyDataMessage);
@@ -175,7 +176,7 @@ test('Connection', function() {
   });
 
   test("no data message", function() {
-    testForMessage(new Buffer([0x6e, 0, 0, 0, 4]), {
+    testForMessage(bufferFrom([0x6e, 0, 0, 0, 4]), {
       name: 'noData'
     });
   });
--- a/test/unit/connection/inbound-parser-tests.js
+++ b/test/unit/connection/inbound-parser-tests.js
@@ -1,4 +1,6 @@
 require(__dirname+'/test-helper');
+return false;
+var allocUnsafe = require('buffer-alloc-unsafe');
 var Connection = require(__dirname + '/../../../lib/connection');
 var buffers = require(__dirname + '/../../test-buffers');
 var PARSE = function(buffer) {
@@ -376,8 +378,8 @@ test('split buffer, single message parsing', function() {
   });
 
   var testMessageRecievedAfterSpiltAt = function(split) {
-    var firstBuffer = new Buffer(fullBuffer.length-split);
-    var secondBuffer = new Buffer(fullBuffer.length-firstBuffer.length);
+    var firstBuffer = allocUnsafe(fullBuffer.length-split);
+    var secondBuffer = allocUnsafe(fullBuffer.length-firstBuffer.length);
     fullBuffer.copy(firstBuffer, 0, 0);
     fullBuffer.copy(secondBuffer, 0, firstBuffer.length);
     stream.emit('data', firstBuffer);
@@ -409,7 +411,7 @@ test('split buffer, single message parsing', function() {
 test('split buffer, multiple message parsing', function() {
   var dataRowBuffer = buffers.dataRow(['!']);
   var readyForQueryBuffer = buffers.readyForQuery();
-  var fullBuffer = new Buffer(dataRowBuffer.length + readyForQueryBuffer.length);
+  var fullBuffer = allocUnsafe(dataRowBuffer.length + readyForQueryBuffer.length);
   dataRowBuffer.copy(fullBuffer, 0, 0);
   readyForQueryBuffer.copy(fullBuffer, dataRowBuffer.length, 0);
 
@@ -442,8 +444,8 @@ test('split buffer, multiple message parsing', function() {
     verifyMessages();
   });
   var splitAndVerifyTwoMessages = function(split) {
-    var firstBuffer = new Buffer(fullBuffer.length-split);
-    var secondBuffer = new Buffer(fullBuffer.length-firstBuffer.length);
+    var firstBuffer = allocUnsafe(fullBuffer.length-split);
+    var secondBuffer = allocUnsafe(fullBuffer.length-firstBuffer.length);
     fullBuffer.copy(firstBuffer, 0, 0);
     fullBuffer.copy(secondBuffer, 0, firstBuffer.length);
     stream.emit('data', firstBuffer);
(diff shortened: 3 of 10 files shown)


joskuijpers:
Thanks! We need this, however:

"preferably Buffer.from(), Buffer.allocUnsafe() or Buffer.alloc()"

So please, use one of those :) Not `new Buffer()`.

JLHwung:
@joskuijpers I saw from https://github.com/brianc/node-postgres/pull/1017#issuecomment-227158408 that using `Buffer.from` will 
> breaks backwards compatibility with a lot of versions of node. 

So I am not sure what is the compatibility plan for the community.

Personally I am more than glad to use `Buffer.from`.

joskuijpers:
Oh stupid, of course! You are right, I wasn't thinking :)

@brianc needs to decide what the changes are in backwards compatibility. We might make a new major update not node-backwards compatible.

LinusU:
I would recommend using the [`buffer-alloc-unsafe`](https://npmjs.com/buffer-alloc-unsafe) package which abstracts the compatibility away.

brianc:
buffer-alloc-unsafe looks good - that way we can keep backwards compatibility!

As an aside - any idea why the tests are failing in every environment on travis?  AFAIK nothing has changed on this end in a long time.  I can't merge this until those tests are passing.  I don't have time to look at it for the next few days, but might have time next week.

JLHwung:
@brianc As of 2016-10-31, nodejs is to [discontinue support](https://github.com/nodejs/LTS#lts-schedule) on v0.10, shall we still need support v0.10?

@LinusU 👍 for `buffer-alloc-unsafe`.

charmander:
> As an aside - any idea why the tests are failing in every environment on travis? AFAIK nothing has changed on this end in a long time. I can't merge this until those tests are passing. I don't have time to look at it for the next few days, but might have time next week.

@brianc The tests are failing because pg doesn’t specify exact dependencies and pg-pool 1.5 is broken. The brokenness is due to an incorrect fix for brianc/node-pg-pool#24 in the form of brianc/node-pg-pool#28. The bug (along with a few others) is fixed properly but backwards-incompatibly in brianc/node-pg-pool#31.

jackycute:
I hope this will be fixed soon.
Thank you guys works!

test/test-buffers.js
```diff
@@ -1,10 +1,11 @@
 require(__dirname+'/test-helper');
+var allocUnsafe = require('buffer-alloc-unsafe');
 //http://developer.postgresql.org/pgdocs/postgres/protocol-message-formats.html
 
 var buffers = {};
 buffers.readyForQuery = function() {
   return new BufferList()
-    .add(Buffer('I'))
+    .add(allocUnsafe('I'))
```
LinusU: This should probably use [`Buffer.from()`](https://www.npmjs.com/package/buffer-from). `Buffer.allocUnsafe` is just for allocating a new buffer with a specific length

test/test-buffers.js
```diff
@@ -23,7 +24,7 @@ buffers.authenticationCleartextPassword = function() {
 buffers.authenticationMD5Password = function() {
   return new BufferList()
     .addInt32(5)
-    .add(Buffer([1,2,3,4]))
+    .add(allocUnsafe([1, 2, 3, 4]))
```
LinusU: same here, `Buffer.from`

test/test-buffers.js
```diff
@@ -71,7 +72,7 @@ buffers.dataRow = function(columns) {
     if(col == null) {
       buf.addInt32(-1);
     } else {
-      var strBuf = new Buffer(col, 'utf8');
+      var strBuf = allocUnsafe(col, 'utf8');
```
LinusU: same here, `Buffer.from`

test/test-buffers.js
```diff
@@ -94,7 +95,7 @@ var errorOrNotice = function(fields) {
     buf.addChar(field.type);
     buf.addCString(field.value);
   });
-  return buf.add(Buffer([0]));//terminator
+  return buf.add(allocUnsafe([0]));//terminator
```
LinusU: same here, `Buffer.from`

test/unit/client/md5-password-tests.js
```diff
@@ -1,9 +1,10 @@
 require(__dirname + '/test-helper');
+var allocUnsafe = require('buffer-alloc-unsafe');
 
 test('md5 authentication', function() {
   var client = createClient();
   client.password = "!";
-  var salt = Buffer([1, 2, 3, 4]);
+  var salt = allocUnsafe([1, 2, 3, 4]);
```
LinusU: same here, `Buffer.from`

test/unit/connection/inbound-parser-tests.js
```diff
@@ -159,7 +161,7 @@ test('Connection', function() {
   testForMessage(plainPasswordBuffer, expectedPlainPasswordMessage);
   var msg = testForMessage(md5PasswordBuffer, expectedMD5PasswordMessage);
   test('md5 has right salt', function() {
-    assert.equalBuffers(msg.salt, Buffer([1,2,3,4]));
+    assert.equalBuffers(msg.salt, allocUnsafe([1, 2, 3, 4]));
```
LinusU: same here, `Buffer.from`

test/unit/connection/inbound-parser-tests.js
```diff
@@ -174,7 +176,7 @@ test('Connection', function() {
   });
 
   test("no data message", function() {
-    testForMessage(Buffer([0x6e, 0, 0, 0, 4]), {
+    testForMessage(allocUnsafe([0x6e, 0, 0, 0, 4]), {
```
LinusU: same here, `Buffer.from`

test/unit/connection/outbound-sending-tests.js
```diff
@@ -104,12 +106,12 @@ test('bind messages', function() {
       .addInt16(0)
       .addInt16(4)
       .addInt32(1)
-      .add(Buffer("1"))
+      .add(allocUnsafe("1"))
```
LinusU: same here, `Buffer.from`

test/unit/connection/outbound-sending-tests.js
```diff
@@ -104,12 +106,12 @@ test('bind messages', function() {
...
       .addInt32(2)
-      .add(Buffer("hi"))
+      .add(allocUnsafe("hi"))
```
LinusU: same here, `Buffer.from`

test/unit/connection/outbound-sending-tests.js
```diff
@@ -120,7 +122,7 @@ test('with named statement, portal, and buffer value', function() {
   con.bind({
     portal: 'bang',
     statement: 'woo',
-    values: ['1', 'hi', null, new Buffer('zing', 'UTF-8')]
+    values: ['1', 'hi', null, allocUnsafe('zing', 'UTF-8')]
```
LinusU: same here, `Buffer.from`

test/unit/connection/outbound-sending-tests.js
```diff
@@ -132,12 +134,12 @@ test('with named statement, portal, and buffer value', function() {
     .addInt16(1)//binary
     .addInt16(4)
     .addInt32(1)
-    .add(Buffer("1"))
+    .add(allocUnsafe("1"))
```
LinusU: same here, `Buffer.from`

test/unit/connection/outbound-sending-tests.js
```diff
@@ -132,12 +134,12 @@ test('with named statement, portal, and buffer value', function() {
...
     .addInt32(2)
-    .add(Buffer("hi"))
+    .add(allocUnsafe("hi"))
```
LinusU: same here, `Buffer.from`

test/unit/connection/outbound-sending-tests.js
```diff
@@ -181,7 +183,7 @@ test('sends sync command', function() {
 
 test('sends end command', function() {
   con.end();
-  var expected = new Buffer([0x58, 0, 0, 0, 4]);
+  var expected = allocUnsafe([0x58, 0, 0, 0, 4]);
```
LinusU: same here, `Buffer.from`

test/unit/utils-tests.js
```diff
@@ -50,7 +50,7 @@ test('normalizing query configs', function() {
 })
 
 test('prepareValues: buffer prepared properly', function() {
-  var buf = new Buffer("quack");
+  var buf = allocUnsafe("quack");
```
LinusU: same here, `Buffer.from`

test/unit/utils-tests.js
```diff
@@ -142,7 +142,7 @@ test('prepareValue: objects with simple toPostgres prepared properly', function(
 });
 
 test('prepareValue: objects with complex toPostgres prepared properly', function() {
-  var buf = new Buffer("zomgcustom!");
+  var buf = allocUnsafe("zomgcustom!");
```
LinusU: same here, `Buffer.from`

test/unit/utils-tests.js
```diff
@@ -165,7 +165,7 @@ test('prepareValue: objects with toPostgres receive prepareValue', function() {
 });
 
 test('prepareValue: objects with circular toPostgres rejected', function() {
-  var buf = new Buffer("zomgcustom!");
+  var buf = allocUnsafe("zomgcustom!");
```
LinusU: same here, `Buffer.from`

LinusU:
This doesn't work, only `new Buffer(number)` should be replaced with `buffer-alloc-unsafe`, `new Buffer(string, string)` and `new Buffer(array)` should be replaced with [`buffer-from`](https://www.npmjs.com/package/buffer-from)

JLHwung:
@LinusU Reworked on your review. Thank you so much for stating the difference between `Buffer.from` and `Buffer.allocUnsafe` 🙏

charmander:
These new packages don’t specify exact dependencies, adding a new person (the maintainer of is-array-buffer-x) to trust. I’d stick with `new Buffer` for now.

LinusU:
> These new packages don’t specify exact dependencies

Why should they? The best practice is to specify `^` dependencies. If you want to lock down your dependencies I would recommend using `npm shrinkwrap`, it will give you much higher guarantees than having exact dependencies.

The node philosophy is too have many small modules that does one thing, and does it well. I think that we should embrace that.  

> adding a new person (the maintainer of is-array-buffer-x) to trust

npms.io rates that package as 97% in quality and 100% in maintenance. That maintainer (@Xotic750) has over 40 modules on npm, and also, we only need to trust him not to publish breaking changes in a patch or minor bump. You can no longer remove packages from npm which other people depend on.

> I’d stick with `new Buffer` for now.

This is printing deprecation warnings on the current release line of Node.js which is unacceptable to a lot of people...

charmander:
> Why should they?

So that all new installations of pg don’t break when a patch version introduces bugs, or even malicious code. This is what I meant by “trust”.

> This is printing deprecation warnings on the current release line of Node.js which is unacceptable to a lot of people...

It isn’t. `Buffer()` *without* `new` is what causes the deprecation warning.

LinusU:
> It isn’t. `Buffer()` _without_ `new` is what causes the deprecation warning.

Correct, my bad :)

AlexTes:
I don't think merging an already deprecated method of creating buffers is the right way forward @charmander. For a more elaborate explanation of why see [here](https://github.com/AlexTes/buffer-deprecation).

I don't think we should look to mitigate the warning, we should fix the underlying issue. After all using `new Buffer()` will just result in a warning in time with the node team currently clearly stating to use neither `Buffer()` or `new Buffer()`.

charmander:
> After all using `new Buffer()` will just result in a warning in time with the node team currently clearly stating to use neither `Buffer()` or `new Buffer()`.

With any luck, it will be possible to drop Node 0.10–0.12 support by then. In the meantime, these “ponyfills” add no benefit, and the suggested ones [break the rules][1] (in addition to the problems previously noted):

  > **Ponyfills should never use the native API, even if available,** as it might have slightly different behavior between environments, which can cause bugs.


  [1]: https://github.com/sindresorhus/ponyfill#how-are-ponyfills-better-than-polyfills

AlexTes:
I'm confused. I'd say they add one clear benefit @charmander. They use the non-deprecated implementation whenever possible. Using deprecated methods is almost by definition not the right way to do something wouldn't you say?

I'll have to have a more thorough think / read on the ponyfill situation. From my understanding both safe-buffer and @LinusU's buffer modules are ponyfills.

AlexTes:
@charmander after reading up more I think the point Sindre is trying to make is that ponyfills could potentially work differently in different environments which would be a very bad thing and I agree. 

In this case both ponyfills I linked in my elaboration use the native api in both cases, which if I understand correctly the node team makes sure result in identical functionality. In this case using the available native api in both cases is using the same implementation and thus doesn't break Sindre's rule.

LinusU:
> and the suggested ones break the rules

On mobile now, but I'm quite sure that that is necessary in this case since if it only relied on the `new Buffer` api the ponyfill in itself would emit the warning

JLHwung:
As node.js has reverted buffer runtime deprecation in https://github.com/nodejs/node/pull/9529, I think this pull request can be closed.

AlexTes:
@JLHwung how so? They still want to move forward with the deprecation. The current implementation this library uses is still marked deprecated and eventually will face harsher deprecation it seems. The core team seems unambiguous in their intention for users to use the new API, just unsure on how to help the ecosystem get there. This PR gets us there, with backwards compatibility and next to zero added dependency. Let's help make this easier on the core team ^_^.

charmander:
> This PR gets us there, with backwards compatibility and next to zero added dependency.

```shellsession
$ npm info buffer-from dependencies
{ 'is-array-buffer-x': '^1.0.13' }

$ npm info is-array-buffer-x dependencies
{ 'is-object-like-x': '^1.0.11',
  'to-string-tag-x': '^1.0.11',
  'has-to-string-tag-x': '^1.0.10' }

$ npm info is-object-like-x dependencies
{ 'is-function-x': '^1.0.6', 'is-primitive': '^2.0.0' }

$ npm info is-function-x dependencies
{ 'has-to-string-tag-x': '^1.0.10',
  'to-string-tag-x': '^1.0.11',
  'is-primitive': '^2.0.0' }

$ npm info has-to-string-tag-x dependencies
{ 'has-symbol-support-x': '^1.0.11' }

$ npm info to-string-tag-x dependencies
{ 'lodash.isnull': '^3.0.0', 'validate.io-undefined': '^1.0.3' }

$ npm info buffer-alloc dependencies
{ 'buffer-alloc-unsafe': '^0.1.0', 'buffer-fill': '^0.1.0' }
```

If you’re dead set on using `Buffer.alloc` and `Buffer.from` before Node <4 support can be dropped, might I suggest:

```javascript
var bufferAlloc = Buffer.alloc || Buffer;
var bufferFrom = Buffer.from || Buffer;
```

AlexTes:
@charmander I don't see why you would look at the _full_ dependency tree unless you're going to argue the library we're suggesting to take on is far too abstract and over complicates. I don't want to argue the details. You are right, it does add a number of dependencies.

I'm trying not to be dead set on anything. I do think merging this is better for you, me and the ecosystem :].

Also, your code suggestion looks fine to me. Beyond a small change I'd make it's just style differences I think. Either abstract away the complexity of using a new API with backward compatibility or kill the dependency and own it. Up to all the people involved I'd say. I would like to add that the friendly cat you linked earlier would probably be in favor of the ponyfills, based on his writing about [small dependencies](https://github.com/sindresorhus/ama/issues/10#issuecomment-117766328).

Thanks for helping us move with the node team Charmander!

charmander:
> I don't see why you would look at the full dependency tree unless you're going to argue the library we're suggesting to take on is far too abstract and over complicates. I don't want to argue the details. You are right, it does add a number of dependencies.

I think the library is well-written! My concern, though, is all these dependencies with caret-versions. We’ve just seen pg-pool 1.5 unintentionally break the build – maintainers can’t be perfect all the time. This is a set of 11 new packages, any one of which could have a new patch version published at some point in the future that affects behaviour in a surprising way.

So, nothing against small packages\*, but in the interest of reducing points of failure I’m for eating the maintenance costs in pg itself in the end, given only those two options. Requiring a supported version of Node with `Buffer.from` built in would be even nicer, though…

<sup>\* Except for `lodash.isnull` and `validate.io-undefined`, which are `=== null` and `=== undefined`, respectively. to-string-tag-x doesn’t use them in a way that requires them to be functions.</sup>

AlexTes:
@charmander cool! I can only agree. I would prefer to drop node < 4.
If owning the dependency is judged the better way forward I'm fine with that too and would even create PR myself!

I guess now we wait for someone with merge powers to show up and start moving!

LinusU:
I don't think that the problem with caret should be solved by dropping them. There are solutions to this problem baked right into npm; [npm shrinkwrap](https://docs.npmjs.com/cli/shrinkwrap).

Another popular solution is [Yarn](https://yarnpkg.com)

AlexTes:
All the way there with you @LinusU! I think perhaps it'd be sensible to wait for a maintainer to voice how they even feel in the first place about the PR as it stands now.

brianc:
Hey y'all - I really appreciate you going through this.  Sorry I've been out for a while. I've been maintaining node-postgres for >6 years most of the time solo, and always for free, and sometimes life gets really hectic, and I have to step away for some time.  I'm finding a bit more time now, things are calming down, so that's good. Just wanted to give a bit of background as to why this has sit for so long.  I'm going to do a better job of finding more maintainers here so I'm not a single point of failure anymore.

As far as this PR is concerned, @charmander did the lovely work of getting CI passing again 🙏 .  I would still like to support `node@0.10.x` _if possible_, but I know it's really old & so I'm fine with doing a major version bump & breaking change to drop support for node < v0.10.x if y'all think that's right!

What I'd like to see: can we get this rebased back with the current master so we pick up the fixes for travis?  Once the CI is running I can get a good look and see what versions of node are broken.  If `node@0.10.x` is failing due to this change we can bump a major version & drop support.  👏 

as an aside: it's annoying that node changed the api for `Buffer`.  👎 

#### tl;dr - rebase this with master so we can get it running in Travis. After that I'll merge this.

JLHwung:
@brianc Rebased on master. Appreciated for your attitude.

vitaly-t:
This PR is no longer needed. The depreciation warning has been removed from Node.js, starting from version 7.2.1

AlexTes:
Awesome to see your face @brianc! Glad you found the time. Totally understandable it took a bit!

@vitaly-t see [this comment](https://github.com/brianc/node-postgres/pull/1154#issuecomment-265326912) a little bit up. I replied to it already.

JLHwung:
@brianc Rebased on master.

charmander:
pg@7 is going to be able to use `Buffer.from` directly! See #1298.

Issue #1141, opened 2016-10-02
connection strings not working in 6.x


vitaly-t:
Moving from v5.1 to v6.x, the same connection strings no longer work.

This used to work up until v5.1 at least:

``` js
var cn = "postgres://postgres@localhost:5433/newone";
pg.connect(cn,...)
```

Now with v6.1, calling:

``` js
var cn = "postgres://postgres@localhost:5433/newone";
var pool = new pg.Pool(cn);
pool.connect(function (err, client, done) {
```

throws the following error:

```
{ error: database "Vitaly" does not exist
    at Connection.parseE (D:\NodeJS\tests\node_modules\pg\lib\connection.js:543:11)                                                                                 at Connection.parseMessage (D:\NodeJS\tests\node_modules\pg\lib\connection.js:370:17)
    at Socket.<anonymous> (D:\NodeJS\tests\node_modules\pg\lib\connection.js:109:22)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)                                                at readableAddChunk (_stream_readable.js:176:18)
    at Socket.Readable.push (_stream_readable.js:134:10)                            at TCP.onread (net.js:543:20)
  name: 'error',                                                                  length: 107,
  severity: 'FATAL',
  code: '3D000',
  detail: undefined,
  hint: undefined,
  position: undefined,
  internalPosition: undefined,
  internalQuery: undefined,
  where: undefined,
  schema: undefined,
  table: undefined,
  column: undefined,
  dataType: undefined,
  constraint: undefined,
  file: 'src\\backend\\utils\\init\\postinit.c',
  line: '775',
  routine: 'InitPostgres' }
```

I had to explicitly switch to the config object to make it work:

``` js
var cn = {
    database: 'newone',
    port: 5433,
    host: 'localhost',
    user: 'postgres'
};
```

Globik:
Your database should be Vitaly? Then in var cn ={database : 'Vitaly'};

vitaly-t:
@Globik what do you mean? The bug is with the connection strings, got nothing to do with connection config objects.

Globik:
See here: error: database "Vitaly" does not exist

Globik:
You should to have a database named "Vitaly". Or create one if not exists. And remove from string "newone". It is not a mongodb where if not exists db create one on the fly.

Globik:
User: postgres, but where is your password? Also should be there in options.

Globik:
[...]

PR #1157, opened 2016-10-27
Parse connection string into config object before creting pool.

KyleLilly:
This should resolve https://github.com/brianc/node-postgres/issues/1141 at least it seems to work locally for me doing something like:

```
var pg = require('pg');
var cn = "postgres://postgres@localhost:5433/testing";
var pool = new pg.Pool(cn);
pool.connect(function (err, client, done) { ... });
```

I didn't add any tests because I'm having trouble getting the existing tests to pass (and it looks like all the recent PR's have failed the Travis build). If you've got advice for getting things to pass locally I can add something to the existing `connection-string` unit tests. Unless you'd prefer them somewhere else.

### DIFF
--- a/lib/pool-factory.js
+++ b/lib/pool-factory.js
@@ -1,10 +1,13 @@
 var Client = require('./client');
 var util = require('util');
 var Pool = require('pg-pool');
+var parse = require('pg-connection-string').parse;
 
 module.exports = function(Client) {
 
   var BoundPool = function(options) {
+    // Handle connection string based pool creation
+    options = typeof options == 'string' ? parse(options) : (options || {});
     var config = { Client: Client };
     for (var key in options) {
       config[key] = options[key];


brianc:
Thanks for the pull request!  Is there a reason we should do this here instead of having it in your own code?  Because the call to `pg.connect` configures both the pool and the client it seems a bit odd to me as an API to use a connection string.

What issue are you having with the tests running locally? If you paste me your output I might be able to help out.  I realize they're not the most straight forward tests in the world - they were written 6+ years ago before node actually had any really viable test frameworks, so I just rolled this stuff by hand.

KyleLilly:
So the main reason I was looking into this was related to pg-promise. Since that library isn't updating to the latest version it's not really impacting anything on our end but I was kind of hoping that they would update to the 6.x release and this seemed to be a blocker for them.

As far as the test failures I was just getting the same failures that travis is reporting. I can try and run them again see if anything changes but it seemed like all the recent PRs were failing.

KyleLilly:
Didn't realize that there was a PR to fix the travis build. Locally I'm getting
```
***Testing Pure Javascript***
api-tests.js........Message: role "sfalsdkf" does not exist
error: role "sfalsdkf" does not exist
<stack trace>
make: *** [test-integration] Error 1
```
Looking at the test I'd expect an error but it seems to be treating it as a failure rather than expected.

charmander:
@KyleLilly That one should be fixed if you upgrade pg-pool from 1.5 to 1.6. (pg depends on 1.\*, so it’s not very obvious.)

KyleLilly:
The more I think about it the less this PR makes sense. Going from 5.1 to 6.x means changing major versions and the API changed so I would expect the onus of to be on the consumers. I'm fine with just closing this.

charmander:
See also #1068.

vvo:
```
const db = new pg.Pool({
  connectionString: process.env.DATABASE_URL,
});
```

Works

PR #1178, opened 2016-12-06
fixes utf password md5 bug

VaoTsun:
fixes the bug described in
http://stackoverflow.com/questions/33935749/node-js-postgres-utf-connect-string
In short: 
utf password md5 generated is wrong because of binary encoding

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -344,7 +344,7 @@ Client.prototype.end = function(cb) {
 };
 
 Client.md5 = function(string) {
-  return crypto.createHash('md5').update(string).digest('hex');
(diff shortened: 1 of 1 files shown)


charmander:
Adding a test – just a quick `Client.md5('😊') === '5deda34cd95f304948d2bc1b4a62c11e'` – might help this go in faster.

VaoTsun:
[example where this small change fixes the bug](http://stackoverflow.com/a/41005963/5315974)

brianc:
@VaoTsun - thanks!  I'll write a test for this & get it merged in.

PR #1179, opened 2016-12-08
Fix CI

charmander:
I’m not sure exactly what changed on Travis CI in the past few months, but this seems to make everything pass while still running all the tests.


brianc:
## beautiful

THANK YOU SO MUCH! Was pulling my hair out on that.

PR #1182, opened 2016-12-13
Run inbound parser tests

charmander:
They were disabled by 4cdd7a116bab03c6096ab1af8f0f522b04993768 without comment; it seems that this might have been unintentional?

In any case, they should probably be enabled, updated, or removed.


brianc:
Oh nice! I probably did that mid-development to get some other test running & then accidentally forgot & left it that way! 😛

PR #1202, opened 2017-01-23
Remove single line comments from query text

coloraggio:
For a complex app with lots of long SQL queries I'm using .sql files to write queries and reading them into memory (an object with .sql filenames as keys and content as string values) before app starts.

In SQL queries small, single line comments sometimes are very useful and this 3 lines of code removes them before feeding to addCString(query.text)

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -222,9 +222,9 @@ Connection.prototype.parse = function(query, more) {
 
   //remove single line comments from query text
   if (query.text) {
-    query.text = query.text.replace(/--.+\n/g, ' ');
+    query.text = query.text.replace(/--.+\n?/g, ' ');
   }
-  
+
   //normalize null type array
   query.types = query.types || [];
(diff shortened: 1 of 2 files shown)


charmander:
Comments should work fine already; why do you need to strip them?

Anyway, regular expressions aren’t the right way to accomplish that; `SELECT '--'` would break, for one.

brianc:
I appreciate the thought here, but I think removing comments from your sql text is something better handled in a 3rd party library.  If you publish something to npm to do this I'll happily link to it in the [extras](https://github.com/brianc/node-postgres/wiki/Extras) section of the wiki!

vitaly-t:
@coloraggio There is [pg-minify] for that, and it is listed in the [Extras].

[Extras]:https://github.com/brianc/node-postgres/wiki/Extras
[pg-minify]:https://github.com/vitaly-t/pg-minify

Issue #1048, opened 2016-06-11
connection.js line 635 infinite loop


l0oky:
I was using `sequelize` to connect to my `postgres` database. However, not paying attention, I didn't configure the right port and username/password. That made my node.js app loop infinitely on [this line](https://github.com/brianc/node-postgres/blob/master/lib/connection.js#L635). I figured this line out by using `node debug` .

```
break in node_modules/pg/lib/connection.js:635
 633 Connection.prototype.parseCString = function(buffer) {
 634   var start = this.offset;
>635   while(buffer[this.offset++] !== 0) { }
 636   return buffer.toString(this.encoding, start, this.offset - 1);
 637 };
debug> o
break in node_modules/pg/lib/connection.js:635
 633 Connection.prototype.parseCString = function(buffer) {
 634   var start = this.offset;
>635   while(buffer[this.offset++] !== 0) { }
 636   return buffer.toString(this.encoding, start, this.offset - 1);
 637 };
debug> o
```

I don't experience any other issues If I use correct data, so this may be a potential bug, just pointing this discovery out.

joskuijpers:
Can you give a piece of code that reproduces this error?

charmander:
This library isn’t built to handle malformed messages at all right now. This case can be fixed by using `indexOf` instead of a manual loop, but the built-in message parsers in general don’t attempt to detect errors. @brianc, would you be interested in changes to that end?

dotchev:
> the built-in message parsers in general don’t attempt to detect errors

This sounds like serious troubles.
As a start can you add it to the readme so people are aware.
Any plans to add errors checks or would you accept PRs in that direction?

brianc:
I wouldn't consider it _serious troubles_. It only causes an issue if you connect the driver to something that isn't postgres at all and that backend starts sending unexpected binary data back to the driver.  Kinda like doing 'curl' on to the port postgres is running on: ain't gonna work. :hushed: 

I will absolutely accept a PR if it fixes bugs or addes enhancements, has tests, and seems reasonable!

MrCheater:
+1

PR #1208, opened 2017-02-08
Avoid infinite loop on malformed message

charmander:
If handling of such messages is deemed unimportant, `indexOf` is still faster (~40%) and cleaner than a manual loop.

Addresses #1048 to an extent.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -647,8 +647,9 @@ Connection.prototype.readBytes = function(buffer, length) {
 
 Connection.prototype.parseCString = function(buffer) {
   var start = this.offset;
-  while(buffer[this.offset++] !== 0) { }
-  return buffer.toString(this.encoding, start, this.offset - 1);
+  var end = buffer.indexOf(0, start);
+  this.offset = end + 1;
+  return buffer.toString(this.encoding, start, end);
 };
 //end parsing methods
 module.exports = Connection;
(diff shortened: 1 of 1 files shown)


charmander:
Oh, Node ≤0.12 didn’t have `indexOf`. … next major?

brianc:
Sorry for my delay in responding here - could you rebase this on master now that I merged your PR to remove the heroku bits?

charmander:
@brianc Rebased!

lib/connection.js
```diff
@@ -647,8 +662,9 @@ Connection.prototype.readBytes = function(buffer, length) {
 
 Connection.prototype.parseCString = function(buffer) {
   var start = this.offset;
-  while(buffer[this.offset++] !== 0) { }
-  return buffer.toString(this.encoding, start, this.offset - 1);
+  var end = indexOf(buffer, 0, start);
+  this.offset = end + 1;
```
LinusU: `this.offset` will be set to `0` if there wasn't a `'\0'` found, is that correct or should it maybe throw? :)

brianc:
Thanks! Merging & pushing patch version now.

lib/connection.js
charmander: I didn’t want to throw since none of the other parsers do yet… it’s a bit sticky, to be sure.

PR #1210, opened 2017-02-08
Update readme

pataiadam:
There is callback which releases the client back to the pool. It works fine, however when `client.query` throws an error (_eg.: "Connection terminated"_), we need to destroy the client instead of release.

Fortunately the `pg-pool` [handles](https://github.com/brianc/node-pg-pool/blob/master/index.js#L99) this issue, but we need to pass the error object to the callback function.


brianc:
Nice! Thank you!

PR #1218, opened 2017-02-15
Handle throws in type parsers

### LINKED ISSUE #1204: Errors thrown in type parser not passed through callback
Errors thrown in type parsers are thrown in a new tick but not passed as an error to the callback.

### Example

```js
const pg = require('pg')

pg.types.setTypeParser(20, value => {
  throw new Error()
})

var client = new pg.Client()

client.connect(err => {
  if (err) throw err

  try {
    client.query(`
      DROP TABLE IF EXISTS "bigint";

      CREATE TABLE "bigint" (
        "value" BIGINT
      );

      INSERT INTO "bigint" ("value") VALUES (1) RETURNING "value"
    `, (err, result) => {
      if (err) {
        console.log('Error passed through callback') // Never logged
        throw err
      }

      client.end(err => {
        if (err) throw err
      })
    })
  } catch (err) {
    console.log('Error thrown in same microtask') // Never logged
    throw err
  }
})
```

### Expected

The error to be passed to the callback and `Error passed through callback` being logged in the example.

### Actual

The error is thrown in a new tick and not handled which causes the callback to never be called. This causes a error that can't be handled easily without use of `process.on('uncaughtException', ...)`

LinusU:
Fixes #1204

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -80,7 +80,19 @@ Query.prototype.handleRowDescription = function(msg) {
 };
 
 Query.prototype.handleDataRow = function(msg) {
-  var row = this._result.parseRow(msg.fields);
+  var row;
+
+  if (this._canceledDueToError) {
+    return;
+  }
+
+  try {
+    row = this._result.parseRow(msg.fields);
+  } catch (err) {
+    this._canceledDueToError = err;
+    return;
+  }
+
   this.emit('row', row, this._result);
   if (this._accumulateRows) {
     this._result.addRow(row);
--- a/test/unit/client/throw-in-type-parser-tests.js
+++ b/test/unit/client/throw-in-type-parser-tests.js
@@ -1,0 +1,111 @@
+var helper = require(__dirname + "/test-helper");
+var types = require('pg-types')
+
+test('handles throws in type parsers', function() {
+
+  types.setTypeParser('special oid that will throw', function () {
+    throw Object.assign(new Error('Test'), { throwInTypeParser: true });
+  });
+
+  test('emits error', function() {
+    var handled;
+    var client = helper.client();
+    var con = client.connection;
+    var query = client.query('whatever');
+
+    handled = con.emit('readyForQuery');
+    assert.ok(handled, "should have handled ready for query");
+
+    con.emit('rowDescription',{
+      fields: [{
+        name: 'boom',
+        dataTypeID: 'special oid that will throw'
+      }]
+    });
+    assert.ok(handled, "should have handled row description");
+
+    assert.emits(query, 'error', function(err) {
+      assert.equal(err.throwInTypeParser, true);
+    });
+
+    handled = con.emit('dataRow', { fields: ["hi"] });
+    assert.ok(handled, "should have handled first data row message");
+
+    handled = con.emit('commandComplete', { text: 'INSERT 31 1' });
+    assert.ok(handled, "should have handled command complete");
+
+    handled = con.emit('readyForQuery');
+    assert.ok(handled, "should have handled ready for query");
+  });
+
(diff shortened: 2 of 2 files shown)


lib/query.js
```diff
@@ -80,7 +80,18 @@ Query.prototype.handleRowDescription = function(msg) {
...
 
 Query.prototype.handleDataRow = function(msg) {
-  var row = this._result.parseRow(msg.fields);
+  var row;
+
+  try {
+    row = this._result.parseRow(msg.fields);
+  } catch (err) {
+    if(this.callback) {
+      return this.callback(err);
```
charmander: This will call the callback at least twice if there are errors: once for each errored row, and then once to incorrectly indicate a successful completion, with errored rows missing.

LinusU: Good catch, looking into it now 👍

LinusU:
@charmander should be fixed now :)

LinusU:
@charmander Studied the code more carefully and came up with what I think is a more proper solution. This should now handle the errors correctly and the tests should test it properly (e.g. was missing `assert.calls` before)

test/unit/client/throw-in-type-parser-tests.js
```diff
@@ -0,0 +1,111 @@
+var helper = require(__dirname + "/test-helper");
+var types = require('pg-types')
+
+test('handles throws in type parsers', function() {
+
+  types.setTypeParser('special oid that will throw', function () {
+    throw Object.assign(new Error('Test'), { throwInTypeParser: true });
```
charmander: `Object.assign` isn’t compatible with Node ≤ 0.12.

charmander: It might be best to create the error outside the type parser and check that it’s equal to the emitted error rather than adding `throwInTypeParser`.

LinusU: Very smart, I'll fix that today 👌

LinusU:
> Error: node-v0.12.18-headers.tar.gz local checksum 05e09ef1bb70bab3bff45f677f8a199c3535c4efce815d9819beb683fd4471d1 not match remote 992f8d2bde6959abd8e5d7e9973b5e87c77af28ee9ca628dbf9c004ef5f7f42b

I don't think this failure is related to my commit :) restarting the build...

LinusU:
🤔 same error again, seems like something is broken on Travis or the Node.js releases server...

LinusU:
...and the build passes! 🎉 🎉

LinusU:
ping @brianc, any chance of getting this merged? :)

LinusU:
ping @brianc, I would really appreciate it if we could merge this one 🙌

brianc:
I appreciate the effort you put into this, but this doesn't look like the right way to handle this to me.  Swallowing the error and setting a `canceledDueToError` property on the query seems non-idiomatic node. Why not have the query emit an `error` event?  

Also, I'm not sure forcing a `try/catch` block on everyone is the right way to go - there's a perf impact every time you do a `try/catch` and that's a pretty hot code-path.  Do you have any benchmarks on how this might impact the speed of reading results?

I think possibly a better way to go is to pass the query object as another parameter to the type parser function, allowing the type parser to force the query object to emit an error if the type parser receives an unparsable type.  Then each parser could decide if it wants to do a `try/catch` internally.  I'd have to think about it a bit more...but this doesn't seem to be the proper approach.  Now every time you do a query you have to check if the query was canceled due to error? 🙅

brianc:
Publishing new patch version to npm now - sorry for the delay.

LinusU:
Thank you ❤️

PR #1235, opened 2017-03-04
Improve Readme.md for not so advanced users

javiertury:
1. Add brief description about the 3 possible ways of executing queries: passing the query to a pool, borrowing a client from a pool or obtaining an exclusive client. Give examples for the 3 of them.

2. Use the examples to teach how to reuse a pool in all of your project. This should be helpful for not so advanced users and prevents mistakes.

3. Open a troubleshooting section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ module.exports.query = function (text, values, callback) {
 // the pool also supports checking out a client for
 // multiple operations, such as a transaction
 module.exports.connect = function (callback) {
-  pool.connect(callback);
+  return pool.connect(callback);
 };
 ```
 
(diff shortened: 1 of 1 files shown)


abenhamdine:
Very good idea !

brianc:
Nice! Thanks @javiertury !!  One small request: could you make the troubleshooting section link to the [wiki](https://github.com/brianc/node-postgres/wiki/FAQ) instead of putting it in the readme? I find the wiki easier for other folks to edit w/o having to go through the PR process and there are already a bunch of questions on there w/ answers.

javiertury:
i agree that it's better to keep the readme brief. The troubleshooting section now just points to the FAQ on the wiki.

Since I found that I can extend some examples in the wiki, I've added how to use promises and async/await there. I've linked these new examples from the "pooling example" section on the readme.

brianc:
awesome! thank you @javiertury - sorry for the big huge delay on this...I have been traveling & working too hard! 😬

PR #1250, opened 2017-03-23
Fix README.md

iRockyZhou:
According to the markdown grammar, repair the title of the article shows the problem of error.


brianc:
Thanks @iRockyZhou !!

PR #1254, opened 2017-04-02
Support for logical streaming replication

kibae:
Hello,

Please check this code.
1. database connection for replication mode
1. WalStream class includes ```data``` event for logical replication stream
1. includes sample code ```test/test-walstream.js```

```js
'use strict';

const WalStream = require(__dirname + '/../lib').WalStream;

var lastLsn = null;

var walStream = new WalStream({});
function proc() {
  walStream.getChanges('test_slot', lastLsn, {
    includeXids: false, //default: false
    includeTimestamp: false, //default: false
    skipEmptyXacts: true, //default: true
  }, function(err) {
    if (err) {
      console.log('Logical replication initialize error', err);
      setTimeout(proc, 1000);
    }
  });
}

walStream.on('data', function(msg) {
  lastLsn = msg.lsn || lastLsn;

  //DO SOMETHING. eg) replicate to other dbms(pgsql, mysql, ...)
  console.log('log recv', msg);
}).on('error', function(err) {
  console.log('Error #2', err);
  setTimeout(proc, 1000);
});

proc();

//If want to stop replication
//walStream.stop();
```

### DIFF
--- a/lib/connection-parameters.js
+++ b/lib/connection-parameters.js
@@ -57,6 +57,7 @@ var ConnectionParameters = function(config) {
   this.binary = val('binary', config);
   this.ssl = typeof config.ssl === 'undefined' ? useSsl() : config.ssl;
   this.client_encoding = val("client_encoding", config);
+  this.replication = val("replication", config);
   //a domain socket begins with '/'
   this.isDomainSocket = (!(this.host||'').indexOf('/'));
 
@@ -81,6 +82,9 @@ ConnectionParameters.prototype.getLibpqConnectionString = function(cb) {
 
   if(this.database) {
     params.push("dbname='" + this.database + "'");
+  }
+  if(this.replication) {
+    params.push("replication='" + (this.database === true ? "true" : this.replication) + "'");
   }
   if(this.host) {
     params.push("host=" + this.host);
--- a/lib/client.js
+++ b/lib/client.js
@@ -26,6 +26,7 @@ var Client = function(config) {
   this.port = this.connectionParameters.port;
   this.host = this.connectionParameters.host;
   this.password = this.connectionParameters.password;
+  this.replication = this.connectionParameters.replication;
 
   var c = config || {};
 
@@ -221,6 +222,9 @@ Client.prototype.getStartupConf = function() {
   var appName = params.application_name || params.fallback_application_name;
   if (appName) {
     data.application_name = appName;
+  }
+  if (params.replication) {
+    data.replication = params.replication === true ? 'true' : params.replication;
   }
 
   return data;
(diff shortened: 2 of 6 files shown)


lib/connection-parameters.js
```diff
@@ -82,6 +83,9 @@ ConnectionParameters.prototype.getLibpqConnectionString = function(cb) {
   if(this.database) {
     params.push("dbname='" + this.database + "'");
   }
+  if(this.replication) {
+    params.push("replication='" + (this.database === true ? "true" : this.replication) + "'");
```
charmander: Is `this.database === true` intended to read `this.replication`?

kibae: I implemented pg_recvlogical's access method the same way.
The replication value can have a value of true or database.

https://github.com/pipelinedb/pipelinedb/blob/bdcd1fcd37584a3ad5427bb1a1b27a1af7521f0c/src/bin/pg_basebackup/streamutil.c#L105

kibae: Oh, sorry. That's pipelinedb link..
please check this out :)

https://github.com/postgres/postgres/blob/839da3c7d9b76b178910ce2999f7cbbdf6e91f70/src/bin/pg_basebackup/streamutil.c#L110

kibae:
Let me know if you have a problem with my code.

brianc:
Hey @kibae - this is pretty awesome! Sorry for the delay.  I'll merge this now!

What I'm thinking of doing is pulling out the Walstream code into it's own module & only adding the things to node-postgres itself to support the replication.  That way the `pg-walstream` can live on its own.  Would you be cool with me doing that or do you want to do that - I'm fine either way? I understand we'll still need the changes inside node-postgres core to support the replication setup & backend-events, but I try to keep things as slim as possible within node-postgres itself - having a separate installable module for the wal-stream will allow it to mature faster on its own.

Does that make sense?

lib/connection-parameters.js
charmander: So it’s intended to have:

```js
{
    database: true,
    replication: 'truthy value',
}
```

resulting in:

```js
[
    "dbname='true'",
    "replication='true'",
]
```

?

brianc:
I'll wait for @charmander's approval - looks like there might be some :interrobang: about the code still.

lib/connection-parameters.js
kibae: "Replication" can have "database", "true" and "false" values.

https://github.com/postgres/postgres/blob/19dc233c32f2900e57b8da4f41c0f662ab42e080/src/backend/postmaster/postmaster.c#L2088

For a clean commit log, I will just clean up the connection related code and request a new pull.

kibae:
@brianc Oh, I see.
For a clean commit log, I will just clean up the connection related code and request a new pull.
And I'll separate project for logical replication :)

PR #1266, opened 2017-04-17
Add semicolons to Pooling example in README.md

kas:
Minor fix to add semicolons to Pooling example code


brianc:
Thanks a bunch!! 💃

PR #1279, opened 2017-05-04
Make pool name consistent on missing config params

rochoa:
When using a config object to connect, if the config object misses some params, it will generate a pool name without those missing params. However, the next time, the object will have the missing params already in place with the default values, generating a different pool name.

By delaying the pool name generation, we ensure the name for the pool is consistent.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -54,7 +54,6 @@ PG.prototype.connect = function(config, callback) {
     callback = config;
     config = null;
   }
-  var poolName = JSON.stringify(config || {});
   if (typeof config == 'string') {
     config = new ConnectionParameters(config);
   }
@@ -66,6 +65,7 @@ PG.prototype.connect = function(config, callback) {
   config.idleTimeoutMillis = config.idleTimeoutMillis || config.poolIdleTimeout || defaults.poolIdleTimeout;
   config.log = config.log || config.poolLog || defaults.poolLog;
 
+  var poolName = JSON.stringify(config || {});
   this._pools[poolName] = this._pools[poolName] || new this.Pool(config);
   var pool = this._pools[poolName];
   if(!pool.listeners('error').length) {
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -66,6 +65,7 @@ PG.prototype.connect = function(config, callback) {
   config.idleTimeoutMillis = config.idleTimeoutMillis || config.poolIdleTimeout || defaults.poolIdleTimeout;
   config.log = config.log || config.poolLog || defaults.poolLog;
 
+  var poolName = JSON.stringify(config || {});
```
sehrope: Would it make sense to use something like [json-stable-stringify](https://www.npmjs.com/package/json-stable-stringify) here to ensure that matching configs always resolve to the same key? Otherwise there's no guarantee of the order of the keys in the serialized JSON.

I'd imagine that most people have a single place in their code that initializes their connection pool so it's likely nobody has really run into this, but if you have two separate spots and they specify the keys in a different order, they'd map to a different pool:

```js
const a = {foo: 123, bar: 456}
const b = {bar: 456, foo: 123}
JSON.stringify(a)
// '{"foo":123,"bar":456}'
JSON.stringify(b)
// '{"bar":456,"foo":123}'

const stableStringify = require('json-stable-stringify')
stableStringify(a)
// '{"bar":456,"foo":123}'
stableStringify(b)
// '{"bar":456,"foo":123}'
```

rochoa: Do you want me to upgrade this PR including that dependency? Or should we use another PR to do that? I prefer to use another PR as it's a separate issue, but I'm fine managing that one 😄 .

lenemson: You do not need to `|| {}` anymore since it's already done above.

sehrope: I'd suggest doing it in two commits submitted as one PR (so you can just update this PR). The first would move the location of JSON.stringify call. The second would add json-stable-stringify and replace the stringify call with the stable one.

rochoa: True that! Thanks a lot 👍 .

brianc:
This is great! Thanks for the help on this! 💃   I'm re-running the tests on travis so it'll turn green. Then I'll merge & push out a patch version!

brianc:
Alrighty! Sorry for the delay on this.  Merging now! Thanks so much for your help. :heart:

brianc:
published `pg@6.2.3`

PR #1285, opened 2017-05-10
Libpq connection string escaping

sehrope:
While silly, the following are perfectly legal identifiers for a username and database:

```sql
-- Username contains a single quote
postgres=# CREATE USER "dumb'user" WITH PASSWORD 'abcd1234';
CREATE ROLE

-- Database name contains a backslash
postgres=# CREATE DATABASE "dumb\db" WITH OWNER "dumb'user";
CREATE DATABASE
```

The existing libpq config code blindly surrounds the passed in parameters using single quotes (ex: `foo => 'foo'`). That breaks down when there are backslashes or single quotes in the values.

This PR fixes handling of single quotes and backslashes in connection config values used to generate libpq connection strings by escaping any occurrences of those characters.

For most property names there is no change in the valid case (i.e. no funky chars) as they were already quoted. This just corrects handling when they have funky chars.

The two exceptions are `host`, and `hostaddr` which previously weren't escaped at all. There's no *valid* value for either of those that could contain a single quote or even whitespace so it was fine for the valid case. But that also means that you have a config injection if you pass an invalid value like `somehost foo=bar`. It's not really possible right now as there's also a DNS lookup which presumably would fail for such a host, but if it were ever removed (which may be a good idea as hostaddr is an optional field anyway and having it unspecified has some value as well) it'd be broken.

Either way, the new code also escapes/quotes them as well and I've updated a couple tests to take that into account.

### DIFF
--- a/lib/connection-parameters.js
+++ b/lib/connection-parameters.js
@@ -65,10 +65,15 @@ var ConnectionParameters = function(config) {
   this.fallback_application_name = val('fallback_application_name', config, false);
 };
 
+// Convert arg to a string, surround in single quotes, and escape single quotes and backslashes
+var quoteParamValue = function(value) {
+  return "'" + ('' + value).replace(/\\/g, "\\\\").replace(/'/g, "\\'") + "'";
+};
+
 var add = function(params, config, paramName) {
   var value = config[paramName];
   if(value) {
-    params.push(paramName+"='"+value+"'");
+    params.push(paramName + "=" + quoteParamValue(value));
   }
 };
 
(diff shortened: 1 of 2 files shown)


brianc:
This is fantastic @sehrope - thank you!  Would you consider this to be a breaking change or a bug fix?

sehrope:
I'd consider it a bug fix. I can't imagine anyone relying on the old broken functionality.

brianc:
Okay awesome! thank you @sehrope - will merge it in shortly!

brianc:
published `pg@6.2.2` :clap:   Thanks for your work on this @sehrope and the really wonderful description of the initial behavior and problem. A pleasure to work with.

sehrope:
@brianc Anytime! And thanks more generally for the awesomeness that is this driver. Been using it happily for years and it's always a pleasure to contribute to it.

Issue #1292, opened 2017-05-17
Client#query hangs if try to get promise result after query completion


ikokostya:
**Versions**

* nodejs:  `7.10.0`
* node-postgres: `6.1.2 `

**Code example**

```js
'use strict';

const pg = require('pg');

const config = {
    user: 'test-user',
    password: 'test',
    database: 'test-db',
    host: 'localhost',
    port: 5432,
};

const client = new pg.Client(config);
client.connect(err => {
    if (err) {
        throw err;
    }

    const query = client.query('select 1');

    setTimeout(() => {
        console.log('after timeout');
        query
            .then(result => {
                console.log(result.rowCount);
                client.end();
            })
            .catch(err => console.error(err));
    }, 50);
});
```

**Expected behavior**

Prints in console:

```
after timeout
1
```
and exit.

**Actual behavior**

Prints

```
after timeout
```
and hangs.

***

**Why this happen**

`Query` object returns `Promise` only when [then](https://github.com/brianc/node-postgres/blob/ce8f215c88583d82b0dde22743f4fdc688942d9b/lib/query.js#L43) method will be called. But at this moment query may be already executed. That's why [`end` or `error` events](https://github.com/brianc/node-postgres/blob/ce8f215c88583d82b0dde22743f4fdc688942d9b/lib/query.js#L54-L55) will not be emitted and resulting promise always will have `pending` state.

**Possible solution**

~~Promise can be created in `Query` constructor.~~
*Edited:* see https://github.com/brianc/node-postgres/pull/1293#issuecomment-302665962

ikokostya:
@brianc Could you take a look?

ikokostya:
This issue fixed in `pg v7`.

viki53:
How about also fixing this is others versions too, like `6.4`?

sehrope:
I'm not sure if that would be possible. There were some major changes in the return type handling for v7 that can't be backported as they're potentially breaking changes. Hence the major version bump to v7.

PR #1293, opened 2017-05-19
Fix getting promise result after query completion

ikokostya:
Fixes: #1292


brianc:
Yeah I agree with you - overloading the query method to return a promise PLUS a query object was a bad API decision I'm going to roll back out in v7.0 - I'm going to push promises into a different file in the repo so you can do something like `require('pg/promise') to get the promise version of API out.

charmander:
`Client#query` returns a `Promise` instead of the query object in pg 7.

Issue #1294, opened 2017-05-19
Config object mutation


lenemson:
There is a mutation on the config object in [PG.prototype.connect](https://github.com/brianc/node-postgres/blob/master/lib/index.js#L65) at lines 65 to 67, it is problematic because the config object is [serialized](https://github.com/brianc/node-postgres/blob/master/lib/index.js#L57) with `JSON.stringify` and then [used as key](https://github.com/brianc/node-postgres/blob/master/lib/index.js#L70) to store/retrieve the clients pool.

If the config object passed to `pg.connect` doesn't contains either `max`, `idleTimeoutMillis` or `log`, they are added by mutation but AFTER the serialization, so the freshly created pool is stored with the non mutated  serialized config object as a key, causing all subsequent calls to `pg.connect` to use the mutated config object as a key thus using a second pool, letting the first pool unused, forever.

This change has been introduced in the code with this [commit](https://github.com/brianc/node-postgres/commit/796a44f54f2bd65611ae3ca09f3fb7d830a95ab4).

lenemson:
I submitted a tiny PR but it seems that another PR submitted 15 day ago fix this issue https://github.com/brianc/node-postgres/pull/1279.
It's the same fix but with a test, so lets merge this ?

brianc:
Gotcha! Sorry for the delay!

PR #1295, opened 2017-05-19
Fix pg.connect pool storage

lenemson:
Fixes: #1294 by serializing the config object after all mutations that affect it.
May be better to non mutate params, but it will require more changes.


charmander:
#1279

Issue #1274, opened 2017-04-24
Translate pg sql NULL to javascript undefined


arusakov:
I don't know, maybe current translation `NULL` to `null` have some reasons, but from my point of view `undefined` is more useful, because `undefined` omitted in `JSON` automatically:
```js
JSON.stringify({ x: null }) // {"x":null}
JSON.stringify({ x: undefined }) // {}
```

@brianc
What do you think about it?

brianc:
I don't think it's a _bad_ idea at all, but it has potential to be a very subtle very breaking change if people are relying on strict equality checks for `null` on some of their query results.  I think at this point we should consider the ship sailed on using `null` instead of `undefined` for result values & then if you want to strip them off in your own app you could do something like `rows.map(row => _.omit(row, (val, key) => val === null` or something?  That's a kinda unfortunate side-effect of having to empty values in JavaScript.  

I tagged to version 7.x so I can think about it when I start working on some backwards incompat changes to the next breaking version....but I'm leaning towards "that is how it is" right now.

charmander:
`null` is the correct value. If you want to omit it from JSON, that’s easy enough:

```js
const replaceNull = (key, value) =>
    value === null ?
        undefined :
        value;

JSON.stringify({ x: null, y: 1.2 }, replaceNull) // {"y":1.2}
```

arusakov:
@charmander 
Do you have real benefits of `null` usage? I don't know examples.
Also I know projects, that don't use `null` at all. For example [TypeScript](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined)

@brianc 
Maybe my proposition can this be achieved through configuration?

charmander:
I use `null` as an intentional “nothing” value, since `undefined` tends to be a mistake (given it’s what you get by reading a property that doesn’t exist).

Strongly −1 on implementing this, as a setting or otherwise.

abenhamdine:
Also -1 on this, null is imho the correct value. As charmander said, undefined is more interpreted as the absence of a given property than the absence of value for this property.

rpedela:
NULL is a real value in the database and has a specific meaning. We shouldn't be tampering with data in a database client library. You can write your own custom type parser(s) if you want to override how NULL is handled.

vitaly-t:
https://www.youtube.com/watch?v=lITBGjNEp08

jfsimon:
It would be a huge BC break!

caub:
[...]

PR #1296, opened 2017-05-24
nullValue parametrizable

caub:
a first attempt for #1274, node-pg-native could have a similar patch too

but I think a complete change to undefined, in a major release would be better

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -60,6 +60,8 @@ var Connection = function(config) {
 
 util.inherits(Connection, EventEmitter);
 
+Connection.nullValue = null;
+
 Connection.prototype.connect = function(port, host) {
 
   if(this.stream.readyState === 'closed') {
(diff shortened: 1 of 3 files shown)


brianc:
I can see the value of letting this be overridden as right now it's kinda deeply embedded in the code and impossible to change.  However - I don't think this is the right place to put this change - it should probably go [here](https://github.com/brianc/node-postgres/blob/master/lib/connection.js#L553) and I'd set it up to allow something like:

```
var Connection = require('pg/connection')
Connection.nullValue = undefined
```

However, I'd need to profile to make sure this doesn't impact field value parsing as that's really hot-path code & v8 can bail out of the JIT for weird reasons.  

Also, the discussion around changing from `null` to `undefined` as the _default_ returned is over. We're never switching the _default_ from `null` to `undefined`.  That ship sailed long ago. :boat:

caub:
great idea, currently it outputs `[{"id":"u1","name":"user1"},{"id":"u2","name":""},{"id":"u3","name":"undefined"}]` though

charmander:
This feature can’t be implemented as a module-global setting, because it could break dependencies; it would need to be per-client. And it needs to have tests.

PR #1303, opened 2017-05-27
Properly insert buffers in arrays.

2Pacalypse-:
Before this commit, when someone tried to insert a Buffer into an array,
the library would try to escape it (by calling the `escapeElement` on
it), which would fail because buffers don't have a `replace` method.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -31,6 +31,9 @@ function arrayString(val) {
     else if(Array.isArray(val[i])) {
       result = result + arrayString(val[i]);
     }
+    else if(val[i] instanceof Buffer) {
+      result += '\\\\x' + val[i].toString('hex');
+    }
     else
     {
       result += escapeElement(prepareValue(val[i]));
--- a/test/unit/utils-tests.js
+++ b/test/unit/utils-tests.js
@@ -126,6 +126,13 @@ test('prepareValue: date array prepared properly', function() {
   helper.resetTimezoneOffset();
 });
 
+test('prepareValue: buffer array prepared properly', function() {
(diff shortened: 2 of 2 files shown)


brianc:
@2Pacalypse- thank you so much for this! Would it be possible to write a test to assert this is working as correctly? Unfortunately we can't accept patches without tests because the risk of regression is too high otherwise!

2Pacalypse-:
I added the test, although I have to rely on the Travis CI to run the tests since I'm on the Windows machine ^^.

Also, not sure how to make the test pass for all the versions of node since `Buffer.from()` has been added in node >v4 and `new Buffer()` is deprecated in newer versions.

brianc:
@2Pacalypse - thanks for adding the test! just curious - what's stopping the test from running on windows?

2Pacalypse-:
Makefiles are not supported on Windows, afaik. I guess I could try running them on windows subsystem for linux or something, but Travis CI did the trick just fine in this case :d.

brianc:
Cool - just curious. If it's just the Makefile I'm pretty sure I can get rid of that in the near future & make testing easier on windows.

2Pacalypse-:
Yeah, makefiles should not be needed and just using the npm scripts in cross-env way should suffice, I think.

Anyway, I've adjusted the test so it works across all supported node versions. Should be ready to merge once you review it.

brianc:
@2Pacalypse- thanks a ton for this! And I apologize for taking so long to getting back to this as well.  I'll release a new patch version right now with your changes. 👏

PR #1313, opened 2017-06-09
Remove fallbacks for unsupported Node versions (#1304)

brianc:
* Remove unsupported Node versions 0.10 and 0.12 from CI

* Replace deprecated Buffer constructor with .from/.alloc

* Remove Promise polyfill

* Make use of Object.assign

* Remove checks for versions of Node earlier than 4

* Remove Buffer#indexOf fallback for Node 0.10


brianc:
I tried a vscode plugin to create a pull request directly from a branch I was working on into the `7.0` branch.  Unfortunately that plugin went crazy and just pushed some of my commits to the tip of your pull request branch which sucks...

Lesson learned: don't use that vscode plugin, create pull requests from the cli with `hub`.  Anyways - this PR now has additional changes in it on 056f42c but once the tests pass we should be good to merge all of this into the 7.0 branch.

Issue #1075, opened 2016-07-11
Handling connection timeouts / dropped network / db failover


jessedpate:
Hello,

I'm working on setting up a health check within our system, and have noticed that our Postgres implementation doesn't seem to handle connection timeouts, dropped network, or a db failover event very gracefully (at least in the way it works when using a multi-az deployment in Amazon RDS). The issue is very similar to one that was fixed in the mysql package a while back -- https://github.com/mysqljs/mysql/issues/821.

We effectively have a class that looks similar to this:

``` js
const Promise = require('bluebird');
const pg = require('pg');

class PostgresSQLInterface extends Service {
    constructor(dbConnectionOptions, logger) {
        const defaults = {
            host: 'localhost',
            port: '5432',

            user: '',
            password: null,

            max: 10, // Poolsize
            min: 10,

            keepAlive: true,
            idleTimeoutMillis: 10000
        };

        this.state = {
            config: _.merge({}, defaults, dbConnectionOptions)
        };

        this.createDatabasePool();
    }

    createDatabasePool() {
        this.state.pool = new pg.Pool(this.state.config);

        this.state.pool.on('connect', () => {
            this.state.healthy = true;
            console.log('HEALTHY');
        });

        this.state.pool.on('error', () => {
            this.state.healthy = false;
            console.log('UNHEALTHY');
        });
    }

    *getDatabaseConnection() {
        try {
            const connection = yield this.state.pool.connect();

            return connection;
        }
        catch (error) {
            throw new Errors.DatabaseError("Could not connect to Postgres", {
                originalError: error
            });
        }
    }    
};
```

In a normal environment, when our application connects, I see it spit out "HEALTHY" x the number of connections it made, as expected. However, there are a few issues:

1.) If the connection is severed (I turn off my wireless, kill my VPN, or trigger a reboot on the RDS instance), no error events are raised even though the documentation for pg-pool states an error is "Emitted whenever an idle client in the pool encounters an error. This is common when your PostgreSQL server shuts down, reboots, or a network partition otherwise causes it to become unavailable while your pool has connected clients." My expectation would be to see a string of "UNHEALTHY" statements.

2.) Similarly, if I initiate a failover in RDS, no error events are raised but I am unable to run queries. This may be due to how AWS handles the DNS entries around multi-az deployments, but I cannot test this while issue #1 remains unresolved.

[...]

PR #1316, opened 2017-06-09
Emit error when backend unexpectedly disconnects

### LINKED ISSUE #967: LISTEN notifications stop after a while
I'm using `LISTEN` to be notified of changes. It used to work perfectly when I was listening to a DB on `localhost`. However, now that I've moved the DB over to another VM, I'm encountering some problems where my client loses "sync" with the DB and stops receiving notifications after some time (in order of minutes/hours). 

I'm not using the connection pool, here's how I connect and listen to updates: 

```
var client = new pg.Client(pgConString);
client.connect(function(err, client) {
    if(err) {
      throw err;
    }
    client.on('notification', function(msg) {
        console.log(msg.payload);
    });
    client.on('error', function(error) {
      console.error('This never even runs:', error);
    });
    var query = client.query("LISTEN watchers");
});
```

Is it possible keep this listener working indefinitely? Am I doing something silly? What is supposed to happen after a momentary network failure? Do I have to resign to polling the DB and diffing changes?

### LINKED ISSUE #1113: No error emitted on ungraceful db server shutdown
When a database server goes down unnoticed, a single connected client will not emit any errors and continue living up to ~15 minutes until (in a specific case, 'EHOSTUNREACH') socket error is emitted. Setting keepAlive to true in client configuration does not have the desired effect, because the connection object sets _ending to true after sending off the first query (https://github.com/brianc/node-postgres/blob/b1b2801c710d550f2b10910dfd49ad51bbf569fd/lib/query.js#L96), which in turn forces 'ECONNRESET' socket events to be ignored (https://github.com/brianc/node-postgres/blob/master/lib/connection.js#L68).

brianc:
Before this patch when the socket to the PostgreSQL server was terminated unexpectedly the client would emit a `close` event but not emit an `error` event.  This means all clients in the pool would silently close without a user having any way to respond.  This several problems around database failovers, silently "hanging" applications, etc.

fixes #1113 #1075 

also possibly fixes #967

I don't think we need to wait for 7.0 for this patch because it is fixing an incorrect behavior which causes silent application failures: it does not change any external APIs...so I think this is a backwards compatible change except sometimes applications which previously silently disconnected without reconnect or became unresponsive will now properly emit errors.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -31,6 +31,9 @@ var Client = function(config) {
   var c = config || {};
 
   this._types = new TypeOverrides(c.types);
+  this._ending = false;
+  this._connecting = false;
+  this._connectionError = false;
 
   this.connection = c.connection || new Connection({
     stream: c.stream,
@@ -50,6 +53,7 @@ util.inherits(Client, EventEmitter);
 Client.prototype.connect = function(callback) {
   var self = this;
   var con = this.connection;
+  this._connecting = true;
 
   if(this.host && this.host.indexOf('/') === 0) {
     con.connect(this.host + '/.s.PGSQL.' + this.port);
@@ -107,6 +111,7 @@ Client.prototype.connect = function(callback) {
   //hook up query handling events to connection
   //after the connection initially becomes ready for queries
   con.once('readyForQuery', function() {
+    self._connecting = false;
 
     //delegate rowDescription to active query
     con.on('rowDescription', function(msg) {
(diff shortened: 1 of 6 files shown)


brianc:
note: the native bindings behaved correctly and emitted errors on unexpected disconnections - only the pure JS driver needed changing.

brianc:
To simulate network partitions and backend disconnections I ended up implementing a [fake PostgreSQL Server backend](https://github.com/brianc/node-postgres/pull/1316/files#diff-1bec4c6b73d41cf90ecf25c7c2c85038R8) I could run within the integration test.

Once this was implemented I could connect node-postgres clients to it and force terminations of the fake server.  After I did this I was able to immediately reproduce the issue of clients 'dying' without emitting an error.

marshallford:
This patch fixes the problem described for my app. Thanks!

okischuang:
Great efforts @brianc !!

Comparing to this, what I've done is implementing a setTimeout timer binding on connection instance, let it emit an error and then let error be fired through the path now the library already has.
connection -> client -> pool -> pool destroy

what do you think of this solution? is it helpful or appropriate to the issue?

Appreciate for the comments if you would like to give :>

PR #1317, opened 2017-06-09
deprecate `pg.connect`

brianc:
Deprecate `pg.connect` with a warning, to be removed fully in `pg@8.0`


brianc:
okay something bad happened on this branch. Trying again w/ a new PR.

PR #1318, opened 2017-06-09
Deprecate pg.* singleton methods

brianc:
This deprecates `pg.connect` `pg.end` and `pg.cancel` in favor of using instances of pools instead of a singleton.

They wont be fully removed until `pg@8.0` and haven't been in the documentation since `5.x`.


lib/index.js
```diff
@@ -28,6 +29,7 @@ var PG = function(clientConstructor) {
 util.inherits(PG, EventEmitter);
 
 PG.prototype.end = function() {
+  deprecate('pg.end() is deprecated - please construct pools directly via new pg.Pool()');
```
charmander: [`util.deprecate`](https://nodejs.org/api/util.html#util_util_deprecate_function_string)?

brianc: Holy smokes! :smoking: I didn't even know about that!  Changing to that now.

brianc:
Thanks for the review! :clap:

PR #1321, opened 2017-06-11
Added MIT License

amilajack:


brianc:
Thanks @amilajack! I'm going to make a few edits to the license so it's exactly what's reflected in the `README.md` then I'll merge it in. :)

PR #1323, opened 2017-06-14
Promisify

brianc:
This is an initial crack at promisifying `client.query` `client.end` and `client.connect` which are the three main async methods.  As we discussed this changes the API to return a promise if no callback is provided instead of a query/promise hybrid.

I haven't really refactored a lot of the tests in over 5 years and some of them were really grating on me how hard they were to maintain and convoluted...while doing so I turned up a few bugs.  Most were written before mocha existed, so it's this kinda weird home-grown test thing with very little to no flow-control.  I added a little test lib to help w/ flow control until I can find a nice light-weight one to use.  I want to hit the entire codebase with [prettier](https://github.com/prettier/prettier) this week so I don't have any more OCD ticks w/ the weird syntax that differs between files.  And upgrade from jshint to eslint since jshint is super old & doesn't have good tooling like eslint does.  Again...this lib has a lot of legacy baggage I'm just now getting around to cleaning up.

I know this isn't the most beautiful pull request but I was in :godmode: all day & so that's how it goes.


charmander:
Merging master into 7.0 would make the diff a bit more accurate here.

Issue #1299, opened 2017-05-24
move promise-based API to separate file


brianc:
Conflating promises & callbacks in the exact same API was a mistake. :frowning: I'm going to move the promise based versions of `client.query` & `client.connect` (and any other overloaded method) out into a separate file within `pg`.  I still like shipping a slim version with the repo, because it helps me keep everything in lock-step and handle some of the weird edge-cases with tests, but they need to be separated into their own API surface area and only included if the user specifically does something like `require('pg/promise')` or something - haven't figured out the API details yet.

I'm also going to give a bit of a look at `pg-pool` and see which methods we might need to de-promisify or split out into a promise & non-promise version.  

tl;dr - callback + promise api on the same method = bad news bears. :finnadie:

vvo:
> tl;dr - callback + promise api on the same method = bad news bears.

Yes we had the same issue on our JavaScript client at Algolia.

Nowadays our plan is to ship only a promise based solution and delegate callback based to another package (require('algoliasearch-cb')) OR to user land (callbackify(algoliasearch)).

The reason for this is that we believe the promise based flow control AND the introduction of the async/await keywords (that are available today in many many environment) is a better async flow control mechanism than callback based.

Was my 2 cents :D

brianc:
Thanks for the input @vvo! Yeah - I really like the `async/await` stuff too & use it in all my own stuff & client work which is what prompted me to put in promises initially.  btw: aloglia is pretty sweet - do y'all use node-postgres there?

brianc:
@charmander @joskuijpers 

Howdy! Pinging y'all since y'all have both mentioned the promise thing before.  I wanted to discuss briefly how to fix the promises API.  Within node-postgres there is only one place where a promise is used - the [Query](https://github.com/brianc/node-postgres/blob/master/lib/query.js#L43) returned from a call to `client.query` implements the promise api, though it's not exactly a promise itself.  I should have stopped and thought more before doing that in v6.0 because it's [caused some issues](https://github.com/brianc/node-postgres/pull/1293), and overloading the `client.query` function _even further_ was just a bad move.  So I have a few different ideas on how to handle this going forward...any of these changes are going to break backwards compatibility so it'll be in 7.0:

[...]

PR #1328, opened 2017-06-18
Remove methods deprecated in 6.x

brianc:
- pg.connect
- pg.end
- pg.cancel
- client.query().on('')
- client.query().promise() (I don't think this one was ever documented)


brianc:
Closes #1299
Closes #1293

Issue #1332, opened 2017-06-20
DeprecationWarning: client.query, even when following the upgrading guide


yoaquim:
Seeing this error 
```
(node:1) DeprecationWarning: Using the automatically created return value from client.query as an event emitter is deprecated. Please see the upgrade guide at https://node-postgres.com/guides/upgrading
```
with this code:

~~~js
let client = postgresClient.getClient(event.stageVariables);
let query = {
    name: 'get-user-by-id',
    text: queries.USERS('GET_USER_BY_ID'),
    values: [event.queryStringParameters.id],
    rowMode: 'array'
};

client.connect((error) => {
        if (error) {
            callback(null, {
                statusCode: '200',
                body: { 'error': error},
                headers: {
                    'Content-Type': 'application/json'
                }
            });
        }
});

client.query(query)
        .then(results => {
            console.log(results);
            let user = results.rows[0];
            callback(null, {
                statusCode: '200',
                body: JSON.stringify(user),
                headers: {
                    'Content-Type': 'application/json'
                }
            });
        })
        .catch(error => {
             callback(null, {
                statusCode: '200',
                body: { 'error': error},
                headers: {
                    'Content-Type': 'application/json'
                }
            });
        })
        .then(() => {
            client.end();
        });
~~~

`postgresClient` module:
~~~js
const { Client } = require('pg');
module.exports = {
    getClient: (stageVariables) => {
        const config = {
            host: stageVariables.DB_ADDRESS,
            port: 5432,
            database: stageVariables.DB_NAME,
            user: 'dbadmin',
            password: stageVariables.DB_PASS
        };
        return new Client(config);
    }
}
~~~

And, AFAIK, I'm following the guidelines on the upgrading doc and regular docs.

I'm running this on an AWS Lambda, but not sure why that would matter for this specific case.

From `package.json`:
```
...
  "dependencies": {
    "pg": "^6.2.4"
  }
...
```

(weird how I'm not using v7, even though it's taking me to the upgrading docs for v7?

Ideas?

brianc:
I'm looking into this right now - judging by your code you shouldn't be getting that warning.

brianc:
it should_only_ warn if you use `query('foo').on('event')` style

yoaquim:
@brianc thanks for the quick turnaround!

yeah – that's exactly what I would expect!

Still digging a bit, maybe I missed something!

brianc:
[...]

PR #1333, opened 2017-06-20
Fix over-eager deprecation warnings

brianc:
Deprecation warnings were being printed even when not using `query.on`.


brianc:
Fixes #1332

Issue #906, opened 2015-12-24
Character encoding of the database


alexvoronov:
Hi,

I can't figure out how to specify character encoding of the database. My database uses `iso-8859-1` instead of `utf8`. All the strings I get from the database are considered `utf8`, and contain some "strange" characters. Is there any way to specify the encoding for automatic conversion?

I use the following connection string:
```
"postgres://myusername@localhost/mydbname"
```

I tried the following with no success:
```
"postgres://myusername@localhost/mydbname?encoding=iso-8859-1"
```

Thanks!
Alex

alexvoronov:
I got a workaround: I "surround" my real query in the following SQL query: 
```javascript
client.query("SET client_encoding to 'latin1';" , function(err, empty_result_to_fix_encoding) {
  client.query("SELECT ...", function (err, result) { /* ... */ });  // "Real" query goes here.
});
```

diegoperini:
Client or Pool configuration object is not very well documented. Complete (afaik) accepted params are like these I suppose. I'm using Node.js 5.9+.

```
const pg = require('pg');
const Pool = require('pg-pool'); // pg.Pool was undefined for some reason
const Promise = require('bluebird'); // or some other Promise implementation
const Client = pg.Client;

config = {
  user: 'postgres',
  password: '123456',
  host: 'db.mydomain.com',
  port: 5432,
  database: 'myDb',
  client_encoding: 'utf8', // This is not documented, I discovered it via inspecting "https://github.com/iceddev/pg-connection-string"
  ssl: true,
  max: 20,
  min: 1,
  idleTimeoutMillis: 1000,
  Client: Client, // This is client constructor, can be interchanged with native client
  Promise: Promise // This is internally used promise constructor
};

var pool = new Pool(config);

// Go on with your new pool
```

PR #1338, opened 2017-06-21
Fix for server enconding when using SQL_ASCII and latin1 enconding

ederavilaprado:
Encodes the string as the same as the `client_encoding` when provided.

This is a way to fix problems when ussing characters outsite the regular ASCII table, like the char `°`, who is the number 176 decimal at the ISO-8859-1 ASCII table (https://en.wikipedia.org/wiki/Degree_symbol#Encoding).

Example using the same char:

```javascript
> new Buffer([176])
<Buffer b0>
> new Buffer([176]).toString('utf8')
'�'
> new Buffer([176]).toString('latin1')
'°'
```


JCornat:
It is superfluous, but shouldn't `client_encoding` parameter be in CamelCase, like `clientEncoding`, for consistency with the whole project ?

ederavilaprado:
I actually kept this name `client_encoding` because this already exists in the config object (is used when creating the connection string).

I found this when reading this comment: https://github.com/brianc/node-postgres/issues/906#issuecomment-260345891

It seems easier to reintegrate without generating side effects.

brianc:
Thanks for putting this together @ederavilaprado! Really appreciated.

@JCornat I see what you're saying about camelCase.  That's something we might be able to change in a semver major change, since it'd subtly break in a patch release, but I'm open to it.  What do you think?

JCornat:
Indeed (https://github.com/ederavilaprado/node-postgres/blob/297f90cc8143297aa2937f61ac88b10b33fb3c21/lib/connection-parameters.js#L59), sorry for my comment !

brianc:
Hey @ederavilaprado - would you have a self-contained script or something I could use to make a test for this feature?  I don't want to accidentally break this in the future because there's no test for it.  I can add the test myself I just want to see a script demonstrating a different encoding & checking to make sure it deserializes correctly.

JCornat:
@brianc Maybe it is not a bad thing to keep it how it currently is, my comment is just unnecessary

brianc:
Published new version at `pg@6.4.0`! 

@JCornat okay cool. 😄 thanks for your help!

ederavilaprado:
Hey @brianc, i will create something ASAP to help you with the tests... i'm not sure if i can do this before friday unfortunatelly.

I'll let you know as soon as I have something ready.

PR #1347, opened 2017-06-26
use consistent syntax for semver ranges

karlbohlmark:
The mixing of `1.*` and `1.x` syntax confused me a little at first.


brianc:
Thanks @karlbohlmark

Issue #1358, opened 2017-07-11
Missing support for Certificate Authentication?


wzrdtales:
Hey together,

currently trying to connect with Certificate Authentication, I do get however denied by the server, as the postgres client seems to try password authentication anyway. I have followed so far https://github.com/brianc/node-postgres/wiki/SSL-support to configure the client.
Is certificate authentication supported though?

At least it looks like it is always added, no matter if it is actually set or not:

https://github.com/brianc/node-postgres/blob/master/lib/connection-parameters.js#L83

wzrdtales:
ok @brianc did a bit of verification here my self. If I use the native bindings, everything works ok.

Here are some results:

 - Does not work with the javascript version of node-pg
 - Does work with the native bindings
 - The native bindings do not pass sqlrootcert correctly for some reason, resulting in verify-ca modes to always fail

wzrdtales:
Ok, have found the bug concerning libpq unavailability of sslrootcert. Seems like it has been forgotten over here: https://github.com/brianc/node-postgres/blob/master/lib/connection-parameters.js#L92

timothyerwin:
@wzrdtales I'm following the example from here https://node-postgres.com/features/ssl but I'm still getting `error: password authentication failed for user "xxxxxxx"`

It seems like it ignores the `ssl` option and tries to user/password method?

timothyerwin:
I was able to get it to work using the following:

```javascript
const fs = require('fs');

const pg = require('pg').native;

pg.defaults.ssl = true;

const { Pool } = pg;

const secrets = require('../secrets');

const { database, host, user, password } = secrets.db;

const options = {
  database,
  user,
  password,
  host,
  port: 5432,
  ssl: {
    sslmode: 'verify-ca',
    sslrootcert: 'server-ca.pem',
    sslkey: 'client-key.pem',
    sslcert: 'client-cert.pem',
  }
};

const pool = new Pool(options);

(async () => {
  await pool.query('SELECT NOW()');
})();
```

chriszrc:
I see this was merged here:

https://github.com/brianc/node-postgres/pull/1359#issuecomment-325149350

Does this mean that this issue is resolved now?

PR #1359, opened 2017-07-11
add missing sslrootcert for native bindings

wzrdtales:
This fixes an error when using the native bindings, where the rootcert can never be found as it is just simply not passed by.

refers to #1358

### DIFF
--- a/lib/connection-parameters.js
+++ b/lib/connection-parameters.js
@@ -91,6 +91,7 @@ ConnectionParameters.prototype.getLibpqConnectionString = function (cb) {
   add(params, ssl, 'sslca')
   add(params, ssl, 'sslkey')
   add(params, ssl, 'sslcert')
+  add(params, ssl, 'sslrootcert')
 
   if (this.database) {
     params.push('dbname=' + quoteParamValue(this.database))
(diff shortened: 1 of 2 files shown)


brianc:
Thanks for the PR @wzrdtales !  Would you be open to adding a super quick test (or extending an additional one) [here](https://github.com/brianc/node-postgres/blob/ee8193673c551e5f11d46150aa4b989f64fd95c2/test/unit/connection-parameters/creation-tests.js) to just ensure its doing the right thing?  While it might seem trivial, every line of code without tests is something I have to carefully dance around or write my own tests for in the years to come, so it's just as helpful as the original code.

I appreciate both! 🤗

wzrdtales:
Sure thing @brianc, I know the importancy of test though :). Have had a look at your tests, but it did not seem like there would have been anything appropriate where I could have placed this. Especially as this has a dependency to the native module, therefore I wanted to wait for your feedback first. 

I will add a simple test for that later though.

wzrdtales:
@brianc I will add that test now shortly, but just to let you know the whole test file. I left that untouched, if you want me to submit a fixed version (I disabled prettier while editing this though) please let me know :)

wzrdtales:
Have pushed you a style fixed version here https://github.com/wzrdtales/node-postgres/commit/46f4326fb88e80d9776ed4bdf45914414c945f18. Just cherry-pick it if you want it :)

brianc:
Thanks for this! Sorry it took me a little while to get to it - going to go ahead & merge this one in & release it in the next minor version.  Appreciate it!

PR #1363, opened 2017-07-13
allow both connectionString and additional options

caub:
allows
```js
var pool = new Pool({connectionString: 'postgres://foo@dom.com/db', ssl:true, max:4})
// will be an ssl connection with this PR
```
connectionString params will still override other options to stay compatible

### DIFF
--- a/lib/connection-parameters.js
+++ b/lib/connection-parameters.js
@@ -11,6 +11,8 @@ var dns = require('dns')
 
 var defaults = require('./defaults')
 
+var parse = require('pg-connection-string').parse // parses a connection string
+
 var val = function (key, config, envVar) {
   if (envVar === undefined) {
     envVar = process.env[ 'PG' + key.toUpperCase() ]
@@ -24,9 +26,6 @@ var val = function (key, config, envVar) {
     envVar ||
     defaults[key]
 }
-
-// parses a connection string
-var parse = require('pg-connection-string').parse
 
 var useSsl = function () {
   switch (process.env.PGSSLMODE) {
@@ -43,12 +42,14 @@ var useSsl = function () {
 
 var ConnectionParameters = function (config) {
   // if a string is passed, it is a raw connection string so we parse it into a config
-  config = typeof config === 'string' ? parse(config) : (config || {})
+  config = typeof config === 'string' ? parse(config) : config || {}
+
   // if the config has a connectionString defined, parse IT into the config we use
   // this will override other default values with what is stored in connectionString
   if (config.connectionString) {
-    config = parse(config.connectionString)
+    config = Object.assign({}, config, parse(config.connectionString))
   }
+
   this.user = val('user', config)
   this.database = val('database', config)
   this.port = parseInt(val('port', config), 10)
--- a/test/unit/connection-parameters/creation-tests.js
+++ b/test/unit/connection-parameters/creation-tests.js
@@ -67,6 +67,28 @@ test('ConnectionParameters initializing from config', function () {
   assert.ok(subject.isDomainSocket === false)
 })
 
+test('ConnectionParameters initializing from config and config.connectionString', function() {
+  var subject1 = new ConnectionParameters({
+    connectionString: 'postgres://test@host/db'
+  })
+  var subject2 = new ConnectionParameters({
+    connectionString: 'postgres://test@host/db?ssl=1'
+  })
+  var subject3 = new ConnectionParameters({
+    connectionString: 'postgres://test@host/db',
+    ssl: true
+  })
+  var subject4 = new ConnectionParameters({
+    connectionString: 'postgres://test@host/db?ssl=1',
+    ssl: false
+  })
+
+  assert.equal(subject1.ssl, false)
+  assert.equal(subject2.ssl, true)
+  assert.equal(subject3.ssl, true)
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -4,3 +4,4 @@ node_modules/
 .lock-wscript
 build/
 *~
+package-lock.json
```
brianc: please remove this

package.json
```diff
@@ -19,8 +19,10 @@
   "main": "./lib",
   "dependencies": {
     "buffer-writer": "1.0.1",
+    "object-assign": "^4.1.1",
     "packet-reader": "0.3.1",
     "pg-connection-string": "0.1.3",
+    "pg-native": "^1.10.1",
```
brianc: this should not be a dependency

package.json
```diff
@@ -19,8 +19,10 @@
   "main": "./lib",
   "dependencies": {
     "buffer-writer": "1.0.1",
+    "object-assign": "^4.1.1",
```
brianc: this needs to be exactly `4.1.1` without the `^`

brianc:
Thanks @caub!  Put a couple comments up there.

caub:
ok, I think npm5 automatically adds pg-native, since it's required in the lib/ source

brianc:
ohhhh I bet it added it during the test run.  I'll need to add some kinda parameter to the test to make it not save when doing an `npm install` during the tests.

lib/connection-parameters.js
```diff
@@ -11,6 +11,9 @@ var dns = require('dns');
 
 var defaults = require('./defaults');
 
+var parse = require('pg-connection-string').parse; //parses a connection string
+var objectAssign = require('object-assign');
```
charmander: `Object.assign` can be used now that pg targets Node 4.

caub:
@charmander done, that's great to be node4+ now

lib/connection-parameters.js
```diff
@@ -43,12 +42,12 @@ var useSsl = function () {
 
 var ConnectionParameters = function (config) {
   // if a string is passed, it is a raw connection string so we parse it into a config
-  config = typeof config === 'string' ? parse(config) : (config || {})
+  config = typeof config === 'string' ? parse(config)
```
charmander: This might be a bit too compact. Maybe:

```javascript
config = typeof config === 'string' ? parse(config) : Object.assign({}, config)

if (config.connectionString) {
  Object.assign(config, parse(config.connectionString))
}
```
?

caub:
it's ok? I've used intermediate vars, because I feel it's better than mutating

lib/connection-parameters.js
```diff
@@ -41,14 +40,14 @@ var useSsl = function () {
...
-  config = typeof config === 'string' ? parse(config) : (config || {})
+  var cfg = typeof params === 'string' ? parse(params) : params || {}
+
   // if the config has a connectionString defined, parse IT into the config we use
   // this will override other default values with what is stored in connectionString
-  if (config.connectionString) {
-    config = parse(config.connectionString)
-  }
+  var config = typeof cfg.connectionString === 'string' ? Object.assign({}, cfg, parse(cfg.connectionString)) : cfg
+
```
caub: I've used intermediate vars, because I feel it's better than mutating

charmander: I’d still suggest the same snippet as earlier, since it always copies the configuration object.

lib/connection-parameters.js
```diff
@@ -43,12 +42,14 @@ var useSsl = function () {
...
 var ConnectionParameters = function (config) {
   // if a string is passed, it is a raw connection string so we parse it into a config
-  config = typeof config === 'string' ? parse(config) : (config || {})
+  config = typeof config === 'string' ? parse(config) : config || {}
+
   // if the config has a connectionString defined, parse IT into the config we use
   // this will override other default values with what is stored in connectionString
-  if (config.connectionString) {
-    config = parse(config.connectionString)
+  if (typeof config.connectionString === 'string') {
```
caub: @charmander ok done, I just check typeof, because parse would probably throw else

charmander: Throwing instead of hiding potential errors is generally good, and it already behaved that way. I’d keep this check unmodified.

caub: ok done

test/unit/connection-parameters/creation-tests.js
```diff
@@ -67,6 +67,30 @@ test('ConnectionParameters initializing from config', function () {
...
+    connectionString: 'postgres://test@host/db?ssl=1'
+  })
+  var subject3 = new ConnectionParameters({
+    connectionString: 'postgres://test@host/db',
+    ssl: true
+  })
+  var subject4 = new ConnectionParameters({
+    connectionString: 'postgres://test@host/db?ssl=1',
+    ssl: false,
+    max: 5
```
charmander: Is this supposed to work? Connection strings seem to work at the client level, whereas `max` is at the pool level.

If it’s not, it should probably be removed to avoid misleading people.

caub: yes done, I need to make a similar PR for pg-pool

brianc:
Thanks so much for this! Pushing out a new minor version w/ the changes now.

PR #1364, opened 2017-07-14
Cache parsed connection strings

vitaly-t:
Cache parsed connection strings, to avoid continuous re-parsing.

A typical client that uses a connection string is passing it in every time it is trying to connect. There is no point running all the parsing every time this happens, which is a lot in a busy environment.


charmander:
This is what pools are for.

PR #1365, opened 2017-07-14
Throw on reconnect attempt

### LINKED ISSUE #1352: Client re-connection doesn't work
I'm running a simple test against the latest `pg`, under Node.js 8.1.2:

```js
const pg = require('pg');

var client = new pg.Client(connection);

function test() {
    client.connect(err => {
        if (err) {
            console.log(err);
            return;
        }
        client.query('select 123', [], (err, data) => {
            if (err) {
                console.log(err);
            } else {
                console.log('DATA:', data.rows[0]);
            }
            client.end();
        });
    });
}

setInterval(test, 1000);
```

And it outputs the following, every time:

```
DATA: anonymous { '?column?': 123 }
DATA: undefined
DATA: undefined
DATA: undefined
DATA: undefined
DATA: undefined
DATA: undefined
DATA: undefined
events.js:182
      throw er; // Unhandled 'error' event
      ^

error: invalid frontend message type 0
    at Connection.parseE (D:\NodeJS\tests\node_modules\pg\lib\connection.js:567:11)
    at Connection.parseMessage (D:\NodeJS\tests\node_modules\pg\lib\connection.js:391:17)
    at Socket.<anonymous> (D:\NodeJS\tests\node_modules\pg\lib\connection.js:129:22)
    at emitOne (events.js:120:20)
    at Socket.emit (events.js:210:7)
    at addChunk (_stream_readable.js:252:12)
    at readableAddChunk (_stream_readable.js:239:11)
    at Socket.Readable.push (_stream_readable.js:197:10)
    at TCP.onread (net.js:588:20)
```

After I successfully call `client.end()`, it will successfully re-connect, successfully execute the query, just won't return any data. This looks odd.

And this - `error: invalid frontend message type 0` doesn't make any sense.

Could somebody, please explain, why we get such an odd output?

**UPDATE**

And if just after creating the client, I add the following:
```js
client.on('error', e => {
    console.log(e);
});
```
The output gets absolutely flooded with a huge error dumb, so big, I can't even paste it here.

brianc:
Clients are not reusable.  This changes the client to raise errors whenever you try to reconnect a client that's already been used.  They're cheap to create: just instantiate a new one (or use the pool) :wink:.

Closes #1352

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -34,6 +34,7 @@ var Client = function(config) {
   this._types = new TypeOverrides(c.types);
   this._ending = false;
   this._connecting = false;
+  this._connected = false;
   this._connectionError = false;
 
   this.connection = c.connection || new Connection({
@@ -54,6 +55,14 @@ util.inherits(Client, EventEmitter);
 Client.prototype.connect = function(callback) {
   var self = this;
   var con = this.connection;
+  if (this._connecting || this._connected) {
+    const err = new Error('Client has already been connected. You cannot reuse a client.')
+    if (callback) {
+      callback(err)
+      return undefined
+    }
+    return Promise.reject(err)
+  }
   this._connecting = true;
 
   if(this.host && this.host.indexOf('/') === 0) {
@@ -135,6 +144,7 @@ Client.prototype.connect = function(callback) {
   //after the connection initially becomes ready for queries
   con.once('readyForQuery', function() {
     self._connecting = false;
+    self._connected = true;
     self._attachListeners(con);
     con.removeListener('error', connectingErrorHandler);
     con.on('error', connectedErrorHandler)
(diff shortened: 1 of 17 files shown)


brianc:
This isn't _technically_ a backwards incompatible change in that previous the client would just raise weird errors from the backend when attempting to reconnect or silently hang if it was already disconnected.

sehrope:
+1 overall.

Regarding the implementation, would it make more sense to have a state type variable rather than separate boolean values? I haven't yet peeked at how many references there are to `_connecting` / `_connected` but if there's going to be additional states (ex: errored out) and they're mutually exclusive, would be easier to keep consistent than separate vars.

brianc:
yeah at some point I think having a state variable makes sense. In this case it's pretty small and since they're marked as private shouldn't be hard to refactor in the future if it gets unwieldy having a single one.

brianc:
Thanks for the review @sehrope! :heart:

PR #1367, opened 2017-07-15
Re-enable eslint with standard format

brianc:
No more semi-colons! :dancer:

Most of these changes were done with `standard --fix`.  There is still a bunch of cruft built up in `test/` directory that I'd like to clean out, but this is a good start and the code under `lib/` is fully checked.  Upgrading to eslint will also allow the addition of new and stricter rules in the future.


sehrope:
+1 to a clean slate with linting.

We should add a lint check to Travis as well. That way going forward all PRs have to be clean prior to being merged.

brianc:
Yah travis is running the lint now as well. It was on earlier production branches, I just disabled it during work in progress on 7.0.  Though up until 7.0 `jshint` was in use because at the time I added linting eslint wasn't in existence. :stuck_out_tongue_closed_eyes: 

Feels good to finally update to more modern version of node and eslint.  Good times ahead! :dancer:

Issue #508, opened 2014-01-17
Results from multiple statements query returned as single dimensional array


traderboy:
I have 3 queries I'm running as multiple statements, but the results are all returned in a single dimensional array.  The MySQL library returns multiple statement queries  in a multidimensional array.  The problem is I can't tell which results come from which query.

SQL: select id from table1;select id from table2;select id from table3;

PG:
row: [ {aaa},{bbb},{ccc}]

MySQL:
row: [ [{aaa}], [{bbb}], [{ccc}] ]

clayzermk1:
:+1: 

What is the current status of this feature? Are you still looking for pull requests?

matt212:
reference topic  
https://github.com/brianc/node-postgres/issues/1190#issuecomment-301101575

brianc:
Added this to the pg@7.0 release milestone. Long standing bug but it should be fixed.

brianc:
This is fixed in pg@7.0

vitaly-t:
While this does improve the driver nicely, for any library that works based off the results returned from the driver, it screws up everything, and requires major changes.

There is probably not a single library out there ready to handle multi-result sets. This can cause a havoc :smile:

alitaheri:
@vitaly-t It has, it broke sequelize badly, and we don't know how to handle multiple statements throughout the library 😅 😅 But it was needed. necessary evil 😈 

Thanks @brianc for this nice feature. 👍 👍 Hope we can migrate soon so others can use it from sequelize too 💟

vitaly-t:
@alitaheri same here, I'm stuck trying to figure out a good approach to supporting multiple results in `pg-promise`. I have already redesigned everything for that, but released nothing, as I'm not too happy about the new architecture, it seems too verbose. Here's me discussing some of it with the author of `massive.js`: https://github.com/vitaly-t/pg-promise/issues/371

PR #1369, opened 2017-07-16
7.0

brianc:
Notable breaking changes:

- drop support for node < 4.x
- `client.query` no longer returns an event-emitter.  If a callback is supplied, it returns undefined. If no callback is supplied it returns a `Promise.`  This was deprecated in `pg@6.x`
- `pg.connect` and `pg.end` global singleton pool was removed.  This was deprecated in `pg@6.x`
- `pg.cancel` which used the global singleton pool was removed. This was deprecated in pg@6.x`
- standardized error message fields between pure JavaScript driver and libpq driver.  Pure JavaScript error fields are used throughout.  This will only impact you if you were using `pg.native` and checking error feilds such as `error.context` or `error.where` [change](https://github.com/brianc/node-postgres/commit/056f42cd6a49c5e639f3247ce133b15c302b7ccd)
- Result from multi-statement plain text query now returned as multiple arrays instead of mangled into a single array. [context](https://github.com/brianc/node-postgres/issues/508)

Enhancements:
- better promisification of all APIs. connect, end, and query
- better support for network partitions & ensuring errors from connections are bubbled up
- overhaul of pg-pool internals & supporting `connectionTimeoutMillis` for a connection timeout feature in the pool


abenhamdine:
Amazing !

ashking:
Excellent work! this is exactly the kind of list I was looking for!

brianc:
@ashking I also maintain [this](https://github.com/brianc/node-postgres/blob/master/CHANGELOG.md) on every release above patch. :smile:

abenhamdine:
> Result from multi-statement text queries such as `SELECT 1; SELECT 2;` are now returned as an array of results instead of a single result with 1 array containing rows from both queries.

This one is the mega bonus / christmas gift 😄

Issue #1373, opened 2017-07-18
Use of Buffer.alloc() breaks Node 4.x compatibility


appdlh:
The package.json "engines" entry claims compatibility with Node >= 4.0.0, but around pg/lib/connection.js:290:26 there is a call to Buffer.alloc(), an API not introduced until Node 5.0.0. This causes the driver to throw when using Node < 5.x.

brianc:
All the tests are passing on travis for node `lts/argon` which I believe is node 4.x:

https://travis-ci.org/brianc/node-postgres/jobs/254701876

Should I increase the engines compatibility in `package.json` to something other than `4.0.0`?

brianc:
Looks like I can bump the engines up to `>=4.2.x`

https://github.com/nodejs/help/issues/28

What do you think?

appdlh:
According to the travis output the tests are running on Node v4.8.4, but that just tells me the tests don't expose this problem... as I said, Buffer.alloc() wasn't introduced until Node 5.0 (according to the API documentation at least).

If the driver is going to continue using 5.x APIs, the minimum engine requirement needs to be at least 5.x; if 4.x support is to be continued, that 5.x API call will need to be addressed.

charmander:
They were [backported in Node 4.5.0][1].

  [1]: https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V4.md#2016-08-15-version-450-argon-lts-thealphanerd

appdlh:
Ah, nice; OK, then 4.5.x should be a viable semver spec for the Node "engines" entry.

PR #1374, opened 2017-07-18
Update minimum Node version to 4.5.0

charmander:
`Buffer.alloc` and `Buffer.from` were only backported then. Addresses #1373.


brianc:
thanks! will push a new patch version now.

PR #1375, opened 2017-07-18
remove unused variable

billouboq:
Remove self since this can be used.

By the way, do you want to keep all the 'bind(this)' or can I use arrow functions since node 4.5 is needed with pg > 7 ?


charmander:
I think arrow functions would be better, yep!

brianc:
Thanks for the cleanup @billouboq!

PR #1379, opened 2017-07-20
Fixing invalid custom type check

vitaly-t:
The old check did nothing useful, and the code would fail if `null` or `undefined` were passed in.


brianc:
👍  thank you!

Issue #1382, opened 2017-07-21
client.end() doesn't always return a promise


bridgelabs:
client.end() doesn't return a promise when client.activeQuery exists.

From /lib/client.js:

```javascript
Client.prototype.end = function (cb) {
  this._ending = true
  if (this.activeQuery) {
    // if we have an active query we need to force a disconnect
    // on the socket - otherwise a hung query could block end forever
    this.connection.stream.destroy(new Error('Connection terminated by user'))
    return
  }
  if (cb) {
    this.connection.end()
    this.connection.once('end', cb)
  } else {
    return new global.Promise((resolve, reject) => {
      this.connection.end()
      this.connection.once('end', resolve)
    })
  }
}
```

brianc:
@bridgelabs thanks for the bug report.  I agree it's a bug!  I should have it fixed here in a few minutes.

PR #1384, opened 2017-07-21
Make client.end return promise with active query

brianc:
Previous it returned `undefined` by mistake, causing an exception to be thrown in certain cases when calling `client.end().then()`.


brianc:
This fixes #1382

brianc:
fix published in `pg@7.0.2`

PR #1406, opened 2017-08-12
Removes extra 'use strict'

sehrope:
Two of the test files have 'use strict' twice at the top. Probably from some copy pasta.


brianc:
Thanks! I think I went a bit bananas with sed at some point. 🍌 😁

brianc:
Thanks! I think I went a bit bananas with sed at some point. 🍌 😁

PR #1407, opened 2017-08-12
Fix result parser handling of escape sequences in column names

sehrope:
The dynamic code generator `inlineParser` in result.js handles escaping of single quotes but not other valid backslash escape sequences. In those situations it can either break entirely (SyntaxError due to generation of invalid JS code) or the query result doesn't match the actual column name in the SQL.

Here are some examples of SQL that break the parser:

```sql
-- # Newline in column name
-- Expected: JS value for column name should be "test\ing"
-- Result in node-postgres: SyntaxError in result.js parser
SELECT 1 as "test
ing";

-- # Backslash as column name
-- Expected: JS value for column name should be "\\"
-- Result in node-postgres: SyntaxError in result.js parser
SELECT 1 as "\";

-- # Backslash-T (i.e. tab escape sequence) as column name
-- Expected: JS value for column name should be "\\t"
-- Result in node-postgres: "\t"
SELECT 1 as "\t";

-- # JS escape sequence for a Unicode escape sequence in column name
-- Expected: JS value for column name should be "\\u00A9"
-- Actual result in node-postgres: "©"
SELECT 1 as "\u00A9";
```

For the record I think anybody would be nuts to have database columns like this but they're perfectly legit queries from PostgreSQL's perspective and should be handled.

This PR changes the inline parser to use the [js-string-escape](https://www.npmjs.com/package/js-string-escape) lib to handle the escaping.

I was hoping to add the examples above as tests but I'm having a bit of trouble wrangling the test suite infrastructure. I've got it working locally and this parser change fixes these issues, but had to make a bunch of unrelated changes to get the tests running so not including it.


brianc:
@sehrope thanks for the pull request! 

Sorry the tests are hard to write. :/  They originated before there was a solid test framework in node.js & I'm loathe to port them because there's a big risk moving working tests to a new framework (do the tests still work the same way? do they still test the same thing? will they still help prevent regressions?) So I did a bit to incrementally migrate them to something that more closely matches mocha to make them more familiar but they're not there yet.

Out of curiosity - what issues did you hit when trying to write tests? Are you on windows?  Each test file should be a self-executing & self-exiting node file. If it exits with `0` then the test is considered to pass and if it exits with a non-zero status code the test is considered to fail and the test suite will abort. This is pretty similar to how node core tests are written.

sehrope:
No I'm on Linux. Not really an issue of writing the test itself which is just running the sample queries I provided. More about setting up the config around the environment to run the tests, specifically getting the tests to point at a non-standard Postgres DB.

I don't run a Postgres DB directly on my computer, I run them in VMs with ports forwarded so it's not accessible at localhost:5432. For a single test case it's fine but getting the entire suite to run against the non-standard location requires editing things in multiple places.

I think any new test setup should handle that case by centralizing things at one place, say via an env variable, and then deferring to localhost:5432 if it's not defined. That way it'd be one place to change things that gets inherited by all tests.

brianc:
took issue to email, will resolve shortly

PR #1415, opened 2017-08-16
Remove unreachable default

charmander:
`match[3]` is always truthy in this block. Introduced in 102a069bd2360783eb106b4bbcfa93b3f7fd7030, for context if the correct fix is something else.

### DIFF
--- a/lib/result.js
+++ b/lib/result.js
@@ -40,10 +40,9 @@ Result.prototype.addCommandComplete = function (msg) {
   }
   if (match) {
     this.command = match[1]
-    // match 3 will only be existing on insert commands
+    // match 3 will only exist on insert commands
     if (match[3]) {
-      // msg.value is from native bindings
-      this.rowCount = parseInt(match[3] || msg.value, 10)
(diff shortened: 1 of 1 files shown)


sehrope:
@charmander I was checking this out this morning and noticed that `else` block of that check never gets executed. The result of the regex always populates `match[3]` even for situations where there's only two components to the command complete response:

```
> var matchRegexp = /([A-Za-z]+) ?(\d+ )?(\d+)?/
undefined
> matchRegexp.exec('INSERT 0 456')
[ 'INSERT 0 456',
  'INSERT',
  '0 ',
  '456',
  index: 0,
  input: 'INSERT 0 456' ]
> matchRegexp.exec('DELETE 123')
[ 'DELETE 123',
  'DELETE',
  undefined,
  '123',
  index: 0,
  input: 'DELETE 123' ]
```

Interestingly everything still works fine! It ends up pulling the rowCount from the right place and the oid field (`match[2]`) is null as the regex doesn't pick up a match for it. Pretty weird eh?

Check out #1419 for a patch for cleaning up the regex and code a bit.

PR #1419, opened 2017-08-19
Refactor addCommandComplete

sehrope:
Refactors addCommandComplete to:

* Tighten regex by adding a starting anchor and wrapping the whitespace separators in non-capturing groups
* Reordering extracted values to match order in regex with comments for sample format

The regex change makes it so that for two part commands (ex: `DELETE 123`) the resulting match will populate `match[1]` and `match[2]`:

```
> var matchRegexp = /^([A-Za-z]+)(?: (\d+))?(?: (\d+))?/
undefined
> matchRegexp.exec('DELETE 123')
[ 'DELETE 123',
  'DELETE',
  '123',
  undefined,
  index: 0,
  input: 'DELETE 123' ]
> matchRegexp.exec('INSERT 456')
[ 'INSERT 456',
  'INSERT',
  '456',
  undefined,
  index: 0,
  input: 'INSERT 456' ]
```

Also adds test cases that include a command completion message without any quantity (ex: `COPY` by itself). Right now only a pre-8.2 server (*which hopefully nobody is running...*) would respond with a command completion message like that but this PR would handle a new command with that format being added in the future as well.


brianc:
Nice! Thank you!

PR #1423, opened 2017-08-21
Pin child modules to tighter semver ranges

brianc:
Installing any semver major child for pg-types and pg-pool is good in theory but it makes it too easy to accidentally introduce a breaking change on semver minor bumps in those modules.  While we do our best to make sure not to break backwards compat in semver minor bumps there, mistakes can happen.  This pins things to patch only which will require me to release a new minor of pg whenever a child minor bump is required, but I think that's probably better for visiblity and changelog-ability.  And safer for the community so definitely a tradeoff I'm willing to support!

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,8 +22,8 @@
     "packet-reader": "0.3.1",
     "js-string-escape": "1.0.1",
     "pg-connection-string": "0.1.3",
-    "pg-pool": "2.*",
-    "pg-types": "1.*",
+    "pg-pool": "~2.0.3",
+    "pg-types": "~1.12.1",
     "pgpass": "1.x",
     "semver": "4.3.2"
   },


Brassrat:
semver is now pinned to 4.3.2 which is now a major version behind 5.6.0.
Do you have plans for upgrading pg for the latest version of semver?

brianc:
Not really - not unless there's some security problem w/ the old one.  It's used in such a small & controlled place I haven't thought about it since I added it.

Brassrat:
Well – I was updating the dependencies on a project I am now working on (which was dormant for a while) and found that some of the modules (e.g., require-optional) require the more recent version. 
(as indicated by npm list)

I am assuming all is well with having the later semver installed – but having just updated the various dependencies haven’t done much testing yet.
Wouldn’t know exactly how to verify that all is well and was wondering why the version of semver was locked.

Not sure how to manage 3rd party modules with differing version requirements. 

**npm hell?**

charmander:
@Brassrat It’s fine for dependencies to depend on different versions of the same package in general.

Brassrat:
yes, it works but potentially adds multiple duplicate copies of the different versions of the dependencies.

brianc:
I think that's okay - bumping the version here to a newer version would have the same effect of two different copies if someone's project had a dependency or child dependency which depended on an _older_ version of semver.  One of the major features of the node module system was allowing multiple different copies of modules to be installed with different child dependencies and each child dependency can use their own version.  As long as an npm module doesn't mutate the global space (which is one of the reasons having global namespace polution in an npm module is a big no-no) this is a big advantage actually....it makes it so child modules can use their own versions of things and not care if some other child module uses some other different version as long as each child's child (yikes) version does the job the consumer of it requires.

brianc:
Your `node_modules` folder might be huge since child modules might have different versions of their own children which are close to one another but not the same...but that's a common thing w/ node.  Thankfully disk space is cheap, npm/yarn install is fast, and I've tried to keep external dependencies with node-postgres to a very small set of pretty small npm modules.  If you're working on client side bundled code having two separate transitive dependency versions _can_ cause issues (e.g. if one react component requires `lodash@2.4` and one requires `lodash@4.1` you'll have two full copies of lodash on the page) but 1) pg should never and will never be required in the browser so it's not a concern 2) most client-side module authors use more relaxed semver ranges or `peerDependencies` to get around this issue.  It can still be a pain, but that's just kinda how things go these days w/ client-side development.  Generally I approach adding npm module deps to client-side code a bit more carefully and check the webpack bundle size more than when adding npm modules to a node project.

brianc:
related:
image

Issue #442, opened 2013-09-16
Returning Invalid syntax for type json on arrays


jrf0110:
When your column is type json and you try and insert a JSON array, you get the following error:

``` javascript
{ [error: invalid input syntax for type json]
  length: 154,
  name: 'error',
  severity: 'ERROR',
  code: '22P02',
  detail: 'Expected string or "}", but found "1".',
  hint: undefined,
  position: undefined,
  internalPosition: undefined,
  internalQuery: undefined,
  where: 'JSON data, line 1: {1...',
  file: 'json.c',
  line: '665',
  routine: 'report_parse_error' }
```

For example, this works in PSQL:

``` sql
insert into my_table ( data ) values ( '[1,2,3]' );
```

But this returns an error in node-pg:

``` javascript
client.query('insert into my_table ( data ) values ( $1 )', [ [1,2,3] ], function( error ){
  /* ERROR */
});
```

eriknyk:
Hi, this issue should be reopened, since in postgres shell is possible execute:

```
INSERT INTO public.my_table ("userId", "personId", books) VALUES (1, 1, '[1,2,3]');
```

Best Regards.

jamesdixon:
I'm still running into this error even after stringifying the input array. Here's my JSON object:

```js
[
   { serviceId: 2, petId: 5, checked: true },
   { serviceId: 3, petId: null, checked: false }
]
```

After stringifying: 

```js
"[{"serviceId":2,"petId":5,"checked":true},{"serviceId":3,"petId":null,"checked":false}]"
```

I insert and receive the following error:

```
[...]

PR #1432, opened 2017-08-29
Save an array value to jsonb type column

lifubang:
```sql
CREATE TABLE public.jsontest (
	test jsonb NULL
)
WITH (
	OIDS=FALSE
);
```

```js
pool.connect(function(errp, client, done) {
  client.query('INSERT INTO jsontest(test1) VALUES($1::jsonb);', [[{"id":1, "name":"LFB"}, {"id":2, "name":"ZHM"}]], function(e,d) {
      console.log(e);
      console.log(d);
      done();
  });
});

* It returns:
{ error: invalid input syntax for type json
    at Connection.parseE (E:\develop\nodejs\mongoOPLog\node_modules\pg\lib\connection.js:546:11)
    at Connection.parseMessage (E:\develop\nodejs\mongoOPLog\node_modules\pg\lib\connection.js:371:19)
    at Socket.<anonymous> (E:\develop\nodejs\mongoOPLog\node_modules\pg\lib\connection.js:114:22)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at readableAddChunk (_stream_readable.js:176:18)
    at Socket.Readable.push (_stream_readable.js:134:10)
    at TCP.onread (net.js:547:20)
  name: 'error',
  length: 174,
  severity: 'ERROR',
  code: '22P02',
  detail: 'Expected ":", but found ",".',
  hint: undefined,
  position: undefined,
  internalPosition: undefined,
  internalQuery: undefined,
  where: 'JSON data, line 1: {"{\\"id\\":1,\\"name\\":\\"LFB\\"}",...',
  schema: undefined,
  table: undefined,
  column: undefined,
  dataType: undefined,
  constraint: undefined,
  file: 'json.c',
  line: '1151',
  routine: 'report_parse_error' }
```

You can fix it by:
```js
pool.connect(function(errp, client, done) {
  client.query('INSERT INTO jsontest(test1) VALUES($1::jsonb);', [JSON.stringify([{"id":1, "name":"LFB"}, {"id":2, "name":"ZHM"}])], function(e,d) {
      console.log(e);
      console.log(d);
      done();
  });
});
```

But if modify the arrayString function in utils.js, It will be more convenient.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -11,25 +11,30 @@ var defaults = require('./defaults');
 // convert a JS array to a postgres array literal
 // uses comma separator so won't work for types like box that use
 // a different array separator.
+// for json type column, it should be return a json string
 function arrayString(val) {
-  var result = '{';
-  for (var i = 0 ; i < val.length; i++) {
-    if(i > 0) {
-      result = result + ',';
+  if (val && val.length > 0 && typeof(val[0]) === 'object' && typeof val[0].toPostgres != 'function') {
+    return JSON.stringify(val);
+  } else {
+    var result = '{';
+    for (var i = 0 ; i < val.length; i++) {
+      if(i > 0) {
+        result = result + ',';
+      }
+      if(val[i] === null || typeof val[i] === 'undefined') {
+        result = result + 'NULL';
+      }
+      else if(Array.isArray(val[i])) {
+        result = result + arrayString(val[i]);
+      }
+      else
+      {
+        result = result + JSON.stringify(prepareValue(val[i]));
+      }
     }
-    if(val[i] === null || typeof val[i] === 'undefined') {
-      result = result + 'NULL';
-    }
-    else if(Array.isArray(val[i])) {
-      result = result + arrayString(val[i]);
-    }
-    else
-    {
-      result = result + JSON.stringify(prepareValue(val[i]));
-    }
+    result = result + '}';
+    return result;
   }
-  result = result + '}';
-  return result;
 }
 
 //converts values from javascript types


sehrope:
If I'm understanding this PR correctly, you'd checking to see if the array has non-scalar children to determine that the user's intent was to treat it as JSON (as opposed to an array):

```js
if (val && val.length > 0 && typeof(val[0]) === 'object' && typeof val[0].toPostgres != 'function') {
  return JSON.stringify(val);
} else {
  // old logic for handling arrays goes here
}
```
I don't think this is a good idea as it doesn't handle every situation. None of the following would be covered because the first element of the array is not an object:

```js
[null]
[null, 1, 2, 3]
[null, {foo: 'bar'}]
[1, {foo: 'bar'}]
[false, {foo: 'bar'}]
```

Each of those can be serialized to valid JSON but wouldn't pass the initial check in the PR. To use a value like that you'd still have to manually serialize it to JSON prior to use.

We could check all values to see if they're all non-scalars but then you'd still have divergent behavior if an empty array is specified. Should `[]` be sent as an empty Postgres array, `{}`, or an empty JSON array, `[]`?

Arrays have been a part of Postgres long before JSON, I think it's fair for them to have priority here. If someone is saving a JSON array (or an arbitrary JSON value which could be a string, number, or null), then they need to explicitly handle it.

abenhamdine:
I agree with @sehrope , I don't see any obvious way to distinguish arrays/json objects.
And you could also want to save a normal array in a jsonb column so it seems difficult to automatically apply the right format.
Personnaly, I indeed stringify specifically the object before passing it to the query. 

Otherwise, it's better to open an issue when you think you enconter a bug : not everyone watchs the pull request, so opening a bug allows users to easily find it & provide feedback if any, thx !

charmander:
As was said, this approach isn’t reliable. `JSON.stringify` is a workable workaround. See also #442.

PR #1434, opened 2017-08-29
Change Makefile to use local eslint

sehrope:
Changes the Makefile lint target to use the local copy of eslint that would be installed in node_modules rather than a global copy that may not be installed.


brianc:
ah yes excellent!

PR #1435, opened 2017-08-29
Centralize md5 hash logic

sehrope:
Two commits in this PR. The second replaces a bit of the first one but I've kept it separate as they have different purposes.

The first commit fixes references to the `client.md5(...)` helper function that was moved to `utils.js` in 2300445646db264fe03a6194e2e7a77887204027.

This had broken the connection integration tests when run with md5 auth for the test credentials. My local setup connects to a PG database inside a VM via md5 auth and I ran into this trying to run the full test suite locally.

I believe we never noticed in Travis because it's connecting to localhost without a password and that code is never executing (the no-password auth piece runs instead). I was initially worried that they weren't being run at all... but they are an all the tests pass both locally and on Travis (:thumbsup:). 

The second commit centralizes the Postgres MD5 password hashing logic so that it's in one function in `utils`. Duplication of that logic in client.js and in the tests has been updated with calls to the new function.

We should look into adding a "with a password" auth environment to CI so that these tests get run regularly as well.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -151,7 +151,7 @@ module.exports = {
     // by accident, eg: from calling values.map(utils.prepareValue)
     return prepareValue(value)
   },
-  normalizeQueryConfig: normalizeQueryConfig,
-  postgresMd5PasswordHash: postgresMd5PasswordHash,
-  md5: md5
+  normalizeQueryConfig,
+  postgresMd5PasswordHash,
+  md5
 }
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -138,6 +138,13 @@ const md5 = function (string) {
   return crypto.createHash('md5').update(string, 'utf-8').digest('hex')
 }
 
+// See AuthenticationMD5Password at https://www.postgresql.org/docs/current/static/protocol-flow.html
+const postgresMd5PasswordHash = function (user, password, salt) {
+  var inner = md5(password + user)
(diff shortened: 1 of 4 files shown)


lib/utils.js
```diff
@@ -138,12 +138,20 @@ const md5 = function (string) {
...
 module.exports = {
   prepareValue: function prepareValueWrapper (value) {
     // this ensures that extra arguments do not get passed into prepareValue
     // by accident, eg: from calling values.map(utils.prepareValue)
     return prepareValue(value)
   },
   normalizeQueryConfig: normalizeQueryConfig,
-  md5: md5
+  postgresMd5PasswordHash: postgresMd5PasswordHash,
+  md5
```
charmander: May as well use the short form for the other two properties if you’re changing this one?

sehrope:
Ah good catch. I think I had removed and added back in the md5 export at one point and that's how it ended up with the short form. I fixed up the prior commit to remove that added another to normalize exports to all use the short form. The linter doesn't seem like it when there's a trailing comma do didn't add one at the end.

brianc:
Ah nice! thanks for doing this!  And for diving in to the tests. Some of those tests were written a _long_ time ago and have accumulated some cruft.  They're valuable because they test things that used to fail, but incrementally refactoring them is +1!

sehrope:
Great. Took a bit to figure out how to get them to run but feels good knowing that I can run them all locally now. Even better to know that they pass too! :+1:

PR #1436, opened 2017-08-30
Support statement_timeout

jkennedy1980:
We need the ability to set `statement_timeout` for a given connection so that all queries timeout  instead of running to completion. This is really important when running nodejs/postgres on Heroku because they have http connection limits that timeout at 30 seconds. I want my SQL queries to stop at this boundary too since the waiting http connection will have already been closed.

I was not able to add this feature to the native side of things because postgres does not seem to support statement timeouts via pgsql connection strings.


.gitignore
```diff
@@ -5,3 +5,4 @@ build/
 node_modules/
 package-lock.json
 *.swp
+/.idea
```
charmander: You can put this in a global `gitignore` instead.

test/integration/client/statement_timeout-tests.js
```diff
@@ -0,0 +1,69 @@
...
+var helper = require('./test-helper')
+var Client = helper.Client
+
+var suite = new helper.Suite()
+
+var conInfo = helper.config
+
+function getConInfo (override) {
+  var newConInfo = {}
+  Object.keys(conInfo).forEach(function (k) {
```
charmander:
```js
return Object.assign(
  {},
  conInfo,
  override
);
```

.gitignore
jkennedy1980: I removed this. I often checkin the webstorm files in my own repos which is why I added it here. Not a problem.

test/integration/client/statement_timeout-tests.js
jkennedy1980: My bad, I started with the `appname-tests.js` file as a starting point. I've updated this function in both files.

package.json
```diff
@@ -19,9 +19,10 @@
   "main": "./lib",
   "dependencies": {
     "buffer-writer": "1.0.1",
-    "packet-reader": "0.3.1",
     "js-string-escape": "1.0.1",
+    "packet-reader": "0.3.1",
     "pg-connection-string": "0.1.3",
+    "pg-native": "^2.2.0",
```
abenhamdine: Why this addition of pg-native ?

jkennedy1980: Ah shit. The tests seem to be adding that when they are run. I'll remove it.

brianc:
Ohhhh this is nice!

test/integration/client/statement_timeout-tests.js
```diff
@@ -0,0 +1,62 @@
...
+  suite.test('statement_timeout string is used', function (done) {
+    var conf = getConInfo({
+      'statement_timeout': '3000'
+    })
+    getStatementTimeout(conf, function (res) {
+      assert.strictEqual(res, '3s')
+      done()
+    })
+  })
+}
```
abenhamdine: It looks like the tests only check for the correct configuration of statement_timeout.
I think we miss a test for checking if the timeout is actually used, by using a query with `pg_sleep(whicheverValueAboveTimeout)`

jkennedy1980: Sure. But isn't that testing Postgres itself? The tests already confirmed that the setting is set on the connection.

.gitignore
```diff
@@ -4,4 +4,4 @@ build/
 *.log
 node_modules/
 package-lock.json
-*.swp
+*.swp
```
sehrope: This looks like an unwanted edit.

test/integration/client/statement_timeout-tests.js
abenhamdine: > Sure. But isn't that testing Postgres itself? The tests already confirmed that the setting is set on the connection.

You may have a point here, but the test seems pretty simple to write, so IMHO it's better to add it.
Whatsoever @brianc has the final word.

.gitignore
jkennedy1980: I've added a newline to the end of the file to remove this change

test/integration/client/statement_timeout-tests.js
jkennedy1980: I've written it. Travis is running so I'm hoping the error response from postgres is the same from all the versions you guys support. I asserted the error code only with the hope that is will be less fragile than expecting a particular error message.

brianc:
Thanks for all the help & work on this! I'll merge it this weekend, push a new minor version, and document it! ❤️

PR #1438, opened 2017-08-30
Remove deprecated pool options from defaults

abenhamdine:
Since the upgrade of pg-pool to version 2.X, some properties seemingly have been removed from the pool options : 
- `reapIntervalMillis`, `returnToHead` and `poolLog` no more exist : I didn't find any occurrence of them in the code
- `poolIdleTimeout` has been replaced by `idleTimeoutMillis`

I also change `poolSize` to `max` : while `poolSize` is still allowed in node-pg-pool, only the option `max` is documented in the readme, see https://github.com/brianc/node-pg-pool/blob/master/README.md

In configuration test, I add more default properties assertions.

I also changed the pool url to link to node-pg-pool in the comments.

### DIFF
--- a/lib/defaults.js
+++ b/lib/defaults.js
@@ -35,23 +35,15 @@ module.exports = {
   // binary result mode
   binary: false,
 
-  // Connection pool options - see https://github.com/coopernurse/node-pool
+  // Connection pool options - see https://github.com/brianc/node-pg-pool
+
   // number of connections to use in connection pool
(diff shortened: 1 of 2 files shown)


brianc:
Awesome, thanks! I'll roll here in a bit. Gonna push a new minor version to support statement timeouts too.

abenhamdine:
Thx for merging !

PR #1441, opened 2017-09-03
Revert "Update Grammatical person"

charmander:
Reverts brianc/node-postgres#1439. Errors and awkward wording were introduced – I don’t think there’s anything wrong with a project’s creator writing its README from their perspective.


brianc:
:+1: appreciate the support. :)

PR #1444, opened 2017-09-05
Add ability to specify checkServerIdentity callback

tobio:


robertrossmann:
FYI: This implementation breaks in Node.js 10.0 because the very presence of the `checkServerIdentity` key in the configuration object, even if left `undefined`, will cause Node to fail on this assertion:

https://github.com/nodejs/node/blob/4289402aa3a7b02554aa0ebcb70af40447083fc8/lib/_tls_wrap.js#L1120-L1124

The configuration's defaults are applied via prototype chain, but the property lookup will stop immediately at the provided configuration object and will not traverse up in the prototype chain to actually find the default `tls.checkServerIdentity` function. I will send a PR with fix shortly (the fix being that this property will be explicitly set to the value of `tls.checkServerIdentity` if not provided).

PR #1446, opened 2017-09-06
Package json keywords

sehrope:
Sorts (for cleaner diffs down the road) and adds "postgresql" to the keywords section of package.json. Kind of odd that it wasn't there till now ...

The list includes an entry for "postgre" (no "s" or "sql"). Not sure if that was intentional or a typo of "postgresql" but I've left it in there for now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "pg",
     "postgre",
     "postgres",
+    "postgresql",
     "rdbms"
   ],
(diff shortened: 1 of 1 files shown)


brianc:
Nice thanks! sorry for the delay - I broke a bone & took some time off non-mandatory computer work. Back now! :dancer:

sehrope:
Cool. Hope you're feeling better!

PR #1453, opened 2017-09-14
Improve error reporting

johncf:
There is scope for improvement in the errors thrown. As of now, it is hard to distinguish postgres errors from other errors when many are mixed together. The proper way, I believe, is to give a unique `name` to the `Error`. If this is acceptable, I'll update this pull request accordingly.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -535,26 +535,14 @@ Connection.prototype._readValue = function (buffer) {
 Connection.prototype.parseE = function (buffer, length) {
   var fields = {}
   var msg, item
-  var input = new Message('error', length)
   var fieldType = this.readString(buffer, 1)
   while (fieldType !== '\0') {
     fields[fieldType] = this.parseCString(buffer)
     fieldType = this.readString(buffer, 1)
   }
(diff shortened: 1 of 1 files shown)


brianc:
You can distinguish by checking [if these fields exist](https://github.com/brianc/node-postgres/blob/master/lib/connection.js#L558) on the error.  If the exist the error is from the PostgreSQL backend.  So you could do...
```js
if (error.hasOwnProperty('detail')) {
  // error is from postgres
}
```

johncf:
@brianc That's true, and I already do that. I thought it would be cleaner to give a Pg-esque value for `name`.

Nevertheless, the PR as it is (minus the comments) is still worth a look at. The `parseE` method creates a `Message` object named `input` which I think is left over from past code, since its usage can be removed without changing program logic.

The test failed due to an unused `item` variable. I'll clean it up as soon as I'm on my laptop.

johncf:
[Updated](https://github.com/brianc/node-postgres/compare/master...johncf:error?diff=split&name=error), but changes won't be reflected here since it is closed. Passed all tests locally.

PR #1464, opened 2017-10-01
Pass through portal properly

justinj:
This happened to work before because `Query.portalName` was undefined,
but in order to be able to set the portal explicitly it should be using
`Query.portal`.

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -175,7 +175,7 @@ Query.prototype.handlePortalSuspended = function (connection) {
 
 Query.prototype._getRows = function (connection, rows) {
   connection.execute({
-    portal: this.portalName,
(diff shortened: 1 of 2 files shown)


langpavel:
Hi @justinj 
Can you describe this more widely, please?

brianc:
@justinj this is fantastic.  Sorry for the delay on this...I mentioned other places on the repo but I broke a bone early september and was __down and out__ for a long time.  I'm back now!  This looks good to merge but I have a request if you have time....could you write an _integration test_ for this as well? One that actually uses a postgres backend? Reason I ask is I am not super familiar w/ how portals are used so the test will both help me maintain the code going forward and also teach me about how portals are used in practice!

PR #1465, opened 2017-10-01
Es6fy

billouboq:
Change all objects with custom prototypes with ES6 classes and remove self variables by using arrow functions.

If you want I can change var with let-const too

I think the only breaking change is that you can't use query without new anymore : 
https://github.com/brianc/node-postgres/blob/e087305f316b364bdbd3045f5c5339b80f757174/lib/query.js#L18

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -8,7 +8,6 @@
  */
 
 var EventEmitter = require('events').EventEmitter
-var util = require('util')
 var utils = require('./utils')
 var pgPass = require('pgpass')
 var TypeOverrides = require('./type-overrides')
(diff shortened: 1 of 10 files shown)


charmander:
To avoid breaking changes, maybe it’s best to only include the arrow function commit for now?

billouboq:
Since all tests pass, don't know if it's a use case and if people are using it alone

charmander:
People are constructing `Query` objects explicitly. It may be that they’re all using `new`, but I don’t think `class` syntax is worth risking that that’s not the case. It also breaks non-ES6-class inheritance of all the pg types.

This isn’t the sort of thing tests would cover.

abenhamdine:
IMHO @charmander is totally right, it would be much safer to include such a potential breaking change in a major version.
It's difficult and risky to guess actual usage by the ecosystem, notably for a module is dowloaded more than 1 million times a month.

billouboq:
@abenhamdine @charmander what about reverting only query class to function but keep the rest with es6 class ?

charmander:
As I mentioned in the latest comment, it also stops people from being able to inherit from pg types without using `class`, so that would still be a breaking change.

abenhamdine:
@billouboq I think it would be better if the PR would be splitted in several ones : 
- use arrow functions 
- potential use const/let
- use es6 class (semver major)
It would ease the review and separate minor/major changes.
Also brianc will have probably something to say about the es6 class stuff.

billouboq:
Ok will seperate them, thanks for the review

PR #1498, opened 2017-11-14
Transaction status is handled by `readyForQuery` message. Please, share!

langpavel:
`readyForQuery` message on `Connection` can receive one of this state:

This is code from [`fe-protocol3.c`](https://github.com/postgres/postgres/blob/master/src/interfaces/libpq/fe-protocol3.c#L1499..L1526) from latest postgres source.
```c
/*
 * getReadyForQuery - process ReadyForQuery message
 */
static int
getReadyForQuery(PGconn *conn)
{
    charxact_status;

    if (pqGetc(&xact_status, conn)              )
        return EOF;
    switch (xact_status)
    {
        case 'I':
            conn->xactStatus = PQTRANS_IDLE;
            break;
        case 'T':
            conn->xactStatus = PQTRANS_INTRANS;
            break;
        case 'E':
            conn->xactStatus = PQTRANS_INERROR;
            break;
        default:
            conn->xactStatus = PQTRANS_UNKNOWN;
            break;
    }

    return 0;
}
```

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -138,7 +138,8 @@ Client.prototype.connect = function (callback) {
 
   // hook up query handling events to connection
   // after the connection initially becomes ready for queries
-  con.once('readyForQuery', function () {
+  con.once('readyForQuery', function (msg) {
+    self._transactionStatus = msg.status
     self._connecting = false
     self._connected = true
     self._attachListeners(con)
@@ -155,8 +156,9 @@ Client.prototype.connect = function (callback) {
     self.emit('connect')
   })
 
-  con.on('readyForQuery', function () {
+  con.on('readyForQuery', function (msg) {
     var activeQuery = self.activeQuery
+    self._transactionStatus = msg.status
     self.activeQuery = null
     self.readyForQuery = true
     if (activeQuery) {


langpavel:
Tests are broken because `readyForQuery` actually returns message, but test does not mock it :-(
 - Fixing tests will be heavy work
 - Hacking `Client` to pass tests is, well, :nauseated_face:

langpavel:
I believe that with this you can decide if connection *cannot* be safely returned to pool
or if closing it can cause potential data loss.
I know that there is more work to do, but please tell me @brianc if I'm on good way and more work on this PR will be welcomed :-)
Thanks

brianc:
@langpavel you're totally on the right track! Sorry for the delay in getting back to you.  If you get stuck or have questions just lemme know! I can help out as much or as little as you'd like.

langpavel:
Yeah, unfortunately I'm quite busy and I have different things, but I can try work on this at, well, free moments… :slightly_smiling_face:
Thanks Brian and please others, ping me in this PR if someone else is interested at this, it can motivate me :slightly_smiling_face:

langpavel:
Todo list is:
 - [ ] Fix tests at first
 - [ ] Make some more tests with query clashing at the place..
 - [ ] especially that active uncommited  transaction in valid state cannot be returned to the pool — this is ***leak***
 - [ ] invalid transaction will not be returned to the pool

langpavel:
Outdated

PR #1502, opened 2017-11-15
Make tests compatible with PostgreSQL 10

charmander:
PostgreSQL 10 reports its version as only `major.minor`, so it can’t be parsed with semver. The `server_version_num` setting is a major version followed by a four-digit minor version since version 10, and was a three-digit major version followed by a two-digit minor version before that.

It might be best not to merge this until CI supports PostgreSQL 10.


brianc:
You are awesome!

Issue #632, opened 2014-08-20
The pool destroys the client with queued queries to it


grudzinski:
The pool destroys the client with queued queries to it. I think the pool should reassign the queries to another client when it decided to destroy the client.

charmander:
Fixed in v7.5.0.

Issue #1500, opened 2017-11-14
Client#query promise stuck when connection closed


rkaw92:
Hi, we've just encountered a problem which seems to predictably occur when a Client instance disconnects. It results in promises returned from `Client#query` never settling as fulfilled nor rejected (when using callbacks, those are not fired, either).

The below gist is a trivial example (with output) to reproduce the behavior. If you cut off the connection to PostgreSQL at some point (in my case, it's a Docker container with `postgres:9.6` being stopped), the next `.query()` call rejects with an error, **but the next call after that never settles its promise**, as illustrated by the output.

https://gist.github.com/rkaw92/9cbd71fdb98d74f01c624bb2a003cd2a

You'll notice that subsequent queries are never executed, and they never reject, either - they just hang.

The issue has several important consequences:
* It breaks libraries and applications that rely on the Promise always being resolved or rejected, such as knex.js, causing eventual client pool exhaustion
  * Indeed, if you do suffer from a disconnect with knex,js and your connection pool is exhausted (because the .query() hangs the second time, never releasing the client), then restoring connectivity has no effect - the pool is forever exhausted. Note, however, that knex.js uses `generic-pool` rather than pg's default pool, and manages acquiring/releasing clients on its own. (*This is another issue that I am currently pursuing - knex's connection recycling code is buggy at the moment.*)
* It leaves unresolved Promises, which are never a good thing and may trigger Node monitoring tools (and rightly so).
* It most likely keeps adding queries to the internal Client queue indefinitely, leaking memory

I think I've pinned down this behavior to the following sequence of events (hold on, it's going to be a bumpy ride over [client.js](https://github.com/brianc/node-postgres/blob/master/lib/client.js)):
* Assume the Client instance is connected and the connection has signalled readiness - the handler in [line 141](https://github.com/brianc/node-postgres/blob/master/lib/client.js#L141) has fired. Any connection error events will now be handled using `connectedErrorHandler`.
* The connection goes down, thus emitting `error` (and `end` shortly after) on the `con` object.
* `connectedErrorHandler` fires. In our example, we will assume that no query is currently active (if there is one, the situation converges to this case later, anyway), so the handler has no effect other than re-emitting the `error` event on the Client object.
* The user sends a query using `Client#query`. The *Query* object is [pushed to the internal query queue](https://github.com/brianc/node-postgres/blob/master/lib/client.js#L392) and we enter `_pulseQueryQueue`.
[...]

Issue #1105, opened 2016-08-09
TypeError: Cannot read property 'name' of null


greghart:
First off, thanks for the library! Overall it's been incredibly performant and straightforward to use. I am running into one issue which may be an issue with the library, though it's a bit hard to follow the protocol for me so my usage may be off as well.

I'm running into an uncaught error during the `parseComplete` handler in client.js (https://github.com/brianc/node-postgres/blob/v6.0.1/lib/client.js#L131).

I am doing a direct `connect` against the pool and then utilizing pg-query-stream

``` javascript
pool.connect().then(function(client){
  stream = client.query(new QueryStream(query, null));

  stream.on('error', function(streamError){
    stream.close(function(closeError){ 
      // What to do with a close error? At this point stream is already hosed :shrug:
      client.release(streamError);
    }
  });

  stream.on('close', function(){
    client.release();
  }); 
});
```

At some point during my app's life cycle, I get an uncaught error

```
TypeError: Cannot read property 'name' of null
    at null.<anonymous> (/var/app/current/node_modules/pg/lib/client.js:139:26)
    at emit (events.js:107:17)
    at Socket.<anonymous> (/var/app/current/node_modules/pg/lib/connection.js:121:12)
    at Socket.emit (events.js:107:17)
    at readableAddChunk (_stream_readable.js:163:16)
    at Socket.Readable.push (_stream_readable.js:126:10)
    at TCP.onread (net.js:540:20)
```

So obviously `activeQuery` is becoming null, and I am trying to narrow down why I would receive a `parseComplete` event when the active query is null.

According to my logs, the stream is _not_ emitting an error, and seems to be closing normally. Therefore it seems like the connection client is getting a `readyForQuery` or `end` event, and then right after that getting the `parseComplete` event. Any idea why this would be happening, or see any issues with my usage? Thanks for any help you can give! I'll keep looking into it as well.

vitaly-t:
I wonder, if the same error would occur within [pg-promise](https://github.com/vitaly-t/pg-promise), if you follow [the example](https://github.com/vitaly-t/pg-promise/wiki/Learn-by-Example#from-database)...

At least this would help ruling out some reasons of why this is happening...

brianc:
@greghart interesting error - any way you could include the query & some data so I could try to reproduce it here?  Maybe using [`generate_series`](https://www.postgresql.org/docs/9.1/static/functions-srf.html) or something if you need lots of data?

dynajoe:
[...]

PR #1503, opened 2017-11-15
Queued query errors

### LINKED ISSUE #1454: Fails to error on queries inside a transaction after a lost connection
When we execute a lengthy operation inside a transaction, and the connection is suddenly lost during that time, then any query we execute on the `client` after that become stuck, i.e. they report nether success no error:

```js
'use strict';

const Pool = require('pg-pool');

const cn = {
    database: 'newone',
    port: 5433,
    user: 'postgres'
};

const p = new Pool(cn);

p.on('error', error => {
    // do nothing
});

p.connect((err, client, release) => {
    if (err) {
        console.log('Failed to connect:', err);
        return;
    }
    const test = async function () {
        try {
            await client.query('BEGIN');
            await client.query('SELECT pg_sleep(10)'); // connection breaks during this one
            await client.query('COMMIT');
            console.log('success!');
        } catch (e) {
            console.log('about to rollback...'); // this is reported
            await client.query('ROLLBACK'); // this one is stuck, never reporting an error, if the error was due to lost connection
            console.log('rollback finished'); // this is never reported, if we are here because of lost connection
        } finally {
            release();
            p.end();
        }
    };

    test();
});
```

I've tried both `pg-pool` directly and via `node-postgres`, versions 6.x and 7.x - all the same.

**Expected Behavior**

Executing any query against `client` at that point should immediately report an error that informs us of the lost connectivity.

charmander:
> Separates socket errors from error messages, sends socket errors to all queries in the queue, marks clients as unusable after socket errors.
>
> This is not very pleasant but should maintain backwards compatibility…?

I don’t know if this is how you want to handle the issue, but it should address #1454 and #632.

Maybe in a future version the query queue and clients pending `connect()` won’t need to exist at all?

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -36,6 +36,7 @@ var Client = function (config) {
   this._connecting = false
   this._connected = false
   this._connectionError = false
+  this._queryable = true
 
   this.connection = c.connection || new Connection({
     stream: c.stream,
@@ -126,15 +127,39 @@ Client.prototype.connect = function (callback) {
   }
 
   const connectedErrorHandler = (err) => {
-    if (this.activeQuery) {
-      var activeQuery = self.activeQuery
-      this.activeQuery = null
-      return activeQuery.handleError(err, con)
+    this._queryable = false
+
+    const enqueueError = (query) => {
+      process.nextTick(() => {
+        query.handleError(err, con)
+      })
     }
+
+    if (this.activeQuery) {
+      enqueueError(this.activeQuery)
+      this.activeQuery = null
+    }
+
+    this.queryQueue.forEach(enqueueError)
+    this.queryQueue = []
+
     this.emit('error', err)
   }
 
+  const connectedErrorMessageHandler = (msg) => {
+    const activeQuery = this.activeQuery
+
+    if (!activeQuery) {
+      connectedErrorHandler(msg)
+      return
+    }
+
+    this.activeQuery = null
+    activeQuery.handleError(msg, con)
+  }
+
   con.on('error', connectingErrorHandler)
+  con.on('errorMessage', connectingErrorHandler)
 
   // hook up query handling events to connection
   // after the connection initially becomes ready for queries
@@ -143,7 +168,9 @@ Client.prototype.connect = function (callback) {
     self._connected = true
     self._attachListeners(con)
     con.removeListener('error', connectingErrorHandler)
+    con.removeListener('errorMessage', connectingErrorHandler)
     con.on('error', connectedErrorHandler)
+    con.on('errorMessage', connectedErrorMessageHandler)
 
     // process possible callback argument to Client#connect
     if (callback) {
@@ -353,7 +380,13 @@ Client.prototype._pulseQueryQueue = function () {
     if (this.activeQuery) {
       this.readyForQuery = false
       this.hasExecuted = true
-      this.activeQuery.submit(this.connection)
+
+      const queryError = this.activeQuery.submit(this.connection)
+      if (queryError) {
+        this.activeQuery.handleError(queryError, this.connection)
+        this.readyForQuery = true
+        this._pulseQueryQueue()
+      }
     } else if (this.hasExecuted) {
       this.activeQuery = null
       this.emit('drain')
@@ -389,6 +422,16 @@ Client.prototype.query = function (config, values, callback) {
     query._result._getTypeParser = this._types.getTypeParser.bind(this._types)
   }
 
+  if (!this._queryable) {
+    query.handleError(new Error('Client has encountered a connection error and is not queryable'), this.connection)
+    return
+  }
+
+  if (this._ending) {
+    query.handleError(new Error('Client was closed and is not queryable'), this.connection)
+    return
+  }
+
   this.queryQueue.push(query)
   this._pulseQueryQueue()
   return result
@@ -396,18 +439,19 @@ Client.prototype.query = function (config, values, callback) {
 
 Client.prototype.end = function (cb) {
   this._ending = true
+
   if (this.activeQuery) {
     // if we have an active query we need to force a disconnect
     // on the socket - otherwise a hung query could block end forever
-    this.connection.stream.destroy(new Error('Connection terminated by user'))
-    return cb ? cb() : Promise.resolve()
-  }
-  if (cb) {
+    this.connection.stream.destroy()
+  } else {
     this.connection.end()
+  }
+
+  if (cb) {
     this.connection.once('end', cb)
   } else {
-    return new global.Promise((resolve, reject) => {
-      this.connection.end()
+    return new Promise((resolve) => {
       this.connection.once('end', resolve)
     })
   }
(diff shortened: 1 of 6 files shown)


charmander:
(Tests are failing because this is missing the changes for lib/native. Does it still offer a significant performance advantage, by the way?)

rkaw92:
Hi, just read your test case. Can you please test with two or more queries after the connection is downed? As reported in #1500 , the first query after the connection fails was already emitting an error. It was the *second* query that was locked forever.

https://gist.github.com/rkaw92/9cbd71fdb98d74f01c624bb2a003cd2a#file-client-query-disconnect-js

lib/query.js
```diff
@@ -147,22 +147,17 @@ Query.prototype.handleError = function (err, connection) {
 
 Query.prototype.submit = function (connection) {
   if (typeof this.text !== 'string' && typeof this.name !== 'string') {
-    const err = new Error('A query must have either text or a name. Supplying neither is unsupported.')
-    connection.emit('error', err)
-    connection.emit('readyForQuery')
-    return
+    return new Error('A query must have either text or a name. Supplying neither is unsupported.')
```
rkaw92: Returning an error doesn't seem like a widely-used practice. Why are we not throwing it?

charmander: It would have to be caught to be passed to the callback and it wouldn’t be easy to distinguish from other errors at that point.

charmander:
@rkaw92 In your case, the connection was terminated between queries. In the test, the connection is terminated during the first query, and there is no subsequent `readyForQuery` message.

lib/query.js
rkaw92: Now that you mention it, do we need to distinguish it from other errors? Reporting an error that occurred in the stack when calling Query.submit as a "query error" sounds rather intuitive.

charmander: There shouldn’t be any other errors so I prefer not to treat those like they’re expected if they do occur.

lib/client.js
```diff
@@ -389,25 +422,36 @@ Client.prototype.query = function (config, values, callback) {
     query._result._getTypeParser = this._types.getTypeParser.bind(this._types)
   }
 
+  if (!this._queryable) {
+    query.handleError(new Error('Client has encountered a connection error and is not queryable'), this.connection)
+    return
```
rkaw92: Returning `undefined` causes breakage for Promise users when calling `.query()` after the connection has gone down. Should return `result` instead.

charmander: Thanks!

rkaw92:
Something is not quite right, still. Tested with my original gist:
* The first query after the failing connection does reject with an error - however:
* The second query after the failing connection encounters `!this._queryable` and this causes an invalid return (which you've now fixed just a moment ago).

It appears I was wrong in my initial assessment of the sequence of events in case the connection goes down (ends) between queries (`activeQuery === null`) - at the time the connection ends (assuming it goes down gracefully, i.e. TCP RST, no timeout), there is no immediate `error` event, just an `end`. Here's what happens with the code from your branch:
* Connection ends, emitting an `end` event
* The `con.once('end', ...)` handler fires
  * `error` event is emitted on the Client object
* Note that there was no `error` event on the actual *connection* object, nor an `errorMessage` (because the connection just went away without sending us anything)
  * We still have not entered `connectedErrorMessageHandler` nor `connectedErrorHandler`
* User fires **query1** - this does not trip the `!this.queryable` check so we go forward and post the query to the connection stream
* Now we get a real `error` from the stream (because we tried to `write()` to it), so we run `connectedErrorHandler` and it sets `this.queryable` to false. Since a query is active, it gets a legit promise rejection.
* User fires **query2** - this time, the `!this.queryable` check causes an early exit.

After your `return` fix, it seems like the whole thing should work, but it may be a bit surprising to users that the first query after the connection goes down behaves differently than the second one (i.e. different error). What do you think?

charmander:
I think a user should be checking error codes if they expect certain types of errors and not sending a second query after the connection failed. The `end` listener for the `Connection` not setting `_queryable = false` is an oversight, though.

rkaw92:
Nice. All queries are now receiving the same error in the scenario from my gist.

charmander:
@brianc, is the right way to implement this for pg-native to change https://github.com/brianc/node-pg-native/blob/25d12eb2a26d10a8c32036acf19253e131c266f5/index.js#L208 to a query error and all other errors to fatal errors?

brianc:
@charmander - yah I think so.  I know it's kinda a pain to have to go change something over in `pg-native` and the over here as well. :( Maybe one day we can combine things into a monorepo and use yarn workspaces or something to manage it.  Thanks for doing this by the way - this is awesome. :hugs:

rkaw92:
Guys, where are we on this? Any roadblocks?

brianc:
@charmander I updated pg-native w/ the breaking change to emit errors more regularly [here](https://github.com/brianc/node-pg-native/pull/64).  Looks like the tests are still failing here for some reason.

charmander:
@brianc Tests fixed. Thanks for all the work today!

brianc:
Ah nice! I'll merge this tonight, put together a final release for the 7.x branch, and the major version bump for the small change to not being able to call `pg.Pool()` without the new keyword from #1541

aelath:
👋Wanted to check in on progress of this PR. I've been following an issue around (#1105) and I think this PR *might* fix our issues we are seeing in production. When do you expect to merge this? Thanks!

charmander:
@brianc ^?

vitaly-t:
@brianc What happened here? Why wasn't it merged?

@charmander

vitaly-t:
Something makes my scratch my head in these changes....

It removed the `callback` verification inside `Client.prototype._connect`, presuming that the value is always set, since the method is always called from `Client.prototype.connect`.

However, the internal method sets `callback` to `null`, just as it receives `readyForQuery` notification: https://github.com/brianc/node-postgres/blob/3828aa86081491282c118884c245b4a19619f02e/lib/client.js#L181

And since the callback was replaced to use `nextTick`, it is now entirely possible that when we get inside that tick, event `readyForQuery` has already fired, and then calling unverified `callback()` inside the tick will throw an error, pending a race condition.

brianc:
yeah that might be possible - could you write a failing test for this? I'll fix it in a PR if you can reproduce.

vitaly-t:
@brianc That's the known specific of race conditions, they are almost impossible to test in a controlled environment. I don't think I can create one, sorry.

brianc:
I thought about this a bit more - I don't think its possible.  Each call to `client.connect` creates a new function scope.  If this 'error you have already used this client' error fires it will never reach the point in the code where `readyForQuery` is fired, and vise versa.  It's also not reproducible even through manually firing `readyForQuery` or manipulating the event listener collection on `client.connection`.

brianc:
and checking for `if(callback)` is no longer actually needed in the `_connect` function because if the `client.connect` function is called without a callback it now returns a promise, which rejects, and either way `_connect` is _always_ called with a callback function (either user supplied or the callbackified promise handler)

vitaly-t:
Ok, I trust you this, but time will tell, if there is any issue there :wink:

Ans in the second post you referred to what I was describing myself, but the issue I mentioned was that internally `callback` is set to `null`, and since `nextTick` is asynchronous, a race condition seemed possible.

PR #1507, opened 2017-11-16
Specify collation when relied on in tests

### LINKED ISSUE #189: Tests fail when running on DB's with certain locales
http://www.postgresql.org/docs/9.1/static/locale.html notes that databases uses locale specific sorting. Because the tests do not create their own database, they are dependent on the locale of the database. This breaks the sorting test.

Using xxx_pattern_ops (http://www.postgresql.org/docs/9.1/interactive/indexes-opclass.html) the test could enforce English sorting.

```
simple-query-tests.js.....Message: undefined
AssertionError: "Brian" == "Aaron"
at /var/www/node-postgres/test/integration/client/simple-query-tests.js:33:14
at test (/var/www/node-postgres/test/test-helper.js:145:16)
at null.<anonymous> (/var/www/node-postgres/test/integration/client/simple-query-tests.js:32:5)
at assert.emits (/var/www/node-postgres/test/test-helper.js:46:16)
at g (events.js:185:14)
at EventEmitter.emit (events.js:88:17)
at p.handleReadyForQuery (/var/www/node-postgres/lib/query.js:81:8)
at p.connect.con.on.self.activeQuery (/var/www/node-postgres/lib/client.js:121:24)
at EventEmitter.emit (events.js:115:20)
at Socket.p.connect (/var/www/node-postgres/lib/connection.js:75:14)
```

charmander:
Fixes #189.

### DIFF
--- a/test/integration/client/api-tests.js
+++ b/test/integration/client/api-tests.js
@@ -26,7 +26,7 @@ suite.test('callback API', done => {
   client.query(config)
   client.query('INSERT INTO peep(name) VALUES ($1)', ['aaron'])
 
-  client.query('SELECT * FROM peep ORDER BY name', (err, res) => {
+  client.query('SELECT * FROM peep ORDER BY name COLLATE "C"', (err, res) => {
     assert(!err)
     assert.equal(res.rowCount, 3)
     assert.deepEqual(res.rows, [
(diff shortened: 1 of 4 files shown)


brianc:
Whoah nice. Thank you!

PR #1518, opened 2017-11-21
Allow a custom type to be used for Client promises

charmander:
Matches the Pool API.


brianc:
Lovely! Thank you as always @charmander  :heart:

PR #1537, opened 2017-12-13
Ignore socket hangup when ending connection also with ssl (#1344).

pasieronen:
See issue #1344 for details.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -97,11 +97,17 @@ Connection.prototype.connect = function (port, host) {
       NPNProtocols: self.ssl.NPNProtocols
     })
     self.attachListeners(self.stream)
-    self.emit('sslconnect')
 
     self.stream.on('error', function (error) {
+      // don't raise ECONNRESET errors - they can & should be ignored
+      // during disconnect
+      if (self._ending && error.code === 'ECONNRESET') {
+        return
+      }
       self.emit('error', error)
     })
+
+    self.emit('sslconnect')
   })
 }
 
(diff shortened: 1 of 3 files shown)


test/unit/connection/error-tests.js
```diff
@@ -1,31 +1,62 @@
...
+})
+
+
+suite.test('connection does not emit ECONNRESET errors during disconnect also when using SSL', function (done) {
+  // our fake postgres server, which just responds with 'S' to start SSL
+  var socket
+  var server = net.createServer(function (c) {
+    socket = c
+    c.once('data', function (data) {
+      c.write(new Buffer('S', 'utf8'))
```
charmander: You can use `Buffer.from`; the versions of Node pg 7 supports all support it.

lib/connection.js
```diff
@@ -97,11 +97,17 @@ Connection.prototype.connect = function (port, host) {
       NPNProtocols: self.ssl.NPNProtocols
     })
     self.attachListeners(self.stream)
-    self.emit('sslconnect')
 
     self.stream.on('error', function (error) {
```
charmander: Can this reuse the function from earlier instead of duplicating it?

ubreddy:
any idea when this will be merged and new version release?

brianc:
This is fantastic. Sorry for the delay on merging this. I'll push a new minor version out right now w/ the fix.

brianc:
Published `pg@7.4.1`

PR #1541, opened 2017-12-15
Use class-extends to wrap Pool

NatalieWolfe:
Using the es5-style pool wrapping from before makes it difficult to subclass `BoundPool`. Since Node 4.5 is the lowest supported version, it is possible to use es6-style classes.

The subsclassing issue stems from the fact that `BoundPool` was not applying `Pool` to `this`, instead returning a new Pool instance. It couldn't safely apply Pool since Pool is an es6 class and you must call es6 class constructors with `new`. By changing `BoundPool` to also be an es6 class we can call `super` and thus "apply" the Pool constructor to `this`.


brianc:
@NatalieWolfe thanks for your work on this...and I apologize for the delay. I'm going to bundle this up with [https://github.com/brianc/node-postgres/pull/1503] because I think that also requires semver major, right @charmander?

charmander:
@brianc It depends on a pull request that should be a major version of pg-native, but I don’t think it needs to be a major version of pg. (The change to pg-native causes it client to emit `error` events in more situations, but the way it affects the pg wrapper can be considered a bugfix.)

brianc:
@charmander I think you might have responded to the wrong pull request here w/ your comment? Because this removes the ability to do `const pool = pg.Pool()` (now requiring new) it's a subtle breaking change which unfortunately requres a semver major. Once the minor patches settle down a bit here I'll do the semver major bump for this.

sehrope:
There's a bunch of stuff in node core that detects if it's not called as a constructor and if so silently returns the constructor invocation instead:

Here's an example from the [zlib](https://github.com/nodejs/node/blob/2d609c5d53a4394bd2a9c6e15ab074c3cc2039a2/lib/zlib.js#L657) module: 

```
function Gzip(opts) {
  if (!(this instanceof Gzip))
    return new Gzip(opts);
  Zlib.call(this, opts, GZIP);
}
```

Not sure if the rest of this PR would end up being semver major anyway but this approach might be something to consider to do this class changeover in a less intrusive way.

sehrope:
Ah good point. Nevermind!

charmander:
@brianc Sorry for the confusing wording; it =

> I'm going to bundle this up with [https://github.com/brianc/node-postgres/pull/1503]

#1503 is a bugfix.

Issue #874, opened 2015-10-09
Time queries?


kevinburkeshyp:
It would be nice to see timing information on how long `node-postgres` spent acquiring a connection from the pool, and how long it spent querying the database, and how long Postgres thinks it spent on the query. (I'm not sure whether Postgres returns that information)

paulovieira:
This related project might be of interest: https://github.com/vitaly-t/pg-monitor

fpietka:
It is a real drawback for production performances monitoring

knutole:
Especially a simple `\timing` option would be great.

jimmiehansson:
We're building a thin layer ontop of this of module and this is essential to our production environment, any schedule on this?

abenhamdine:
>  and how long it spent querying the database, and how long Postgres thinks it spent on the query. 
pg logs are made for this, not this library.
see for example libraries like https://github.com/dalibo/pgbadger

brianc:
@jimmiehansson if you have a need for an essential production feature I'm happy to prioritize it if you'd like to enlist my consulting services! Please send me an email at brian.m.carlson@gmail.com if you want to discuss further.  Otherwise: pull requests will always be reviewed and accepted as soon as I get time! 😄

jimmiehansson:
Thanks guys! We'll have a look at the alternatives :)

PR #1563, opened 2018-01-24
Emit an event after submitting the query.

emilsedgh:
Makes it possible to profile queries.
Fixes brianc/node-postgres#874

If this event is emitted, then anyone who wants can profile their queries by comparing the time of this and the first row returned. like this:

``` javascript
let submit_time, response_time

query.on('submit', () => {
   submit_time = process.hrtime()
})

query.once('row', () => {
  response_time = process.hrtime(submit_time)
  const [seconds, nanoseconds] = process.hrtime(submit_time)
  const elapsed = Math.round((seconds * 1000) + (nanoseconds * 1e-6))
  console.log(`Elapsed: ${elapsed}`)
})

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -158,11 +158,13 @@ Query.prototype.submit = function (connection) {
     connection.emit('readyForQuery')
     return
   }
+
   if (this.requiresPreparation()) {
     this.prepare(connection)
   } else {
     connection.query(this.text)
   }
+  this.emit('submit')
 }
 
 Query.prototype.hasBeenParsed = function (connection) {


emilsedgh:
This isn't useful for the mentioned purpose as the `row` event isn't fired if there are no rows.

I'll revisit.

dt-rush:
@emilsedgh is there any reason why we couldn't listen for `'end'`?

neongreen:
@emilsedgh have you found a way to get query durations after all? I've stumbled upon the same issue.

PR #1570, opened 2018-02-09
Allow custom Promise dependency to be used on clients instead of the global Promise

pedrocarrico:
Hi,

I was using connection pooling and passed in a bluebird Promise configuration and noticed that whenever I got a new Client from the pool it wouldn't get a bluebird Promise.

With these changes clients should pickup the Promise configuration from their pool as well as the option to specify any Promise dependency one might want.

Thanks,


charmander:
Covered with tests by #1518. I’ll reopen it after the bugfix it conflicts with (#1503) is in.

PR #1572, opened 2018-02-12
Handle SSL negotiation errors more robustly

mble:
This commit adds some finer grained detail to handling the postmaster's
response to SSL negotiation packets, by accounting for the possibility
of an 'E' byte being sent back, and emitting an appropriate error.

In the naive case, the postmaster will respond with either 'S' (proceed
with an SSL connection) or 'N' (SSL is not supported). However, the
current if statement doesn't account for an 'E' byte being returned
by the postmaster, where an error is encountered (perhaps unable to
fork due to being out of memory).

By adding this case, we can prevent confusing error messages when SSL is
enforced and the postmaster returns an error after successful SSL
connections.

This also brings the connection handling further in line with
`libpq`, where 'E' is handled similarly as of this commit:

https://github.com/postgres/postgres/commit/a49fbaaf8d461ff91912c30b3563d54649474c80

('E' is handled when the status is `CONNECTION_AWAITING_RESPONSE`)

Given that there are no longer pre-7.0 databases out in the wild, I
believe this is a safe change to make, and should not break backwards
compatibility (unless matching on error message content(!)).

* Replace if statement with switch, to catch 'S', 'E' and 'N' bytes
  returned by the postmaster
* Return an Error for non 'S' or 'N' cases
* Expand and restructure unit tests for SSL negotiation packets

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -82,8 +82,13 @@ Connection.prototype.connect = function (port, host) {
 
   this.stream.once('data', function (buffer) {
     var responseCode = buffer.toString('utf8')
-    if (responseCode !== 'S') {
-      return self.emit('error', new Error('The server does not support SSL connections'))
+    switch (responseCode) {
+      case 'N':
+        return self.emit('error', new Error('The server does not support SSL connections'))
+      case 'S':
+        break
+      default:
+        return self.emit('error', new Error('There was an error establishing an SSL connection'))
     }
     var tls = require('tls')
     self.stream = tls.connect({
--- a/test/unit/connection/error-tests.js
+++ b/test/unit/connection/error-tests.js
@@ -37,26 +37,52 @@ suite.test('connection does not emit ECONNRESET errors during disconnect', funct
   done()
 })
 
+var SSLNegotiationPacketTests = [
+  {
+    testName: 'connection does not emit ECONNRESET errors during disconnect also when using SSL',
+    errorMessage: null,
+    response: 'S',
+    responseType: 'sslconnect'
+  },
+  {
+    testName: 'connection emits an error when SSL is not supported',
+    errorMessage: 'The server does not support SSL connections',
+    response: 'N',
+    responseType: 'error'
+  },
+  {
+    testName: 'connection emits an error when postmaster responds to SSL negotiation packet',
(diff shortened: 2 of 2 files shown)


brianc:
ah really nice.  This looks good to me...and +1 for the tests.  @charmander what do you think?

test/unit/connection/error-tests.js
```diff
@@ -37,26 +37,52 @@ suite.test('connection does not emit ECONNRESET errors during disconnect', funct
...
-    assert.emits(con, 'sslconnect', function () {
-      con.end()
-      socket.destroy()
-      server.close()
-      done()
+    server.listen(7778, function () {
+      var con = new Connection({ssl: true})
+      con.connect(7778, 'localhost')
+      assert.emits(con, tc.responseType, function (err) {
+        if (err) {
```
charmander: Should this run `if (tc.errorMessage !== null)`?

mble: If there is no `errorMessage`, then no. Good catch.

charmander: Ah, I meant that this won’t check for an expected error if an error isn’t emitted.

mble: Ah, yeah, I'm with you. That's some short sightedness on my part – I was checking for the other case (when an error is emitted and one isn't expected). I've now modified this test to run if either an `errorMessage` is specified (for expected errors), or if `err` is present (for unexpected errors).

lib/connection.js
```diff
@@ -82,8 +82,13 @@ Connection.prototype.connect = function (port, host) {
...
   this.stream.once('data', function (buffer) {
     var responseCode = buffer.toString('utf8')
-    if (responseCode !== 'S') {
-      return self.emit('error', new Error('The server does not support SSL connections'))
+    switch (responseCode) {
+      case 'N':
+        return self.emit('error', new Error('The server does not support SSL connections'))
+      case 'S':
+        break
+      default:
```
abenhamdine: perhaps adding a short comment for every case would be useful, these codes are not well known for many users.

mble: Agreed, done.

abenhamdine: Thx !

Issue #1574, opened 2018-02-23
Package main should point to index


arichiardi:
Hello folks!
I have recently had a problem with importing this package and realized that the `main` field in `paxhage.json` does not point to `lib/index.js`.

Would you be interested in a small patch?

While it is true that it seems to work fine in a standard node environment, the `package.json` doc requires a module there and other less lenient tools are failing.

What do you think?

charmander:
`./lib` *is* a module by virtue of containing an `index.js`; you should report a bug with the failing tools.

PR #1575, opened 2018-02-23
Update the package main to point to index.js

arichiardi:
The specs demands it and some tooling that are not as forgiving as npm complain
about it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "url": "git://github.com/brianc/node-postgres.git"
   },
   "author": "Brian Carlson <brian.m.carlson@gmail.com>",
-  "main": "./lib",
+  "main": "./lib/index.js",
   "dependencies": {
     "buffer-writer": "1.0.1",
     "js-string-escape": "1.0.1",


arichiardi:
I confirm this solves my issue.

arichiardi:
See #1574.

charmander:
@arichiardi … although this not being a pg bug doesn’t mean it can’t be merged. =) Which tools are they?

arichiardi:
@charmander Oh, your issue seemed pretty clear :smile: 

I am using `node-postgres` from ClojureScript and apparently there is something there that does not allow correctly loading of the `main` in its current form. This fixes it, but of course it is a workaround, at some point the real reason will need to be investigated. I have a fork on my side that does its job for now.

arichiardi:
So ClojureScript uses the the Google Closure Compiler which seems to be CommonJS focused, it cannot understand that main. This is my limited understanding of it. This is a small and probably simple change that would make it compatible so maybe worth considering?

I haven't tried without the ClojureScript harness so this is pure fiction for now.

arichiardi:
This was fixed in ClojureScript: https://dev.clojure.org/jira/browse/CLJS-2580

PR #1592, opened 2018-03-14
use net.Socket instead of net.Stream

chentsulin:
`net.Stream` is a undocumented legacy naming from node 0.x
https://github.com/nodejs/node/blob/4ae320f2b3c745402955019d6a57a22ee2b8d3bd/lib/net.js#L1762


charmander:
Thank you!

PR #1603, opened 2018-04-02
dont use dynamic functions to parse rows, closes #1417

### LINKED ISSUE #1417: Potential memory leak
Hi,

Im' investigating a memory leak in one of my app. It seems that this module is generating thousands of strings containing:

```js
(function(parsers,rowData
/**/) {

this['p1'] = rowData[0] == null ? null : parsers[0](rowData[0]);
this['p2'] = rowData[1] == null ? null : parsers[1](rowData[1]);
})
```

(displayed here as a function, but it is actually a String).

Those seems to be generated in `pg/lib/result.js::inlineParser`, related to https://github.com/brianc/node-postgres/issues/507:

```js
var inlineParser = function(fieldName, i) {
  throw new Error("PG bug");
  return "\nthis['" +
    //fields containing single quotes will break
    //the evaluated javascript unless they are escaped
    //see https://github.com/brianc/node-postgres/issues/507
    //Addendum: However, we need to make sure to replace all
    //occurences of apostrophes, not just the first one.
    //See https://github.com/brianc/node-postgres/issues/934
    fieldName.replace(/'/g, "\\'") +
    "'] = " +
    "rowData[" + i + "] == null ? null : parsers[" + i + "](rowData[" + i + "]);";
};
```

Here is an example result from memory analysis using dev tools:
image

(screenshot: only displaying items created after establishing a memory baseline, and after a forced Garbage collection. I'm only displaying one screen here, but I can scroll for ages on the exact same string).

This leak isn't super fast, it takes roughly a day to OOM on a relatively busy server (70k rpm, around 100 requests to PG/s).

I could also be using the library in a bad way, but the fact that all those strings remains and not my queries is weird.

Thanks,

contra:
Test pass, next step is benchmarks - I could use a hand with those if you have time, but I think we should be able to optimize this further to beat the dynamic function performance.

### DIFF
--- a/lib/result.js
+++ b/lib/result.js
@@ -8,7 +8,6 @@
  */
 
 var types = require('pg-types')
-var escape = require('js-string-escape')
 
 // result object returned from query
 // in the 'end' event and also
@@ -65,27 +64,22 @@ Result.prototype._parseRowAsArray = function (rowData) {
   return row
 }
 
-// rowData is an array of text or binary values
-// this turns the row into a JavaScript object
 Result.prototype.parseRow = function (rowData) {
-  return new this.RowCtor(this._parsers, rowData)
+  var row = {}
+  for (var i = 0, len = rowData.length; i < len; i++) {
+    var rawValue = rowData[i]
+    var field = this.fields[i].name
+    if (rawValue !== null) {
+      row[field] = this._parsers[i](rawValue)
+    } else {
+      row[field] = null
+    }
+  }
+  return row
 }
 
 Result.prototype.addRow = function (row) {
   this.rows.push(row)
-}
-
-var inlineParser = function (fieldName, i) {
-  return "\nthis['" +
-    // fields containing single quotes will break
-    // the evaluated javascript unless they are escaped
-    // see https://github.com/brianc/node-postgres/issues/507
-    // Addendum: However, we need to make sure to replace all
-    // occurences of apostrophes, not just the first one.
-    // See https://github.com/brianc/node-postgres/issues/934
-    escape(fieldName) +
-    "'] = " +
-    'rowData[' + i + '] == null ? null : parsers[' + i + '](rowData[' + i + ']);'
 }
 
 Result.prototype.addFields = function (fieldDescriptions) {
@@ -97,18 +91,11 @@ Result.prototype.addFields = function (fieldDescriptions) {
     this.fields = []
     this._parsers = []
   }
-  var ctorBody = ''
   for (var i = 0; i < fieldDescriptions.length; i++) {
     var desc = fieldDescriptions[i]
     this.fields.push(desc)
     var parser = this._getTypeParser(desc.dataTypeID, desc.format || 'text')
     this._parsers.push(parser)
-    // this is some craziness to compile the row result parsing
-    // results in ~60% speedup on large query result sets
-    ctorBody += inlineParser(desc.name, i)
-  }
-  if (!this.rowAsArray) {
-    this.RowCtor = Function('parsers', 'rowData', ctorBody)
   }
 }
 
--- a/package.json
+++ b/package.json
@@ -20,7 +20,6 @@
   "main": "./lib",
   "dependencies": {
     "buffer-writer": "1.0.1",
-    "js-string-escape": "1.0.1",
     "packet-reader": "0.3.1",
     "pg-connection-string": "0.1.3",
     "pg-pool": "~2.0.3",


lib/result.js
```diff
@@ -65,29 +64,24 @@ Result.prototype._parseRowAsArray = function (rowData) {
   return row
 }
 
-// rowData is an array of text or binary values
-// this turns the row into a JavaScript object
 Result.prototype.parseRow = function (rowData) {
```
contra: This is a simple `reduce` but I kept the same style as the array parser for consistency.

brianc:
Is there a reason for this change?  Last time we checked (~6 months ago) the eval provides around at 30% speed up over setting fields like you're doing.  Is the current code causing you problems?

brianc:
Ah i see now in the referenced issue, sorry missed that in the PR comments here.

contra:
@brianc Would love to know how you got those #s so I can reproduce them and refine this to match, or think of an alternative solution so we don't lose any performance. Ultimately though, if the dynamic function stuff is an inherent memory issue I think not crashing from out of memory errors is more important than 30% faster.

brianc:
One thing is the dynamic function stuff has been in the code for ~5 years.  I am slightly circumspect of it causing issues that have only been discovered now given how widely and heavily used this module is across the node ecosystem.  As far as the benchmarking goes - just run a query that returns 500 rows w/ the npm published version of pg.  Run it 10,000 times and take the median execution time.  Now do the same w/ this patch - compare.  When I get some time I'll take a look at the script provided in the issue & see if there's anything to track down.  Also interested in @charmander's thoughts.

contra:
@brianc Yeah, that also strikes me as strange. We only started having issues with actual OOM crashes when our writes hit ~10k RPM - I would assume somebody out there has a workload of > ~10K queries per minute on this library, but maybe not.

Neamar:
@contra interestingly I had the exact same thing, when using pg "reasonably" I had no issue, until one of my workers had to insert thousands of records per minute. Really curious why the rate at which you send queries would impact memory usage!

knownasilya:
I used this fix and here are the results:

before
before


after
screen shot 2018-04-17 at 10 29 40 am

calvinmetcalf:
I'm a colleague of @knownasilya and can vouch that his app isn't doing anything obviously stupid or weird, just a bunch of select statements

lib/result.js
```diff
@@ -97,18 +91,11 @@ Result.prototype.addFields = function (fieldDescriptions) {
...
   }
-  var ctorBody = ''
   for (var i = 0; i < fieldDescriptions.length; i++) {
     var desc = fieldDescriptions[i]
     this.fields.push(desc)
     var parser = this._getTypeParser(desc.dataTypeID, desc.format || 'text')
     this._parsers.push(parser)
-    // this is some craziness to compile the row result parsing
-    // results in ~60% speedup on large query result sets
-    ctorBody += inlineParser(desc.name, i)
```
calvinmetcalf: if you guys are dead set against this pull, then at least one improvement that could be definitely made to pg is to make ctorBody an array, have this line push the inline parser string into the array and then do an array.join at the end, currently here we are making 2 strings for each field of the results

brianc:
I'm not dead set against it. One thing you can do in the mean time before I merge this is use `rowMode: array` in your queries and then re-assemble the object shape yourself from the query result.  But I think this needs to be merged since it causes a memory leak.

jasonfill:
+1 on this merge. We are facing this and currently looking at options to resolve if the module is not updated. Fixing this leak would greatly help us as well.

knownasilya:
I'm using this branch in an app, and have started getting this error https://github.com/brianc/node-postgres/issues/1105

```
TypeError: Cannot read property 'name' of null at Connection.<anonymous> (/usr/src/project/node_modules/pg/lib/client.js:266:26) at emitOne (events.js:116:13) at Connection.emit (events.js:211:7) at Socket.<anonymous> (/usr/src/project/node_modules/pg/lib/connection.js:123:12) at emitOne (events.js:116:13) at Socket.emit (events.js:211:7) at addChunk (_stream_readable.js:263:12) at readableAddChunk (_stream_readable.js:250:11) at Socket.Readable.push (_stream_readable.js:208:10) at TCP.onread (net.js:607:20)
```

mikecules:
++1 We are also having issues with the leak on a high traffic endpoint. Would love to to see this merged 🚀

brianc:
Going to merge this into the `6.x` branch. It's not a breaking change in any way & w/ the new v8 having similar performance it should only improve things.

bleupen:
will there be a v6 release with this change? we are unable to upgrade to v7 at this time unfortunately.

knownasilya:
We backported this fix to v5 because we ran into other issues. Will try to mention those once I can find the error we had 👍

PR #1606, opened 2018-04-04
Test against current node lts

Elexy:
Also test against node.js 9


Elexy:
Good one @brianc, I added one boron test back in.

brianc:
👏 thanks!

PR #1607, opened 2018-04-04
Stream end

Elexy:


Vratislav:
@Elexy hello, this is something that I was thinking of adding because our connections were not closing correctly when running against postgres on Azure. Why have you closed it?

Elexy:
The tests aren't passing: https://travis-ci.org/brianc/node-postgres/jobs/362160404
It's a bit beyond me why.

Vratislav:
@Elexy I will look into it.

Vratislav:
@Elexy I've submitted new PR with passing tests and a proper write-up: #1608

If you are ok with it, you can close this PR in favor of the one mentioned above.

Thanks for pointing the fix into right direction. 👍

PR #1626, opened 2018-04-25
Don't call connect until connect handler is set up

titanous:
If a `stream` is specified, the connect event will be emitted immediately, and it will be dropped unless the event handlers are set up.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -65,12 +65,6 @@ Client.prototype.connect = function (callback) {
   }
   this._connecting = true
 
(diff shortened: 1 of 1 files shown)


titanous:
Looks like the tests have some subtle dependency on the order of operations here. Perhaps someone more familiar with the codebase can fix that and pull in my patch?

charmander:
Tests might pass after #1503.

charmander:
They didn’t. Time to find out what else is in `client.connection.stream.packets`, then; it doesn’t look like md5-password-tests covers the full connection startup.

PR #1638, opened 2018-05-05
tls.connect({checkServerIdentity}) option cannot be a null - must be …

matthiasak:
…a method or not exist.

Defaults to built-in `tls.checkServerIdentity` method in the event one is not passed into `pgConfig.ssl`

Found breaking in v7.4.2 vs v7.4.1 a la 49054717b4ec0c6d477f04c2becd1f9680b2d13a

cc @tobio @brianc

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -94,7 +94,7 @@ Connection.prototype.connect = function (port, host) {
     self.stream = tls.connect({
       socket: self.stream,
       servername: host,
-      checkServerIdentity: self.ssl.checkServerIdentity,
+      checkServerIdentity: self.ssl.checkServerIdentity || tls.checkServerIdentity,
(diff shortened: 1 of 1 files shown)


madd512:
the error:

```
AssertionError [ERR_ASSERTION]: false == true
    at Object.exports.connect (_tls_wrap.js:1036:3)
```

vitaly-t:
The issue is easy to test, so the fix should include a test where we pass in `ssl = true`.

charmander:
@vitaly-t Tests aren’t set up for SSL. Working on it.

brianc:
oops sorry about that - pushing new patch version now

brianc:
published fix at `pg@7.4.3`

PR #1639, opened 2018-05-07
fix: tls.checkServerIdentity cannot be undefined

robertrossmann:
If this configuration option is provided explicitly and is not a function (ie. the intention was to "use default value") this assertion will not pass because Node applies default values to the configuration options via prototypal inheritance, however the property lookup will stop immediately because the `checkServerIdentity` property actually exists on the provided object (with a value of `undefined`).

Related: #1444


robertrossmann:
Oh, closing as duplicate of #1638.

PR #1651, opened 2018-05-17
Error Handling for null Params to Client.prototype.query()

### LINKED ISSUE #628: client.query breaks if null is passed into query
This isn't too much of a concern, but if the query param is passed into client.query as null, node crashes and the callback is never made. For instance:

``` javascript
var query = null
client.query(query, function(err, res) {
  if(err) { } //handle err
}) 
```

Produces: 

``` javascript
/Users/dbell/work/curious/node_modules/pg/lib/client.js:323
  var query = (typeof config.submit == 'function') ? config :
                            ^
TypeError: Cannot read property 'submit' of null
    at Client.query (/Users/dbell/work/curious/node_modules/pg/lib/client.js:323:29)
```

patrickrgaffney:
Add's a basic assertion to `Client.prototype.query()` ensuring that `config` is not `null` or `undefined`. This fixes #628 — which was closed, but the behavior described in that issue lives on. I added some additional tests around the new logic.

I ran into the same problem described in that issue — some bad destructing logic on my end led to `config` being `undefined` when calling `query()`. That results in a `TypeError` when the `typeof config.submit === 'function'` assertion is executed. The behavior is still basically the same, except that the error is detected, described, and sent to the users `callback` or `Promise.reject`ed.

### DIFF
--- a/test/unit/client/simple-query-tests.js
+++ b/test/unit/client/simple-query-tests.js
@@ -120,4 +120,24 @@ test('executing query', function () {
       })
     })
   })
+
+  test('handles errors', function () {
+    var client = helper.client()
+
+    test('throws an error when config is null', function () {
+      try {
+        client.query(null, undefined)
+      } catch (error) {
+        assert.equal(error.message, 'Client was passed a null or undefined query', 'Should have thrown an Error for null queries')
+      }
+    })
+
+    test('throws an error when config is undefined', function () {
+      try {
+        client.query()
+      } catch (error) {
+        assert.equal(error.message, 'Client was passed a null or undefined query', 'Should have thrown an Error for null queries')
+      }
+    })
+  })
 })
--- a/lib/client.js
+++ b/lib/client.js
@@ -365,7 +365,10 @@ Client.prototype.query = function (config, values, callback) {
   // can take in strings, config object or query object
   var query
   var result
-  if (typeof config.submit === 'function') {
+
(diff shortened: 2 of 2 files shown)


sehrope:
Looks good to me.

abenhamdine:
why not simply check config falsyness ?
I don't think a falsy value could be useful here, or maybe I miss something ?

charmander:
Type errors tend to throw synchronously. I don’t see a reason to break that pattern for `null` and `undefined` specifically – it just needs a more specific error message – so how about:

```js
if (config == null) {
  throw new TypeError('Query was null or undefined')
}
```

patrickrgaffney:
Just made the changes and `rebase`d such that the `TypeError` would always `throw` synchronously — the tests have also been updated.

test/unit/client/simple-query-tests.js
```diff
@@ -120,4 +120,25 @@ test('executing query', function () {
       })
     })
   })
+
+  test('handles errors', function () {
+    var client = helper.client()
+    var error = new Error('Client was passed a null or undefined query')
```
charmander: Is this variable unused?

patrickrgaffney: Yep, must have missed it when I fixed the tests. I'll push up again.

brianc:
Thanks for this! sorry for the delay. Will push a new version soon.

PR #1660, opened 2018-05-25
Tie default DB name to database user name

solidsnack:
The previous behaviour was at variance with the native client behaviour, per:
    
https://www.postgresql.org/docs/10/static/libpq-connect.html#LIBPQ-PARAMKEYWORDS
    
> The database name. Defaults to be the same as the user name.
> In certain contexts, the value is checked for extended formats...

### DIFF
--- a/lib/connection-parameters.js
+++ b/lib/connection-parameters.js
@@ -13,7 +13,7 @@ var defaults = require('./defaults')
 
 var parse = require('pg-connection-string').parse // parses a connection string
 
-var val = function (key, config, envVar) {
+var val = function (key, config, envVar, computed) {
   if (envVar === undefined) {
     envVar = process.env[ 'PG' + key.toUpperCase() ]
   } else if (envVar === false) {
@@ -22,9 +22,13 @@ var val = function (key, config, envVar) {
     envVar = process.env[ envVar ]
   }
 
+  if (computed === undefined) {
+    computed = function (key) { return defaults[key] }
+  }
+
   return config[key] ||
     envVar ||
-    defaults[key]
+    computed(key)
 }
 
 var useSsl = function () {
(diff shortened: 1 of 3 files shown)


solidsnack:
I am really not sure about the test `'modified values are passed to created clients'` in `test/integration/client/configuration-tests.js`, with regards to setting `defaults.database`.

* Maybe the ability to set `defaults.database` should be removed? It's not clear how to reconcile it with the behavior of `libpq`, "Defaults to be the same as the user name.". Maybe it should only take effect if no username is set?

* I'm not sure I understand the use case for the defaults.

* The motivation for this patch, is that we have some `DATABASE_URLS` set across our infra that rely on the `libpq` behavior, that is also broadly compatible with other Postgres clients.

* It seems like something changed here, perhaps in the way Knex uses `node-postgres`, because these URLs used to work fine with our Node applications, too.

charmander:
Note that this is a breaking change for anything setting or relying on `pg.defaults.database`.

lib/connection-parameters.js
```diff
@@ -50,8 +54,11 @@ var ConnectionParameters = function (config) {
     config = Object.assign({}, config, parse(config.connectionString))
   }
 
+  var dbDefaulting = function () { return this.user || defaults['database'] }
```
charmander: Is it possible for `this.user` to be falsy?

solidsnack: Looking at the implementation, it's set with `val('user', config)` which can be false when (a) there is no default set and (b) the corresponding environment variable is not set.

From the outside, URLs like `postgres:///` don't have a user set, so it stands to reason the implementation has to allow it to be falsey.

solidsnack:
> Note that this is a breaking change for anything setting or relying on `pg.defaults.database`.

Perhaps there is some intermediate possible? The current behavior is such that `node-postgres` interprets Postgres URLs in a manner different from `libpq` and other implementations. I guess the first question to answer is: is that something we're okay with or would we like to see Postgres URLs work the same with Node apps as they do with things like `psql` and `psycopg2`?

charmander:
A breaking change just means it needs a major version increment.

lib/connection-parameters.js
charmander: The default has to be set to something; if it’s `undefined`, the connection is going to fail anyway. I think this can just be `const getDefaultDatabase = () => this.user`.

brianc:
I will merge this as soon as I cut the next major version. Sorry for the delay.

brianc:
Actually looks like 1679 supersedes this, so closing in favor of that one.

PR #1666, opened 2018-05-31
Update dependencies

kibertoad:


charmander:
If there’s a specific dependency that needs updating, the PR should just update that dependency. Need to be careful about backwards compatibility. (pg-types 1.13, for example, is not actually backwards-compatible with 1.12.)

kibertoad:
@charmander What about dev dependencies? Should be safe enough to update these?..

charmander:
Which dev dependency needs an update?

kibertoad:
@charmander Yeah, pretty much ESLint. Shall I create a PR just for that?

PR #1678, opened 2018-06-17
Add optional support for BufferList as zero-copy binary parameters

sehrope:
Adds handling of BufferList (via "bl" module) so that they are treated like Buffer parameters. Without built-in support in pg itself users have to call `.slice()` on the BufferList which forces creating a copy of all the data. With built-in support the BufferList parameters get written directly as binary parameters to the outbound socket just like regular Buffer parameters.

If the "bl" module is not found at init then this patch reverts to the original behavior of treating only Buffer objects as binary data. If it is found it treats either Buffer or BufferList objects as binary parameters. The "bl" module is listed as a peer dependency in package.json so it's not installed by default but is indicated as such.

PR passes all the existing tests with and without bl installed and I've tested it locally using BufferList parameters and parameters that are arrays that include BufferLists. Haven't added the test here though as the "bl" module isn't a required dependency so it doesn't get included in the Travis harness. Need to figure a more pleasant way of doing that as I think it'd involve breaking out the test infrastructure to be a custom script so we have more control over the order of installing deps, removing them, etc. Didn't want to include that in this yet as it seems more invasive and should best be handled separately (though I still do want to look into it).

Thoughts?

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,9 @@
     "eslint-plugin-standard": "3.0.1",
     "pg-copy-streams": "0.3.0"
   },
+  "peerDependencies": {
+    "bl": "^2"
+  },
   "minNativeVersion": "2.0.0",
   "scripts": {
     "test": "make test-all"
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -13,6 +13,7 @@ var util = require('util')
 
 var Writer = require('buffer-writer')
 var Reader = require('packet-reader')
+var utils = require('./utils')
 
 var TEXT_MODE = 0
 var BINARY_MODE = 1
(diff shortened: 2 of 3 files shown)


sehrope:
FYI the Travis errors aren't related to this PR. Looks like some delay in resolving dummy hostnames is causing some other test to not terminate within 5-seconds:

```
promise-api-tests.js
  valid connection completes promise ✔
  valid connection completes promise ✔
  invalid connection rejects promise 
Error: test: invalid connection rejects promise did not complete withint 5000ms
    at Timeout.setTimeout [as _onTimeout] (/home/travis/build/brianc/node-postgres/test/suite.js:54:19)
    at ontimeout (timers.js:498:11)
    at tryOnTimeout (timers.js:323:5)
    at Timer.listOnTimeout (timers.js:290:5)
```

Here's the code for that test:

```js
suite.test('invalid connection rejects promise', (done) => {
  const client = new pg.Client({ host: 'alksdjflaskdfj' })
  return client.connect()
    .catch(e => {
      assert(e instanceof Error)
      done()
    })
})
```

package.json
```diff
@@ -38,6 +38,9 @@
     "eslint-plugin-standard": "3.0.1",
     "pg-copy-streams": "0.3.0"
   },
+  "peerDependencies": {
+    "bl": "^2"
```
brianc: Sorry I'm kinda bad w/ `peerDependencies` versus `optionalDependencies` and when to use each.  Does making `bl` a `peerDependency` mean it'll be installed along w/ the main `pg` module?  Or just that if the user installs it on their end it needs to be `bl@2.x`?

sehrope: Yes it's a bit confusing as they kind of work backwards to how you'd expect from the naming.

Peer dependencies aren't installed by default as its expected the "main" app's package.json will include them. I think npm prints a warning at npm install time if there are unmet peer dependencies.

Optional dependencies are installed by default if there's a version available for your target platform. The primary usage is if you've got a platform specific module (say with an arch specific DLL) and you only want it installed if it exists. For pure-JS modules that don't restrict the arch I think they're always installed as they'd always match up.

It's a bit of a hack to use peerDependencies as true optional dependencies as it's a warning and not an error if they're not actually installed.

Thinking about this some more, maybe the simplest solution is not to include any dependency item at all and just add a note to the docs about it. Anybody using this would have to install "bl" at the top level anyway to use it in their app and the require statement within pg will be able to find it.

sehrope:
I'm closing this out for now as there's probably a better / generic way to support things like this.

PR #1679, opened 2018-06-17
Use user name as default database when user is non-default

charmander:
Not entirely backwards-compatible. Improved version of #1660.


brianc:
@charmander would you mind re-targeting `bmc/8.0` branch instead of master? I'm gonna land all the PRs there & then do one merge + semver bump + release on master. 🙏

brianc:
Just so I'm clear on the release notes here this change is something like...:

> If a user name is not specified, available in the environment, or available at `pg.defaults`, we used to use the username of the process user.  Now we will use the username supplied to the client.  What this means is if you used to have
> 
> ```
> new Client({
>   username: 'foo'
> })
> ```
> We would default the database name _to the process user_ where as now we will use the username _supplied to the client_.  If you have not supplied `username` to the client, and it isn't available through any of its existing lookup mechanisms (environment variables, `pg.defaults`) then it will still use _the process user_ for the database name.

(aside: we should just nuke pg.defaults in 9.0...I feel like it was a bad api design)

charmander:
@brianc Yep! That’s nicely detailed. (`username` should be `user` in the final version, mind.)

> (aside: we should just nuke pg.defaults in 9.0...I feel like it was a bad api design)

I’m all for that too.

PR #1684, opened 2018-07-09
Replace poolSize by max in error test

dominicletz:
It seems poolSize has been replaced by max end of 2017 in most places of the codebase but this one. I assume that was a miss.

### DIFF
--- a/test/integration/connection-pool/error-tests.js
+++ b/test/integration/connection-pool/error-tests.js
@@ -2,10 +2,8 @@
 var helper = require('./test-helper')
 const pg = helper.pg
 
-// first make pool hold 2 clients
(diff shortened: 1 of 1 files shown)


charmander:
Can the line be removed entirely? I’m not sure what it’s for.

dominicletz:
I think the line is there to make it explicit that at least two concurrent connections are needed for this test. Of course the current default is 10 and so you could remove it now - but then the functioning of the test is bound to the default being 2 or bigger.

charmander:
It might be worth passing `{ max: 2 }` directly to the constructor, then, since it looks like it affects other tests right now.

dominicletz:
Agreed! Made that change

Issue #1674, opened 2018-06-13
Errored QueryStream causes Client to throw on message receive


rkaw92:
Hi,
I have encountered a really bizarre issue when using `knex` with `pg` and query streams. A parameterized QueryStream which fails in the parsing stage (e.g. because of a column type / value type mismatch) breaks *subsequent queries* on the same connection - either by:
* throwing errors on incoming data packets from Postgres, terminating the Node process like this:
```
/home/thewanderer/Devel/Node/pg-prepared-statements/node_modules/pg/lib/client.js:212
    self.activeQuery.handleRowDescription(msg)
                    ^
TypeError: Cannot read property 'handleRowDescription' of null
    at .<anonymous> (/home/thewanderer/Devel/Node/pg-prepared-statements/node_modules/pg/lib/client.js:212:21)
    at emitOne (events.js:101:20)
    at emit (events.js:188:7)
    at Socket.<anonymous> (/home/thewanderer/Devel/Node/pg-prepared-statements/node_modules/pg/lib/connection.js:123:12)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at readableAddChunk (_stream_readable.js:177:18)
    at Socket.Readable.push (_stream_readable.js:135:10)
    at TCP.onread (net.js:542:20)
```
* or by messing with the output of subsequent queries - queries receive empty `result.rows` (this is much harder to reproduce and will not be covered here, but it occurs in one of our apps using `knex`)

Below are two ways to reproduce this issue, one using `knex` and another simplified to bare `pg` (which simulates how `knex` interacts with `pg`).

**Test conditions:**
* A table in the DB with a column of type "uuid": `CREATE TABLE events (id UUID);`
  * Note: if the table does not exist, this bug also occurs the same way.
* `pg@7.4.3`
* `pg-query-stream@1.1.1`
* `knex@0.14.6` (optional, for the knex-enabled test)
* `PostgreSQL 9.6.7` (this probably does not matter)

```js
'use strict';

function runKnex() {
	const knex = require('knex');
	const db = knex({
		client: 'pg',
		connection: {
			host: process.env.PGHOST,
			user: process.env.PGUSER,
			password: process.env.PGPASSWORD,
			database: process.env.PGDATABASE
		},
		pool: {
			min: 1,
			max: 1
		}
	});
	const resultStream = db('events').select('*').where({ id: 'not a UUID' }).stream();
	resultStream.on('data', function() {});

[...]

PR #1685, opened 2018-07-11
Fixes #1105

### LINKED ISSUE #1105: TypeError: Cannot read property 'name' of null

mrhwick:


brianc:
Hey nice! Is there any way to write a test for this?

brianc:
Thanks for the test! I'll wait on @charmander for a second pair of :eyes: and then merge if it looks good to him.

charmander:
Silently ignoring invalid data without knowing why it’s there seems dangerous.

mrhwick:
If the query isn't there, then the name isn't there either. This doesn't
change the semantics of the conditional or the functioning of this event
listener. It only safeguards against a null reference exception.

charmander:
> This doesn't change the semantics of the conditional

It does. The original is “mark this query as parsed if it has a name”, because preparing a query without a name is normal (it’s how all parameterized queries work). The new meaning is “make sure the active query exists because it might not because ???”

mrhwick:
If it doesn't exist, it can't have a name.

mrhwick:
https://github.com/brianc/node-postgres/blob/ecab41c3f3d07230b24253e804971367cd04fb53/lib/client.js#L128-L135

charmander:
> Since this can happen **under some set of circumstances**

Which ones? This is really important.

mrhwick:
[Here is a set of circumstances where this error occurs](https://github.com/brianc/node-postgres/issues/1674). 

Maybe this PR could be changed to throw a more descriptive error rather than the TypeError that is currently thrown? If you would rather continue throwing a TypeError up to your users, I'll close this PR.

mrhwick:
Understood. No appetite for throwing a more meaningful error to your users. Closing this PR instead of changing it to do that.

PR #1706, opened 2018-08-10
Replace crypto usage for crypto-js to generate md5 when using AuthenticationMD5Password

sramirezayuso:
Replaces the usage crypto.md5 in favour of crypto-js.md5 to generate the md5 when using AuthenticationMD5Password.
When using a FIPS build of NodeJS which uses the OpenSSL FIPS module calling crypto.md5 causes the underlying OpenSSL FIPS module to fail with the error: 
`Message: Error: error:060A80A3:digital envelope routines:FIPS_DIGESTINIT:disabled for fips`

By replacing the usage of crypto this allows AuthenticationMD5Password to be used even with a FIPS build of NodeJS. In my particular case I ran into this issue because I'm using Amazon RDS and it does not allow configuring the authentication method.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -7,7 +7,7 @@
  * README.md file in the root directory of this source tree.
  */
 
-const crypto = require('crypto')
+const cryptojs = require('crypto-js')
 
 const defaults = require('./defaults')
 
@@ -142,7 +142,7 @@ function normalizeQueryConfig (config, values, callback) {
 }
 
 const md5 = function (string) {
-  return crypto.createHash('md5').update(string, 'utf-8').digest('hex')
+  return cryptojs.MD5(string).toString()
 }
 
 // See AuthenticationMD5Password at https://www.postgresql.org/docs/current/static/protocol-flow.html
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
   "main": "./lib",
   "dependencies": {
     "buffer-writer": "1.0.1",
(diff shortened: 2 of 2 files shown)


charmander:
This doesn’t seem like a great idea. Use a non-FIPS Node.js?

sramirezayuso:
Unfortunately it's not always an option like in this particular project for reasons of FIPS compliance. I believe that enabling such scenarios would be a good thing.
Out of curiosity, why exactly doesn't it seem like a great idea? Because of the need to add a dependency to pg?

charmander:
> Out of curiosity, why exactly doesn't it seem like a great idea? Because of the need to add a dependency to pg?

In a way. It’s extra complexity with next to no benefit. You want to disable MD5, but keep using MD5…? Is that even FIPS-compliant at all?

sramirezayuso:
FIPS disallows the use of MD5 as a way of securing data for its known vulnerabilities but usage of MD5 itself doesn't create any inherent security vulnerability. Assuming the connection to Postgres is done using SSL the fact that the MD5 of the password is transmitted is not a security concern.

As to your point of being extra complexity with next to no benefit I agree that the number of use cases is limited. That being said, considering that FIPS support is part of NodeJS having pg support it as well would benefit at least some users.

sramirezayuso:
Yes, that is going to be our approach but I just wanted to see if we could include it in the library.

Thanks for taking the time of reviewing the PR anyway!

PR #1711, opened 2018-08-29
Multiple hosts support fix #1470

### LINKED ISSUE #1470: PostgreSQL 10 support
PostgreSQL v10 was released on Oct 05, 2017

---

One of the most interesting feature that affects `node-postgres` is [multiple host names or addresses in libpq connection strings and URIs](https://www.postgresql.org/docs/10/static/libpq-connect.html#libpq-connect-host).

This feature allows automatic support for replicas on the connection level, i.e. the first connection is tried, and if fails - the next one, and so on.

[From documentation](https://www.postgresql.org/docs/10/static/libpq-connect.html#libpq-multiple-hosts):

> It is possible to specify multiple hosts to connect to, so that they are tried in the given order. In the Keyword/Value format, the host, hostaddr, and port options accept a comma-separated list of values. The same number of elements must be given in each option, such that e.g. the first hostaddr corresponds to the first host name, the second hostaddr corresponds to the second host name, and so forth. As an exception, if only one port is specified, it applies to all the hosts.

denchistyakov:


denchistyakov:
Fix #1711

charmander:
Looks like this is missing implementation for the pure JS driver? Also, a second module for connection string parsing that isn’t compatible with [libpq connection strings][1] AIUI doesn’t seem right vs. adding to pg-connection-string. Tests are necessary as well.

  [1]: https://www.postgresql.org/docs/10/static/libpq-connect.html#id-1.7.3.8.3.6

vitaly-t:
> It works with all types of documented URI except for unix sockets

@denchistyakov why doesn't it work with the sockets?

The standard for sockets inside a connection string is to have the server address ending with `.sock`, which works perfectly well with [connection-string].

Here's example:

```js
const parse = require('connection-string');
const s = parse('postgres://user:pass@server-address.sock/myDB');
console.log(s);
```
outputs:
```
ConnectionString {
  protocol: 'postgres',
  user: 'user',
  password: 'pass',
  hosts: [ { name: 'server-address.sock', isIPv6: false } ],
  path: [ 'myDB' ] }
```

[connection-string]:https://github.com/vitaly-t/connection-string

charmander:
> In what kind of cases? It's work with all types of documented URI except for unix sockets, for them I use previous module.

For example: when there are spaces in some part. libpq handles them fine. I’m sure there are many other differences (since MongoDB and libpq are unrelated), but that’s an obvious one because tests had to be removed.

vitaly-t:
> when there are spaces in some part

[MongoDB spec for connections strings](https://github.com/mongodb/specifications/blob/master/source/connection-string/connection-string-spec.rst) is the most sensible one. And it requires that the server address is URL-encoded, which means spaces must be passed in as `%20`.

This generic approach is much safer than all the custom/proprietary ones used so far by this driver. It should finally shift towards use of the generic connection string.

charmander:
> MongoDB spec for connections strings is the most sensible one.

For a PostgreSQL driver? No, I think libpq’s implementation is the most sensible compatibility target.

jimlindeman:
Is there anything blocking this functionality from being merged at this time?  Is it the missing tests for this capability?

A project I'm working on needs this functionality to use multiple HA-proxy layers  in front of the postgres DB's, so we 'HA' the HA-proxies.

radoslawroszkowiak:
@charmander, which implementation do you mean? PG native?
Why not to allow multiple hosts when on libpq, and i.e. ignore any but first host and throw some warning when native in use?

vitaly-t:
@radoslawroszkowiak Libq does support the feature natively, b.t.w., so it should be done for both at the same time, IMO.

charmander:
@radoslawroszkowiak I mean the non-pg-native implementation. The JS driver.

PR #1724, opened 2018-09-12
Remove excess `module.exports.native` assignment

tlhunter:
I believe assigning `native` to `module.exports.native` is redundant.
The getter definition on line 43 should immediately replace the value.


charmander:
`module.exports.native = native` is inside the getter (and happens after `delete module.exports.native`).

tlhunter:
My bad, I completely glossed over the lazy loading purpose of the function.

PR #1730, opened 2018-09-20
remove query.stream references

tlhunter:
* This hasn't been supported since 0b2344b6b5afbb68e89eff1ef2b57ecf0726d80b
  * The `startStreamingToConnection` method only existed in `lib/copystream.js`
  * See: https://www.google.com/search?q=startStreamingToConnection
* `node-pg-copy-streams` relies on overriding the `handleCopyInResponse` method:
  * https://github.com/brianc/node-pg-copy-streams/blob/e15feb19/index.js#L53
  * Which is why I wrap the method calls in an if in the EE handlers

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -25,7 +25,6 @@ var Query = function (config, values, callback) {
   this.types = config.types
   this.name = config.name
   this.binary = config.binary
-  this.stream = config.stream
   // use unique portal name each time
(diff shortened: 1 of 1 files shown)


tlhunter:
@charmander good point, I enabled it again

brianc:
Thanks for the little cleanup!

PR #1753, opened 2018-10-23
Update ESLint

kibertoad:
Also make minor formatting change to make new rules pass


charmander:
Not introduced by this pull request, but I’m not sure why eslint-plugin-import is a dependency.

kibertoad:
@charmander Should I remove it?

brianc:
@kibertoad - thanks for this! If it's not needed it'd be nice to remove it, yeah.  If you find out it's needed lmk.  Either way I'll merge this whenever you're ready.  Thanks for doing the somewhat thankless work of upgrading deps.

kibertoad:
Sorry, will create a separate pr for eslint-plugin-import removal.

Issue #1713, opened 2018-09-03
Feature: query timeout


juliusza:
I want to start a discussion about built in query timeout. 

In our production environment, we encountered a problem with TCP connections getting stuck. Knex pool would fill up with connections that are not responsive. Note that this would be different from statement timeout, where DB decides to terminate query. Instead pg client would decide that connection is unresponsive and emit an error.

This situation can be simulated with firewall rule:

> sudo iptables -A INPUT -j DROP -p tcp --destination-port 5432

Or just use this code snippet:
```
const { Client } = require('pg')
const client = new Client({user: "user", password: "pass", database: "db"})
const util = require('util');
const exec = util.promisify(require('child_process').exec);

async function main() {
   await exec("sudo iptables -D INPUT -j DROP -p tcp --destination-port 5432");
   await client.connect()

   console.log("Client connected");
   await exec("sudo iptables -A INPUT -j DROP -p tcp --destination-port 5432");

   console.log("Now waiting for DB to respond... forever")
   const res = await client.query('SELECT $1::text as message', ['Hello world!'])
   console.log(res.rows[0].message) // Hello world!
   await client.end()
   process.exit(0)
}

main()
```

A possible implementation can be found here: https://github.com/brianc/node-postgres/compare/master...juliusza:query_timeout
I know I need to add tests and also edit the native driver code for a proper PR. But first I'm looking for feedback from the community to assert if such a thing would be useful.

Also perhaps
> socket.setTimeout(3000);

could be a viable solution. I need to look into that.

abenhamdine:
Happy to see that !

Do you plan to add tests and send a PR ?

edevil:
Yes, please!

edevil:
@juliusza You need help with the PR?

@brianc Would this be merged?

brianc:
I'm definitely down for a pull request for this feature.  I think it'd be helpful & something the library should provide since it's pretty difficult to implement in your own code.

edevil:
@brianc I’ve submitted a PR for this. https://github.com/brianc/node-postgres/pull/1760

neylsongularte:
Same problem with me, my temporaly solution:

```
        var socket = new require('net').Socket();
        socket.setTimeout(20000)

        const client = new pg.Client({
            connectionString: urlPostgres,
            stream: socket
        });

        socket.on('timeout', function () {
            // console.log('timeout')
            client.end();
        });

        await client.connect()
```

[...]

PR #1760, opened 2018-11-02
Add configurable query timeout

edevil:
Addresses #1713 
Based on work from @juliusza

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -421,7 +421,7 @@ Client.prototype.query = function (config, values, callback) {
   }
 
   if (readTimeout) {
-    queryCallback = query.callback;
+    queryCallback = query.callback
 
     readTimeoutTimer = setTimeout(() => {
       var error = new Error('Query read timeout')
--- a/lib/client.js
+++ b/lib/client.js
@@ -427,11 +427,15 @@ Client.prototype.query = function (config, values, callback) {
       var error = new Error('Query read timeout')
 
       this.emit('error', error)
-      queryCallback(error)
 
-      // we already returned an error,
-      // just do nothing if query completes
-      query.callback = () => {}
+      if (typeof queryCallback === 'function') {
+        queryCallback(error)
+
+        // we already returned an error,
+        // just do nothing if query completes
+        query.callback = () => {
+        }
+      }
 
       // Remove from queue
       var index = this.queryQueue.indexOf(query)
--- a/lib/client.js
+++ b/lib/client.js
@@ -424,20 +424,22 @@ Client.prototype.query = function (config, values, callback) {
     queryCallback = query.callback;
 
     readTimeoutTimer = setTimeout(() => {
-      var error = new Error('Query read timeout');
+      var error = new Error('Query read timeout')
 
       this.emit('error', error)
       queryCallback(error)
 
       // we already returned an error,
-      // just do nothing when query completes
-      delete query.callback
+      // just do nothing if query completes
+      query.callback = () => {}
 
       // Remove from queue
       var index = this.queryQueue.indexOf(query)
       if (index > -1) {
         this.queryQueue.splice(index, 1)
       }
+
+      this._pulseQueryQueue()
     }, readTimeout)
 
     query.callback = (err, res) => {
--- a/lib/client.js
+++ b/lib/client.js
@@ -431,14 +431,11 @@ Client.prototype.query = function (config, values, callback) {
         query.handleError(error, this.connection)
       })
 
-      if (typeof queryCallback === 'function') {
-        queryCallback(error)
+      queryCallback(error)
 
-        // we already returned an error,
-        // just do nothing if query completes
-        query.callback = () => {
-        }
-      }
+      // we already returned an error,
+      // just do nothing if query completes
+      query.callback = () => {}
 
       // Remove from queue
       var index = this.queryQueue.indexOf(query)
--- a/lib/client.js
+++ b/lib/client.js
@@ -403,6 +403,10 @@ Client.prototype.query = function (config, values, callback) {
   if (config === null || config === undefined) {
     throw new TypeError('Client was passed a null or undefined query')
   } else if (typeof config.submit === 'function') {
+    var readTimeoutTimer
+    var queryCallback
+    var readTimeout = config.query_timeout || this.connectionParameters.query_timeout
+
     result = query = config
     if (typeof values === 'function') {
       query.callback = query.callback || values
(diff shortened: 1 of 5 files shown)


test/integration/client/api-tests.js
```diff
@@ -15,6 +15,17 @@ suite.test('pool callback behavior', done => {
   })
 })
 
+suite.test('query timeout', (cb) => {
+  const pool = new pg.Pool({query_timeout: 1000})
+  pool.connect().then((client) => {
+    client.query('SELECT pg_sleep(2)', assert.calls(function (err, result) {
+      assert(err)
```
kibertoad: can we assert what kind of error we get?

test/integration/client/api-tests.js
```diff
@@ -15,6 +15,28 @@ suite.test('pool callback behavior', done => {
...
+      client.release()
+      pool.end(cb)
+    }))
+  })
+})
+
+suite.test('query timeout', (cb) => {
```
kibertoad: is name correct?

edevil: Right, I'll change it right away. Thanks.

test/integration/client/api-tests.js
edevil: i'll check the message.

edevil: @kibertoad done

test/integration/client/api-tests.js
edevil: @kibertoad done

lib/client.js
```diff
@@ -416,6 +421,40 @@ Client.prototype.query = function (config, values, callback) {
...
+      if (index > -1) {
+        this.queryQueue.splice(index, 1)
+      }
+
+      this._pulseQueryQueue()
+    }, readTimeout)
+
+    query.callback = (err, res) => {
+      clearTimeout(readTimeoutTimer)
+      queryCallback(err, res)
```
joelmukuthu: Not sure, but should there be a check for `typeof queryCallback` before making this call (seeing as there was one on line 434)?

edevil: Yes, I don't think it's needed after all. I've removed it.

juliusza:
Nice! I have briefly tested my code with query timeout enabled in production and it seemed to work fine.

Eventually I found that the issue was caused by misconfigured tcp proxy, and decided not to pursue this. Regardless, this could be very useful to recover stuck TCP sessions due to lossy network conditions @brianc. Also this is completely optional, and the lib will function as it did without the setting enabled.

edevil:
@juliusza Is that documentation in some other repository?

abenhamdine:
> @juliusza Is that documentation in some other repository?  

Unfortunately, so far the docs are not available for contributions.

brianc:
> Unfortunately, so far the docs are not available for contributions.

I'm thinking of porting them to [https://www.gitbook.com] so it's easier to contribute.

edevil:
@brianc what about this PR?

brianc:
Have the flu in bed. Will check as soon as im back in my feet. Havenf
forgotten

abenhamdine:
> Have the flu in bed. 

Winter is coming :mask:

edevil:
@brianc how’s that flu coming along? :)

brianc:
The flu suuuuuuuuuuuuuuucked!  Thanks for asking. :smile: It cleared up early last week & now I just have left-over bronchitis.  Definitely going to get the flu shot from here on out.  Now...lemme look at this PR! Sorry for the review (thanks for pinging me gently on it!)

abenhamdine:
> This is great, thanks! I'll get this merged & push out a new minor version today. heart

That's great :+1: Could you also bump the version of pg-pool to 2.0.4, in order to be able to benefit from https://github.com/brianc/node-pg-pool/commit/1871d0f9e18330afe8149ff5df8b3e54ac71469a ? thx !

brianc:
@abenhamdine you bet!

neylsongularte:
I'm looking forward to this functionality

edevil:
@brianc Hello! Any news on this?

kibertoad:
@brianc Is pg-pool bump also happening?

brianc:
@kibertoad yah definitely, right after this

published `pg@7.7.0`

brianc:
pg-pool is already published at `2.0.4` as of a [few weeks ago](https://github.com/brianc/node-pg-pool/commit/4b9669eaa764126135f0bbf64af9be1e9d5092a9).  Am I missing something / are you not able to install it?

kibertoad:
@brianc It's listed in package.json as `"pg-pool": "~2.0.3"` which means that projects with package-lock.json are not going to pick it up unless you regenerate entire lock file.

brianc:
ohhh duh yeah i can fix that no problem, sorry

brianc:
should I just change it to `pg-pool:^2.0.3`?  I think that'd be better.

brianc:
hmmm wait the `~2.0.3` doesn't prevent you from installing `2.0.4` it just doesn't upgrade it if you already have 2.0.3 installed.  K, i'll just bump it.

brianc:
K published `pg@7.7.1` bumping the min version for pg-pool to `2.0.4`

abenhamdine:
great, thx @brianc !

kibertoad:
Thank you!

jfromaniello:
Does this actually cancel the query in the server once it timedout in the client side? isn't clear from the diff.

Thank you!

juliusza:
Hey @jfromaniello, this does not cancel query on the server. It would be reasonable to configure both query_timeout and statement_timeout, making the statement_timeout slightly longer so that it times out right after the client stopped waiting for response.

jfromaniello:
@juliusza thank you very much!

johan13:
How is the documentation for this feature coming?

The `query_timeout` property is missing from the typescript typings, and the typings seem to be based on the docs, so I thought I'd start here before I put in a merge request to [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/pg).

brianc:
I recently open-sourced the docs!
https://github.com/brianc/node-postgres-docs . so feel free to add any
documentation you want over there!

tmtron:
> Hey @jfromaniello, this does not cancel query on the server. It would be reasonable to configure both query_timeout and statement_timeout, making the statement_timeout slightly longer so that it times out right after the client stopped waiting for response.

Hm.. I thought it should be the other way around: the query-timeout should be slightly longer than the statement-timeout.
i.e. when the statement-timeout is reached, the database will cancel the query and return the [error ](https://www.postgresql.org/docs/11/errcodes-appendix.html) `57014 query_canceled` to the client: so the client gets the error-details and knows exactly what happened.
If, the server cannot return any result/error back to the client for some reason, then the query-timeout will kick in - and just stop listening for the server result.

@juliusza what do you think?



`

juliusza:
Yes @tmtron, make sense to have `query-timeout` longer than `statement-timeout`.

`[CRITICAL] TypeError: Cannot read property 'handleCommandComplete' of null
at Connection.<anonymous> (/opt/acn/release_20190925/api/node_modules/pg/lib/client.js:325:22)`

I found that sometimes I get this error on production. Because pg lib is event based, there's no proper stack trace to work with. Obviously we haven't thought of a certain edge case here, that causes above error to crash node process.

I no longer need the query-timeout setting, so I don't think I'll want to spend time working on a fix.

PR #1764, opened 2018-11-07
Bump buffer-writer version

brianc:
Fixes the deprecation warning for using `new Buffer`.

The change is semver major in buffer-writer since we dropped support for node < 4.x, but otherwise it's a non-breaking change.  Since node-postgres already requires node >= 4.x it's fine and can be released here as semver patch.


brianc:
published `pg@7.6.1` - thanks for your help @LinusU

PR #1788, opened 2018-12-12
Include raw SQL statement and values in query errors

### LINKED ISSUE #881: Display the SQL text in the errors.
Right now you get an error but it is not linked to the query text and it is hard to debug.

I've added this in lib/query.js line 103 (in the function handleError):

```js
err.sql = this.text;
```

Seems to work, but I don't know the library well enough and I don't know what else it could affect.

marshall007:
Fixes #881

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -142,9 +142,15 @@ function normalizeQueryConfig (config, values, callback) {
 }
 
 function handleQueryError (err, query) {
-  err.sql = {
-    text: query.text,
-    values: query.values
+  if (query.sql_in_error === true || query.sql_in_error === 'text') {
+    err.sql = {
+      text: query.text,
+      values: query.values
+    }
+  }
+
+  if (query.sql_in_error === 'text') {
+    delete err.sql.values
   }
 
   // if callback supplied do not emit error event as uncaught error
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -141,6 +141,20 @@ function normalizeQueryConfig (config, values, callback) {
   return config
 }
 
+function handleQueryError (err, query) {
+  err.sql = {
+    text: query.text,
+    values: query.values
+  }
+
+  // if callback supplied do not emit error event as uncaught error
+  // events will bubble up to node process
+  if (query.callback) {
+    return query.callback(err)
+  }
+  query.emit('error', err)
+}
+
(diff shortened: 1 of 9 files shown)


lib/utils.js
```diff
@@ -141,6 +141,20 @@ function normalizeQueryConfig (config, values, callback) {
   return config
 }
 
+function handleQueryError (err, query) {
+  err.sql = {
+    text: query.text,
+    values: query.values
```
joelmukuthu: Including values introduces a security issue. This could lead to user data e.g. passwords being logged in production. Perhaps this should be configurable or removed altogether.

marshall007:
I've addressed @joelmukuthu's concern in dd7352c. There is a new config option, `sql_in_error` (totally open to a better name there), which accepts the following values:
- `false`: none (default, current behavior)
- `true`: `text` and `values` from query
- `"text"`: only `text` from query

mledom:
This is great, I was actually trying to find a way to get that information.

brianc:
I'm not a fan of doing this - there's a security issue here in that we could be causing apps to accidentally log out sensitive information (e.g. a hashed password in the query params).  It also is something that can/should be handled by a higher level library or your own application code if this kind of logging is required.  Our responsibility as a database driver is to propagate the error back to the caller w/ the details included from the PostgreSQL backend.

abenhamdine:
> there's a security issue here in that we could be causing apps to accidentally log out sensitive information

It would be indeed unacceptable in many businesses (including our) and uncompliant with European regulation on personal data protection (GDPR).
Please enable it only with an option

marshall007:
> ... there's a security issue here in that we could be causing apps to accidentally log out sensitive information (e.g. a hashed password in the query params).

@brianc because the new behavior is strictly opt-in, I tend to disagree here. I'd be happy to add some logic that explicitly warns the user of your concern when `NODE_ENV === 'production'` and `sql_in_error !== false`. That would seem like a reasonable middle ground?

brianc:
I think the best approach here is to log your query text & sql out in your own application if you need to see it in your logs.  I'm not comfortable putting code in that has the possible to leak secure information...even less so if it's non-compliant with regulation.  I don't see a reason this has to be the responsibility of node-postgres.

maxpain:
Any updates?

PR #1800, opened 2019-01-03
Update license date

benhc123:
From 2018 to 2019


brianc:
Thanks & Happy new year!

Issue #1769, opened 2018-11-12
pg does not support TLS/SSL protocol restriction.


glepsky:
We need to restrict communication protocols to TLS 1.1 and TLS 1.2. I have tried to set `poolOptions.ssl.secureOptions =  SSL_OP_NO_TLSv1 | SSL_OP_NO_SSLv3 | SSL_OP_NO_SSLv2` and then call `new pg.Pool(poolOptions);`. 
However,  `ssl.secureOption` is not supported by pg - in connection.js, during streamcreation the `secureOptions` property is not copied to `tls.connect()` argument (`ConnectionOptions`).

Adding `secureOptions: self.ssl.secureOptions` to the `ConnectionOptions` instance seems to fix the problem.

Would you accept a fix that would copy over the `ssl.secureOptions` to `tls.connect()` argument?
Something like 
```
   self.stream = tls.connect({
      socket: self.stream,
      servername: host,
      checkServerIdentity: self.ssl.checkServerIdentity || tls.checkServerIdentity,
      rejectUnauthorized: self.ssl.rejectUnauthorized,
      ca: self.ssl.ca,
      pfx: self.ssl.pfx,
      key: self.ssl.key,
      passphrase: self.ssl.passphrase,
      cert: self.ssl.cert,
      secureOptions: self.ssl.secureOptions,
      NPNProtocols: self.ssl.NPNProtocols
    })
```

brianc:
Yes I definitely would accept a fix for that!

glepsky:
@brianc Created a pull request https://github.com/brianc/node-postgres/pull/1804 for this issue.

PR #1804, opened 2019-01-07
Issue #1769 - Inability to restrict communication protocols to set in…

glepsky:
… ssl configuration

- Propagate client's ssl.secureOptions config to TLS.


brianc:
Nice thank you!

glepsky:
Thanks for accepting the change!

PR #1806, opened 2019-01-08
deps: update pg-types to 2

bendrucker:
pg-types@2 drops node < 4 support but pg previously did so, so the upgrade isn't major for pg itself

See also: https://github.com/brianc/node-pg-types/pull/76

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "packet-reader": "0.3.1",
     "pg-connection-string": "0.1.3",
     "pg-pool": "^2.0.4",
-    "pg-types": "~1.12.1",
+    "pg-types": "~2.0.0",
     "pgpass": "1.x",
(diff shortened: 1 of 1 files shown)


brianc:
Just to confirm - this is semver patch for this library once we merge this?

bendrucker:
> this is semver patch for this library once we merge this

Yup! Because `{"engines": {"node": ">= 4"}}` was declared in an earlier release, it's not breaking to upgrade pg-types to a version that has the same version requirements (where it previously had none).

brianc:
sweet! merging now & releasing.  Gonna release semver minor because [this](https://github.com/brianc/node-postgres/commit/ccbccc9716a62adf06515143a575d26b73db0789) was added which is _technically_ a new feature.

PR #1814, opened 2019-01-15
compare prepared statement text with previous to prevent incorrect queries

### LINKED ISSUE #1813: Error if prepared statement name isn't unique
This is more an admission of incompetence on my part than anything, but on a number of occasions I've copy-pasted a query like this...

```js
const { rows } = await pool.query({
  name: 'some-descriptive-name',
  text: `select thingamajig from flibbertigibbets where id = $1`,
  values: [id]
});
```

...and tweaked the query but forgotten to change the name:

```js
const { rows } = await pool.query({
  name: 'some-descriptive-name',
  text: `select thingamajig from flibbertigibbets where id = $1 and active = true`,
  values: [id]
});
```

Typically, this results in subtle errors, by returning plausible but incorrect data.

I wonder if `pg` could help here by storing the `text` of a query that specifies a `name`, and erroring if it encounters another query with the same name but different text? Arguably not the library's job, but if my experience is anything to go by it could save a few developers from tearing their hair out, and deepen their gratitude towards this wonderful project.

Does that seem like a worthwhile addition? If so I'd be glad to try and rustle up a PR. Thank you!

Rich-Harris:
Closes #1813

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -276,7 +276,7 @@ Client.prototype._attachListeners = function (con) {
   // it again on the same client
   con.on('parseComplete', function (msg) {
     if (self.activeQuery.name) {
-      con.parsedStatements[self.activeQuery.name] = true
+      con.parsedStatements[self.activeQuery.name] = self.activeQuery.text
     }
   })
 
--- a/lib/native/query.js
+++ b/lib/native/query.js
@@ -140,7 +140,7 @@ NativeQuery.prototype.submit = function (client) {
     // if so...just execute it again - skip the planning phase
     if (client.namedQueries[this.name]) {
       if (this.text && client.namedQueries[this.name] !== this.text) {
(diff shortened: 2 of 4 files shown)


sehrope:
+1. Overall looks good but I'd changed the error message a bit. It's got a Unicode &mdash; (`&mdash;`) rather than regular dash (`"-"`). Also the reference to plural "queries" should be singular and it could be any type of statement.

How about:

```text
Prepared statements must be unique - '${this.name}' was used for a different statement
```

brianc:
Nice I'll release a new semver patch w/ this change ASAP! Sorry for the delay.

Rich-Harris:
excellent, thank you!

brianc:
published `pg@7.8.2` - thanks again for your help!

Issue #1680, opened 2018-06-30
Drastic difference between pg and pg-native performance inserting bytea data


gajus:
Schema:

```sql
CREATE TABLE test (
    id SERIAL PRIMARY KEY,
    data bytea
);

CREATE UNIQUE INDEX test_pkey ON test(id int4_ops);

```

`test.js`:

```js
import fs from 'fs';
import pg from 'pg';

const image = fs.readFileSync('./test.jpg');

const main = async () => {
  const client = new pg.native.Client();

  await client.connect();

  let index = 1000;

  while (index--) {
    await client.query('INSERT INTO "test" (data) VALUES ($1) RETURNING id', [
      image
    ]);
  }

  await client.end();
};

main();

```

[test.jpg](https://user-images.githubusercontent.com/973543/42129684-ba81c4b6-7cc3-11e8-8b5d-87c367423319.jpg)

Evaluation:

```js
$ time node ./test.js

```

Result with `pg`:

```
0.88s user
0.39s system
0% cpu
2:37.37 total

```

Result with `pg-native`:

```

4.95s user
0.16s system
19% cpu
26.118 total

```

That is a huge difference. What is the overhead?

sehrope:
It's likely the additional overhead of copying the large block of memory for the image that you're inserting. 

There are two types of async external APIs in node. The first is APIs that are truely async, this includes things like interactions with TCP sockets and the file system. APIs that are not natively async pretend to be async by utilizing an internal thread pool. From the perspective of a running node app the operations happen async but under the covers they're being run in a background thread rather than using real event notification. That covers most external libraries that are integrated with node as they're *usually* not async.

The `pg-native` module is built atop the `libpq` module which exposes the libpq C library to node. It uses the internal thread pool rather that true async operation as that's how the primary API for libpq itself works (i.e. blocking sockets): https://www.npmjs.com/package/libpq#connection-functions

One negative side effect of using the internal thread pool is that it causes extra copying of memory from the shared context of the node app by the thread that will execute in the thread pool. This is to prevent concurrent memory corruption as otherwise the shared buffer could be modified while the thread pool is operating on it. The bigger the data being copied (ex: a large bytea image), the more overhead that copy is going to impose.

FYI, I've noticed a similar drop in performance with the zlib module when performing gzip compression async vs. sync. The async operation uses the internal thread pool and results in an extra memory copy that's noticeable when compressing large Buffers.

[...]

PR #1850, opened 2019-03-07
Use bytea parameter type for Buffers

### LINKED ISSUE #1849: Error with null character in parameter
```js
client.query('SELECT $1::bytea as buffer', [new Buffer(1)])
```

results in

```
ERROR: invalid byte sequence for encoding "UTF8": 0x00
```

pauldraper:
Fixes #1849

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -230,7 +230,11 @@ Connection.prototype.parse = function (query, more) {
     .addCString(query.text) // actual query text
     .addInt16(len)
   for (var i = 0; i < len; i++) {
-    buffer.addInt32(query.types[i])
+    if (query.types.length <= i && query.values[i] instanceof Buffer) {
+      buffer.addInt32(17); // bytea
+    } else {
+      buffer.addInt32(query.types[i]);
+    }
   }
 
   var code = 0x50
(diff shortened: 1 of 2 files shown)


brianc:
Thanks @pauldraper - would you mind fixing the lint so the tests can run & pass? Really appreciate the work here!

brianc:
I'll need to double-check w/ the current behavior of using buffers as parameters.  This _might_ be backwards incompatible if it changes existing behavior.

sehrope:
@pauldraper Are you seeing this error with the native driver (i.e. "pg-native"), the pure JS one, or both?

I'm not able to reproduce this with the pure-JS driver and I think it's specific to pg-native. The native driver transforms all query parameters to strings and that's where the zero-byte issue comes in (PostgreSQL strings can't have zero bytes). See here for more details on pg-native parameter handling: https://github.com/brianc/node-postgres/issues/1680#issuecomment-401631727

Rather than deal with this one case here, the real fix may be changes to pg-native to handle parameter transformation for Buffer (ex: to hex) and possibly other types as well.

charmander:
#1849:

> I was not using native. It appears that I actually had the problem with the string `'\0'`, not `new Buffer(1)`.

PR #1855, opened 2019-03-16
fix: Catch and emit query parameter prepareValue(...) errors

### LINKED ISSUE #1854: Rollback issues
This program will never terminate, as the rollback query infinitely blocks:

```
const { Pool } = require('pg')
const pool = new Pool()

async function main() {

  try {
    await client.query('BEGIN');
    await client.query('SELECT $1', [{x: 1n}]);
    await client.query('COMMIT')
  } catch (e) {
    throw e
  } finally {
    client.release()
  }

  console.log('great success! Terminated!');
}

main();
```

Note, the value `1n` works,  `{x: 1}` works, but `{x: 1n}` is where the problems start

sehrope:
Adds a try/catch block around the `prepareValue(...)` invocations in `query.prepare(...)` to ensure that any that throw an error are caught and bubbled up to the caller. Otherwise the `.prepare(...)` stops halfway through without reporting the error back, hanging the connection.

Also adds a test that uses a custom serialization function to simulate an error. That should match up with any other serialization errors include circular references or unsupported `JSON.stringify(...)` types. The test is disabled for `pg-native` as it can't handle non-string parameters (it just segfaults if you try giving it an object).

Should fix #1854 to actually report the `JSON.stringify(...)` bignum error back to the caller.

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -194,7 +194,12 @@ Query.prototype.prepare = function (connection) {
   }
 
   if (self.values) {
-    self.values = self.values.map(utils.prepareValue)
+    try {
+      self.values = self.values.map(utils.prepareValue)
+    } catch (err) {
+      this.handleError(err, connection)
+      return
+    }
   }
 
   // http://developer.postgresql.org/pgdocs/postgres/protocol-flow.html#PROTOCOL-FLOW-EXT-QUERY
--- a/test/integration/gh-issues/1854-tests.js
+++ b/test/integration/gh-issues/1854-tests.js
@@ -1,0 +1,33 @@
+"use strict"
+var helper = require('./../test-helper')
+
+const suite = new helper.Suite()
+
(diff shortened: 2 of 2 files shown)


madeken:
Does this testcase fail when the fix is not applied?

sehrope:
@madeken Yes without the fix applied the new test fails complaining about unhandled promise rejections. Also, with the fix applied, if you replace the test object with `{x: 1n}` it'll reject with an error from the JSON serialization failing: `TypeError: Do not know how to serialize a BigInt`

I didn't use a BigInt value for the test case to ensure that it runs on the older versions of node still supported by this module. The flow being verified is any synchronous error serializing parameter values so it should cover that as well though.

brianc:
awesome! thanks @sehrope! sorry for the delay - releasing a new version today w/ this fix.

PR #1864, opened 2019-03-30
Fix input preparation for date objects with a BC year

noinkling:
JavaScript `Date` objects can represent BC years, represented as negative years (or `0`). However they're currently not serialized correctly for input:

```js
> var date = new Date('-000024-01-01')
> date.toISOString()
'-000024-01-01T00:00:00.000Z'
> prepareValue(date)
'0-24-01-01T00:00:00.000+00:00'  // Just causes an error when sent to the DB
```

This PR fixes it so that it produces PostgreSQL's accepted format:

```js
> prepareValue(date)
'0025-01-01T00:00:00.000+00:00 BC'
```

There is a bit of a tricky aspect to all this. You may notice that the "year -24" gets changed to "25 BC" - this is not an accident, it arises because of the discrepancy in representation of BC dates between JS and Postgres. Please read the issue I opened at bendrucker/postgres-date#5, as it outlines the issue and provides proof that this is the correct behaviour.

I should note that with this PR, currently a round trip will fail, because the output parsing doesn't account for this nuance, as seen in the linked issue. Moreover fixing that end of things might be considered a breaking change.

Edit: I've opened a PR for that at bendrucker/postgres-date#6

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -97,7 +97,12 @@ function pad (number, digits) {
 
 function dateToString (date) {
   var offset = -date.getTimezoneOffset()
-  var ret = pad(date.getFullYear(), 4) + '-' +
+
+  var year = date.getFullYear()
+  var isBCYear = year < 1
+  if (isBCYear) year = Math.abs(year) + 1 // negative years are 1 off their BC representation
+
+  var ret = pad(year, 4) + '-' +
     pad(date.getMonth() + 1, 2) + '-' +
     pad(date.getDate(), 2) + 'T' +
     pad(date.getHours(), 2) + ':' +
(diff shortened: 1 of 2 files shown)


noinkling:
The fixes to postgres-date were published in 1.0.4, which pg-types will automatically pick up as it specifies `"postgres-date": "~1.0.0"`. This in turn breaks an integration test here which I have a separate PR for at #1867.

I also opened a PR at brianc/node-pg-types#79 to explicitly depend on at least that version of postgres-date. That will ensure that round trips behave properly if this PR is merged.

brianc:
Fantastic! Thank you! 💃

PR #1867, opened 2019-04-04
Fix integration test for v1.0.4 of postgres-date sub-dependency

noinkling:
Should probably depend on brianc/node-pg-types#79, which contains more details.

Edit: That PR was merged.


brianc:
nice! thanks!

PR #1890, opened 2019-05-08
Skip TLS SNI if host is IP address

jrmerz:
Currently the ```host``` is always set to the tls.connect() servername option, even if the host is a ip address.  This triggers the following depreciation warning: ```Setting the TLS ServerName to an IP address is not permitted by RFC 6066```.

The patch will only set the servername option in tls.connect() if ```net.isIP(host) === 0```.

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -91,7 +91,7 @@ Connection.prototype.connect = function (port, host) {
         return self.emit('error', new Error('There was an error establishing an SSL connection'))
     }
     var tls = require('tls')
-    let options = {
+    const options = {
       socket: self.stream,
       checkServerIdentity: self.ssl.checkServerIdentity || tls.checkServerIdentity,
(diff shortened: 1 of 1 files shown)


lib/connection.js
```diff
@@ -103,7 +102,12 @@ Connection.prototype.connect = function (port, host) {
       cert: self.ssl.cert,
       secureOptions: self.ssl.secureOptions,
       NPNProtocols: self.ssl.NPNProtocols
-    })
+    };
+    if( net.isIP(host) === 0 ) {
```
charmander: Formatting for the check to pass:

```suggestion
    if (net.isIP(host) === 0) {
```

charmander: (Also, no semicolons)

brianc:
thank you! sorry for the delay - new release w/ this coming out on Wednesday.

PR #1912, opened 2019-06-20
Update CI to allow tests to pass on Node.js >= 10.16.0

darkgl0w:
Hello ^^

Since Node.js 10.16.0 `libpq` must be compiled with `OpenSSL 1.1.1b` at least to avoid [those failures](https://travis-ci.org/brianc/node-postgres/jobs/546971383).

With this PR the Travis CI environment will meet all these requirements without any breaking change.

> NB: the configuration scripts have been added to .npmignore in order to avoid bundling them on release

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -7,3 +7,4 @@ script/
 *.swp
 test/
 .travis.yml
+ci_scripts/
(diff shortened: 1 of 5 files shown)


brianc:
hmmm looks like something weird changed w/ deep-equal checking in node v12.  I can fix this if you want - thanks so much for this work here!

brianc:
I'm gonna merge this & fix the node 12 failure on another commit.

PR #1918, opened 2019-07-07
Using correct JSON format for Array

anonimizerme:
There is fix for array to string conversion.
It's correct to use square brackets instead braces.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -23,7 +23,7 @@ function escapeElement (elementRepresentation) {
 // uses comma separator so won't work for types like box that use
 // a different array separator.
 function arrayString (val) {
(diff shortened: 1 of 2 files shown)


anonimizerme:
Need update some tests.

sehrope:
The array handling for this module is fine, those are Postgres arrays, not JSON arrays and the driver defaults to serializing arrays as Postgres arrays. If you want to send a JSON array you'll need to `JSON.stringify(...)` it yourself and use the text representation of the JS array as the parameter value.

See the Postgres docs (https://www.postgresql.org/docs/current/arrays.html) and https://github.com/brianc/node-postgres/issues/442 for some more info on this.

PR #1922, opened 2019-07-16
Fix deepEqual compare

brianc:
In node 12 assert.deepEqual against a buffer & array no longer passes if the values are the same.  This makes sense.  Updated the test to not use deepEqual in this case.


brianc:
holy smokes vscode went nuts with the code formatting...lemme try that again....

PR #1925, opened 2019-07-17
feat: extend errors for trackability

eljefedelrodeodeljefe:
Naive PR incoming. Intention, below.

While the low level pg error message is sufficient for tracking low level issues with pg, finding the root cause for a deeply passed error is hard.

Implementers should be able to opt into enhanced error behaviour, such es simply sending the root query / text.

Signed-off-by: Robert Jefe Lindstaedt <robert.lindstaedt@gmail.com>

The error trace is often not useful, e.g. on statement cancelled due to timeout. Ideally the error being returned should have a stacktrace actually stating the origin of that call to client.query. However I assume this is rather hard and a nice opt-in addition would be sending query.text

error__canceling_statement_due_to_statement_timeout

error__canceling_statement_due_to_statement_timeout

### DIFF
--- a/lib/query.js
+++ b/lib/query.js
@@ -29,6 +29,7 @@ var Query = function (config, values, callback) {
   this.portal = config.portal || ''
   this.callback = config.callback
   this._rowMode = config.rowMode
+  this.extendedError = config.extendedError || false
   if (process.domain && config.callback) {
     this.callback = process.domain.bind(config.callback)
   }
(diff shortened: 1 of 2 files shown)


MattGohr:
@eljefedelrodeodeljefe Are you still working on this do you need any help?  I need this to go in as well to get better debug sql errors.

eljefedelrodeodeljefe:
@brianc what would you think?

dekdevy:
+1, this would be very useful. or is there a workaround?

charmander:
Duplicate of #1788

PR #1926, opened 2019-07-17
feat: Add dynamic retrieval for client password

sehrope:
PR to add support for dynamic password retrieval. There was already something quite similar to this to support reading passwords from a PGPASS file. This PR expands that section a bit to support invoking an arbitrary function.

Rather than supporting a callback parameter, this PR allows for either a synchronous function that returns a string or one that returns a Promise that resolves to a string. As there's Promise support on every supported runtime I figure it's a better choice for anything new being added.

Here's some examples of how to use it:

```js
// Existing hard coded password:
const pool = new pg.Pool({
    password: 'test',
});

// Sync return a static string
const pool = new pg.Pool({
    password: () => {
        return 'test';
    }
});

// Async API
const pool = new pg.Pool({
    password: async () => {
        await new Promise(resolve => setTimeout(resolve, 1000));
        const value = await someAsyncFunc();
        return value;
    }
});
```

This PR does not have any new tests yet, I just tested it with some local examples. If the concept itself seems workable then should be easy to add a few that use a function and check for its side effects. Let me know.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -114,27 +114,23 @@ Client.prototype._connect = function (callback) {
 
   function checkPgPass (cb) {
     return function (msg) {
-      if (typeof(self.password) === 'function') {
-        try {
-          self._Promise.resolve(self.password())
-            .then(pass => {
-              if (undefined !== pass) {
-                if (typeof(pass) !== 'string') {
-                  con.emit('error', new TypeError('Password must be a string'))
-                  return
-                }
-                self.connectionParameters.password = self.password = pass
-              } else {
-                self.connectionParameters.password = self.password = null
+      if (typeof self.password === 'function') {
+        self._Promise.resolve()
+          .then(self.password)
+          .then(pass => {
+            if (pass !== undefined) {
+              if (typeof pass !== 'string') {
+                con.emit('error', new TypeError('Password must be a string'))
+                return
               }
-              cb(msg)
-            })
-            .catch(err => {
-              con.emit('error', err)
-            })
-        } catch (err) {
-          con.emit('error', err)
-        }
+              self.connectionParameters.password = self.password = pass
+            } else {
+              self.connectionParameters.password = self.password = null
+            }
+            cb(msg)
+          }).catch(err => {
+            con.emit('error', err)
+          })
       } else if (self.password !== null) {
         cb(msg)
       } else {
--- a/lib/client.js
+++ b/lib/client.js
@@ -116,7 +116,7 @@ Client.prototype._connect = function (callback) {
     return function (msg) {
       if (typeof self.password === 'function') {
         self._Promise.resolve()
-          .then(self.password)
+          .then(() => self.password())
           .then(pass => {
             if (pass !== undefined) {
               if (typeof pass !== 'string') {
--- a/lib/client.js
+++ b/lib/client.js
@@ -114,7 +114,28 @@ Client.prototype._connect = function (callback) {
 
   function checkPgPass (cb) {
     return function (msg) {
-      if (self.password !== null) {
+      if (typeof(self.password) === 'function') {
+        try {
+          self._Promise.resolve(self.password())
+            .then(pass => {
+              if (undefined !== pass) {
+                if (typeof(pass) !== 'string') {
+                  con.emit('error', new TypeError('Password must be a string'))
+                  return
+                }
+                self.connectionParameters.password = self.password = pass
+              } else {
+                self.connectionParameters.password = self.password = null
+              }
+              cb(msg)
+            })
+            .catch(err => {
+              con.emit('error', err)
+            })
+        } catch (err) {
+          con.emit('error', err)
+        }
+      } else if (self.password !== null) {
         cb(msg)
       } else {
         pgPass(self.connectionParameters, function (pass) {
--- a/test/integration/connection/dynamic-password.js
+++ b/test/integration/connection/dynamic-password.js
@@ -8,7 +8,7 @@ const Client = pg.Client;
 const password = process.env.PGPASSWORD || null
 const sleep = millis => new Promise(resolve => setTimeout(resolve, millis))
 
-suite.testAsync('Get password from a sync function', function () {
+suite.testAsync('Get password from a sync function', () => {
     let wasCalled = false
     function getPassword() {
         wasCalled = true
@@ -24,7 +24,7 @@ suite.testAsync('Get password from a sync function', function () {
         })
 })
 
-suite.testAsync('Throw error from a sync function', function () {
+suite.testAsync('Throw error from a sync function', () => {
     let wasCalled = false
     const myError = new Error('Oops!')
     function getPassword() {
@@ -47,7 +47,7 @@ suite.testAsync('Throw error from a sync function', function () {
         })
 })
 
-suite.testAsync('Get password from a function asynchronously', function () {
+suite.testAsync('Get password from a function asynchronously', () => {
     let wasCalled = false
     function getPassword() {
         wasCalled = true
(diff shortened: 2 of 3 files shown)


sehrope:
I force pushed an update to this that changes:

* Explicitly sets the password field to null if the password function does not return a usable value
* Changes the error for a non-string return value to be a TypeError

Also added some tests for retrieving a sync, async, and errant passwords via functions. For the tests I added a new helper `suite.testAsync(...)` to simplify writing them using promises.

lib/client.js
```diff
@@ -114,7 +114,28 @@ Client.prototype._connect = function (callback) {
 
   function checkPgPass (cb) {
     return function (msg) {
-      if (self.password !== null) {
+      if (typeof(self.password) === 'function') {
+        try {
+          self._Promise.resolve(self.password())
```
brianc: ohhh nice use on supporting BYOP - not sure how popular BYOP is now that promises have shipped natively in node for years, but still nice for backwards compatibility 👌

lib/client.js
```diff
@@ -114,7 +114,28 @@ Client.prototype._connect = function (callback) {
...
+                self.connectionParameters.password = self.password = null
+              }
+              cb(msg)
+            })
+            .catch(err => {
+              con.emit('error', err)
+            })
+        } catch (err) {
+          con.emit('error', err)
+        }
```
brianc: wow - impressive error handling

test/suite.js
```diff
@@ -72,6 +72,19 @@ class Suite {
     const test = new Test(name, cb)
     this._queue.push(test)
   }
+
+  testAsync (name, action) {
```
brianc: i apologize the tests here aren't written w/ a modern node testing framework.  There's this cost-benefit thing I keep running into where it's like "Should I re-write all these tests into jest/mocha/etc? They already work and I *know* they work...but maybe rewriting will introduce some breakage."  The original tests in the repo were written before mocha or jest existed so it's kinda like...well....if it ain't broke...otoh I feel like it's probably more confusing to contribute to.  I also at the time, coming from ruby, thought i'd be 'clever' and attach a bunch of stuff to the global namespace like assert and other things.  That was a mistake. :(

test/integration/connection/dynamic-password.js
```diff
@@ -0,0 +1,113 @@
...
+const Client = pg.Client;
+
+const password = process.env.PGPASSWORD || null;
+const sleep = millis => new Promise(resolve => setTimeout(resolve, millis))
+
+suite.testAsync('Get password from a sync function', function () {
+    let wasCalled = false
+    function getPassword() {
+        wasCalled = true
+        return password;
```
charmander:
```suggestion
        return password
```

test/suite.js
```diff
@@ -72,6 +72,19 @@ class Suite {
...
+    const test = new Test(name, cb => {
+      try {
+        Promise.resolve(action())
+          .then(() => cb(null))
+          .catch((err) => cb(err))
```
charmander: This can call `cb` a second time if it throws. How about:

```js
Promise.resolve()
  .then(action)
  .then(() => { cb(null) }, cb)
```

lib/client.js
sehrope: Apparently not as impressive as @charmander ;-)

test/suite.js
sehrope: No worries. I've thought about cleaning up some of them more than once and it's always the same conclusion. Maybe having a separate harness for things going forward as it's definitely easier to write async tests with an async harness. Doesn't even have to be be full on mocha/jest/whatever ... just having something to call out to with async actions should be enough.

I like how you can easily run any one of the tests in this project via "node path/to/test.js".

test/suite.js
sehrope: Oh yes this is way better. In fact I don't think the double wrapped try/catch for sync errors is needed either as the `action` will get invoked via regular Promise semantics and auto convert it to a rejection.

sehrope:
I pushed some updates to address the comments (thanks @charmander!). As part of this I also cleaned up the error handling in client.js to use the same style of `Promise.resove()` chaining as it removes the need for separate sync error handling.

I left the error handling for that piece on L131 of client.js as a separate `.catch(...)` to handle anything being thrown by the supplied `cb` rather than a single `.then(onSuccess, onError)`. Anything that errors out would then bubble up as an error event on the client. 

FYI, taking a peek at the callers of that `checkPgPass(...)` function, I think I may have found a potential future issue with the SCRAM auth, I'm going to open a separate issue to track it.

lib/client.js
```diff
@@ -114,7 +114,24 @@ Client.prototype._connect = function (callback) {
 
   function checkPgPass (cb) {
     return function (msg) {
-      if (self.password !== null) {
+      if (typeof self.password === 'function') {
+        self._Promise.resolve()
+          .then(self.password)
```
charmander: I didn’t comment on this one intentionally – both because of the different error-handling needs but because using `.then` here causes it to act a bit differently. Having only one error case is nice, though, I guess, so do we mind that

- `.then` passes `undefined` to `self.password` instead of no arguments, and
- the `this` value is a promise instead of `self` in `self.password`?

If we do, the fix isn’t terrible: `.then(() => self.password())`.

test/integration/connection/dynamic-password.js
```diff
@@ -0,0 +1,113 @@
+'use strict'
+const assert = require('assert')
+const helper = require('./../test-helper')
+const suite = new helper.Suite()
+const pg = require('../../../lib/index');
+const Client = pg.Client;
```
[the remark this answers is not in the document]
charmander: (This one still exists)

lib/client.js
sehrope: Nice catch! Yes having it be explicit via `() => self.password()` is a good idea to ensure that nothing is supplied as an arg. I was thinking down the road it make make sense to add the connection parameters or some other structure to the caller similar to how the `pgPass(...)` piece works.

sehrope:
No clue why it failed on node v12 in the PR build but passed for me (https://travis-ci.org/sehrope/node-postgres/builds/561008527). There's no specific error in the travis logs either. I think there was some transient error building Postgres.

I'm going to push a dummy commit to see if it happens again.

test/integration/connection/dynamic-password.js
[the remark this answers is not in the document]
brianc: yeah i need to turn on linting on the test folder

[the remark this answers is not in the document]
sehrope: Would also help if we turned back on the semi rule. This PR passed "make lint" as the semi rule got removed when the eslint config was last cleaned up (by me...).

alekstr:
Btw, checked and tested this change in Prod with RDS Postgres 11.4, works fine!

brianc:
Right now! & I'll document it and release a new minor version. 😄

brianc:
Published `pg@7.12.0` - thanks!

benkeil:
You missed the types...

sehrope:
@benkeil Yes the typings are maintained externally in DefinitelyTyped (i.e. `@types/pg`) so it can't be updated in tandem. I submitted a PR for that so hopefully that'll get synced soon: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/49257

benkeil:
Naive question: why not migrate the project to typescript?

sehrope:
I think it will eventually and some of the internal components have already been migrated to Typescript (e.g. see the packet parser).

brianc:
> Naive question: why not migrate the project to typescript?

yeah i'm very slowly working on that

vitaly-t:
This PR and the change are confusing. I thought at first this would actually allow dynamic passwords support, but all it does - makes only one call in the beginning. After that, if the password changes, the callback is never called again. So it is not very useful then.

I had to use a connection object, and then do `pool.options.password = 'new-password'`, or if connection string used, then set new one for `pool.options.connectionString`, in order to start using a new password.

PR #1928, opened 2019-07-19
Fix eslint and add back standard

sehrope:
Enables back standard mode for eslint and updates the dependencies. Some of the dep updates break on older node versions. Since linting is the same regardless of which node version is running it, I disabled it for those versions. It only runs for node >=8 now.


brianc:
ah nice! Yeah I'm gonna drop support for LTS versions of node which are no longer supported on the next breaking release.

sehrope:
:tada:

+1 to dropping end of life versions. Allows for graceful modernization as well.

PR #1937, opened 2019-07-26
Update package.json

### LINKED ISSUE #1936: Missing builtins / old pg-types reference
According to [this](https://github.com/brianc/node-pg-types), we have access to `types.builtins` from the `pg` driver, in the following manner:

```js
const types = require('pg').types;
const someType = types.builtins.TIMESTAMPTZ;
```
In reality, however, `pg.types` does not have property `builtins`, because `pg-types` implemented it with version `2.1.0`, but we include it as `~` instead of `^` like everything else.

And so there is discrepancy between the official documentation and reality.

vitaly-t:
Fixes #1936

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "packet-reader": "1.0.0",
     "pg-connection-string": "0.1.3",
     "pg-pool": "^2.0.4",
-    "pg-types": "~2.0.0",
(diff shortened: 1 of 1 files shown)


brianc:
Fabuloso! thanks!

brianc:
published `pg@7.12.1`

PR #1946, opened 2019-08-17
Fix disconnection tests for pg-pool 2.0.7

charmander:
In pg-pool 2.0.7, checked-out clients became responsible for their own 'error' events.

brianc/node-pg-pool#123

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -64,9 +64,8 @@ Connection.prototype.connect = function (port, host) {
   })
 
   const reportStreamError = function (error) {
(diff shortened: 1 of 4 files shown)


Gobie:
@brianc it would be great to fix the tests since pg-pool@2.0.7 contained breaking changes.

brianc:
Sorry for missing this - I'm merging this now.  I'm traveling today so I'll do some cleanup and a release either when I get back tomorrow if I'm not too jet-lagged or on Wednesday.

PR #1947, opened 2019-08-17
Add PostgreSQL 10 to CI

charmander:
and change to Ubuntu 18.04 so building OpenSSL isn’t necessary, and move old PostgreSQL version tests to the latest Node LTS.

Depends on #1946.


brianc:
Sorry for not seeing this - if you rebase on master does this build cleanly now? Would be happy to merge when it does!

charmander:
@brianc Not quite, but I’ve fixed it. :)

brianc:
amazing!! Thank you.  Do you have experience w/ postgres & self-signed certs? I'd love to write some tests against our whole client certificate checking snafu from earlier.

charmander:
@brianc Yep, I can try that in a few days.

PR #1967, opened 2019-09-05
fix Check Result._types have getTypeParser method 

mrTrunk:
query pass this.types to conn.parse. Conn.parse wont config.types as array of bind varible type.
query pass this.types to Result object on the next step, for type conversion features, but this one is array.


charmander:
Can you add a test that exposes the problem? It helps make sure it won’t come back, too. [Examples of tests](https://github.com/brianc/node-postgres/tree/master/test/integration/client)

PR #1968, opened 2019-09-06
Add support for describing prepared statements

akheron:
Describing a prepared statement is useful determining inferred input parameter types and result column types for prepared statements.

Usage:
```javascript
client.query({
  text: 'your statement here',
  describe: true
}).then(result => {
  // result.fields contains the output fields as usual
  // result.params (added in this PR) contains the OIDs of input parameters
})
```

When `describe` is given for a query, the query is not bould or executed at all. The only commands sent to Posgres are `Parse`, `Describe` and `Flush`. `Sync` will be sent after the `ParameterDescription` message is obtained .


akheron:
Tests fail because it seems I still have add support to native.

akheron:
Tests fail because it seems I still need to add support to native.

Before doing that, it would be nice to know whether this is a desirable feature or not in the first place :)

brianc:
Sorry for the delay - I like this approach.  I'm cool releasing functionality in the non-native bindings w/o requiring it to be released in the native bindings as well...though it's always nice to have as much parity as possible.  I'm down with merging this & doing a new minor version bump w/ the new functionality later this week.  Mind rebasing on master? The tests should pass now.

akheron:
Rebased on master, but the native tests still seem to fail :thinking: Did you do something that should prevent them from failing?

charmander:
I think a more explicit and restricted API would be better (`Client#describe(…)`), since there’s not enough overlap with querying with regard to what it expects and how it behaves to justify making it an option. Plus it doesn’t get ignored by pg-native/older versions of pg and do something completely different. If it has to be *implemented* as a secret `_describe: true` option on `query`, that’s okay.

charmander:
(It doing something completely different on pg-native is exactly why the tests are failing, too.)

Issue #131, opened 2012-05-30
Array parser drops decimal section of double precision values


makkeu:
Currently, `parseIntegerArray` function is used for both integer and float value parsing.
This isn't working for actual decimal numbers, since parseInt is used:

``` javascript
var parseIntegerArray = function(val) {
  if(!val) return null;
  var p = arrayParser.create(val, function(entry){
    if(entry != null)
      entry = parseInt(entry, 10);

    return entry;
  });

  return p.parse();
};
```

This is related to the pull request #96.

One possible fix could be checking, if the string contains "." and using parseFloat instead:

``` javascript
      if(entry.indexOf(".") > -1) {
        entry = parseFloat(entry);
      } else {
        entry = parseInt(entry, 10);
      }
```

Tho, more elegant way would be to write separate `parseFloatArray` function for types 1021, 1022 and 1231.

PR #1975, opened 2019-09-16
Destroy socket when there was an error on it

Gobie:
When error happens on socket, potentially dead socket is kept open indefinitely by calling "connection.end()".
Similar issue is that it keeps socket open until long-running query is finished even though the connection was ended.

It has to wait for #1946 to fix tests

### DIFF
--- a/packages/pg/lib/client.js
+++ b/packages/pg/lib/client.js
@@ -545,7 +545,7 @@ Client.prototype.query = function (config, values, callback) {
 Client.prototype.end = function (cb) {
   this._ending = true
 
-  if (this.activeQuery) {
+  if (this.activeQuery || !this._queryable) {
     // if we have an active query we need to force a disconnect
     // on the socket - otherwise a hung query could block end forever
     this.connection.stream.destroy()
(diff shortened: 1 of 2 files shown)


brianc:
I'll try to take a look at this tomorrow/wed - sorry for the delay.  It looks reasonable but I need to be at a computer to run some tests with it.

test/integration/connection-pool/error-tests.js
```diff
@@ -98,4 +98,18 @@ suite.test('connection-level errors cause future queries to fail', (cb) => {
       }))
     }))
   }))
+// })
```
charmander: Leftover?

Gobie: yep, fixed

brianc:
one of the tests is still failing here - do you need my help getting it to pass?  You may just need to rebase w/ master.

Gobie:
Hey, after the fix of tests in #1946 it waits for #131. With this fix it works fine.

I also added version for pg-native even though it doesn't have this problem, just to stay consistent.

Gobie:
@brianc as https://github.com/brianc/node-pg-pool/pull/131 was merged I rebased it and it looks ok.

Gobie:
ping @brianc

brianc:
Hey nice! This looks good! Sorry for missing your earlier comment, I was having some email configuration issues up until recently.  Fixed now though. 😄

PR #1987, opened 2019-10-05
Multiple hosts support, try n2

### LINKED ISSUE #1470: PostgreSQL 10 support

zxqfox:
Closes #1470

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,16 +20,16 @@
   "main": "./lib",
   "dependencies": {
     "buffer-writer": "2.0.0",
+    "connection-string": "^3.0.4",
     "packet-reader": "1.0.0",
     "pg-connection-string": "0.1.3",
+    "pg-native": "^3.0.0",
     "pg-pool": "^2.0.4",
     "pg-types": "^2.1.0",
     "pgpass": "1.x",
     "semver": "4.3.2"
   },
   "devDependencies": {
-    "async": "0.9.0",
-    "bluebird": "3.5.2",
     "co": "4.6.0",
     "eslint": "^6.0.1",
     "eslint-config-standard": "^13.0.1",
--- a/lib/connection-parameters.js
+++ b/lib/connection-parameters.js
@@ -8,10 +8,12 @@
  */
 
 var dns = require('dns')
+var async = require('async')
 
 var defaults = require('./defaults')
 
-var parse = require('pg-connection-string').parse // parses a connection string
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -20,16 +20,16 @@
   "main": "./lib",
   "dependencies": {
     "buffer-writer": "2.0.0",
+    "connection-string": "^3.0.4",
     "packet-reader": "1.0.0",
     "pg-connection-string": "0.1.3",
+    "pg-native": "^3.0.0",
```
charmander: pg isn’t supposed to have a dependency on pg-native, since pg-native is optional.

package.json
```diff
@@ -20,16 +20,16 @@
   "main": "./lib",
   "dependencies": {
     "buffer-writer": "2.0.0",
+    "connection-string": "^3.0.4",
```
charmander: connection-string follows a different spec from libpq. I think a PostgreSQL driver should stick with the libpq format.

zxqfox: I know.

zxqfox:
Oh, not sure I can finish this in the near future. Sorry ;-(

Todo:
- [ ] `connection-string` drop (or persuade to make a BREAKING change)
- [ ] JS-version of server lookup with attributes `target_session_attribute=read-write` (analogue to libpq build-in)
- [ ] drop `pg-native` dep

Still not sure what is actually right to do with `connection-string`. Fellows in `pg-connection-string` tells me to replace it with `connection-string` here, and here you saying we need to make it work with non-valid spec
Prob we should drop old behaviour because it disrespect modern URL/connstring specs

Issue #1894, opened 2019-05-10
Ability to force *not* loading native client


gabegorelick:
node-postres supports setting `NODE_PG_FORCE_NATIVE` to force the loading of the native client, but there's no way to force it to *not* load the native client:

```js
if (typeof process.env.NODE_PG_FORCE_NATIVE !== 'undefined') {
  module.exports = new PG(require('./native'))
} else {
  module.exports = new PG(Client)

  // lazy require native module...the native module may not have installed
  module.exports.__defineGetter__('native', function () {
    delete module.exports.native
    var native = null
    try {
      native = new PG(require('./native'))
    } catch (err) {
      if (err.code !== 'MODULE_NOT_FOUND') {
        throw err
      }
      /* eslint-disable no-console */
      console.error(err.message)
      /* eslint-enable no-console */
    }
    module.exports.native = native
    return native
  })
}
```

This can cause spurious warnings about pg-native module not found. See https://github.com/sequelize/sequelize/issues/3781 for an example caused by cloning node-postgres objects.

Potential solution: support a `NODE_PG_NO_LOAD_NATIVE` environment variable. Setting this would cause the `native` getter to not be defined. Setting both `NODE_PG_NO_LOAD_NATIVE` and `NODE_PG_FORCE_NATIVE` could throw an error. 

I'm happy to throw together a PR if there's support for something like this.

seeden:
I see Cannot find module 'pg-native' in my logs. This will help

rasilva-mission:
Does anyone know how to run pg without 'pg-native'? 
`ERROR in ./node_modules/pg/lib/native/client.js
Module not found: Error: Can't resolve 'pg-native' in 'C:\Users\xxxxxx\Documents\projects\xxxxxx\node_modules\pg\lib\native'`

This project compiles electron windows app and I don't want to include the native module just need client.

imjordanm:
I am looking for this same solution. All I need is a way of not forcing pg-native, which I can't use because of its libpq dependency causing build failures. Like OP said, if there was an environment variable to bypass the above lazy require module, that would be a life-saver. The last version that works without pg-native is 3.6.4, but that has it's own issues for me with a missing helper.js file.

charmander:
@imjordanm Last version of what? pg doesn’t require pg-native.

imjordanm:
It does, if you look at the code OP posted there is no way of getting around require("./native") which in turn calls pg-native. 

This is the error code unless you install pg-native:

[...]

PR #1992, opened 2019-10-17
Make `native` non-enumerable

gabegorelick:
Making it non-enumerable means less spurious "Cannot find module" errors in your logs when iterating over `pg` objects.

`Object.defineProperty` has been available since Node 0.12.

See https://github.com/brianc/node-postgres/issues/1894#issuecomment-543300178

### DIFF
--- a/packages/pg/lib/index.js
+++ b/packages/pg/lib/index.js
@@ -44,20 +44,27 @@ if (typeof process.env.NODE_PG_FORCE_NATIVE !== 'undefined') {
   module.exports = new PG(Client)
 
   // lazy require native module...the native module may not have installed
-  module.exports.__defineGetter__('native', function () {
-    delete module.exports.native
-    var native = null
-    try {
-      native = new PG(require('./native'))
-    } catch (err) {
-      if (err.code !== 'MODULE_NOT_FOUND') {
-        throw err
+  Object.defineProperty(module.exports, 'native', {
+    configurable: true,
+    get () {
+      var native = null
+      try {
+        native = new PG(require('./native'))
+      } catch (err) {
+        if (err.code !== 'MODULE_NOT_FOUND') {
+          throw err
+        }
+        /* eslint-disable no-console */
+        console.error(err.message)
+        /* eslint-enable no-console */
       }
-      /* eslint-disable no-console */
-      console.error(err.message)
-      /* eslint-enable no-console */
+
+      // overwrite module.exports.native so that getter is never called again
+      Object.defineProperty(module.exports, 'native', {
+        value: native
+      })
+
+      return native
     }
-    module.exports.native = native
-    return native
   })
 }


gabegorelick:
Test failures are from tests that are flaky on master.

lib/index.js
```diff
@@ -40,20 +40,25 @@ if (typeof process.env.NODE_PG_FORCE_NATIVE !== 'undefined') {
...
-    delete module.exports.native
-    var native = null
-    try {
-      native = new PG(require('./native'))
-    } catch (err) {
-      if (err.code !== 'MODULE_NOT_FOUND') {
-        throw err
+  var native
+  Object.defineProperty(module.exports, 'native', {
+    enumerable: false,
```
gabegorelick: The use of a `get` here implies `writable: false`, and `Object.defineProperty` defaults to `configurable: false`. Any of those could be breaking changes, depending on how clients are interactive with `native` (are they overwriting it?).

lib/index.js
```diff
@@ -40,20 +40,25 @@ if (typeof process.env.NODE_PG_FORCE_NATIVE !== 'undefined') {
   module.exports = new PG(Client)
 
   // lazy require native module...the native module may not have installed
-  module.exports.__defineGetter__('native', function () {
-    delete module.exports.native
```
gabegorelick: This implementation had to change somewhat since `Object.defineProperty` doesn't like deleting a getter. If the goal was to only run the getter once and then cache the result, I think I've reimplement that in a way that's more standard.

charmander:
Cloning or iterating over `pg` is a very strange thing to do to begin with… not that I think this change shouldn’t go in.

gabegorelick:
> Cloning or iterating over pg is a very strange thing to do to begin with

Unintentional cloning is pretty common in ORMs like Sequelize where models have a reference to DB connections. So making a deep copy of a model instance ends up enumerating `pg` instances. See https://github.com/sequelize/sequelize/issues/3781.

But cloning a `pg` instance directly? I agree that is strange. 

I think a more likely use case that this PR could break is consumers overriding or deleting `native`, perhaps to avoid the `console.log`. Unfortunately, this change will cause such use cases to throw an error.

1valdis:
Can we have this merged and released ASAP, please? In sequelize, there are weird errors `cannot find module "pg-native"` happening at seemingly random situations and I couldn't find the root cause...

charmander:
@1valdis That’s probably a bug you or Sequelize should fix, not hide. Deep-cloning clients is a bad idea.

gabegorelick:
Any chance this will get merged?

brianc:
yah we _try_ to be as pedantic as possible about breaking changes and this is _technically_ a non-backwards compatible change.  I agree it's super bizarre to enumerate the pg instance but folks do bizarre things sometimes.

It's also a small thing, but i'd like to see at least a simple unit test that does something like

```
assert.equals(Object.keys(pg).includes('native'), false)
```
or something just to future-proof against this changing.

I also totally agree w/ charmander...if code is 'randomly' throwing an error about pg-native not being installed it's nothing to do w/ this module directly (though making the prop non-enumerable will probably help those situations) because pg doesn't do anything randomly wrt requring native so it's the consuming module's responsibility to fix that random error, not here.

gabegorelick:
> yah we try to be as pedantic as possible about breaking changes and this is technically a non-backwards compatible change. I agree it's super bizarre to enumerate the pg instance but folks do bizarre things sometimes.

@brianc Does that mean this PR should be reworked to be backwards compatible, or that you're fine releasing a new major version of node-pg? I'm not sure there's an easy way to make it backwards compatible. Instead we may have to introduce something like #1894 that allows consumers to opt-in to this change.

brianc:
I think leaving this as non-backwards compatible is fine.  I need to do a semver major bump soon and I'll sweep it into those changes when I do that.  If things pan out well I'll be able to do it this weekend, but if not definitely in the next few weeks.  Holidays + lots of travel make things extra slow...but I appreciate the work and will release it ASAP!

1valdis:
Awesome news. Waiting for the release, hoping that Sequelize won't break with major version change of pg. Is there a (preliminary) list of breaking changes?

brianc:
not yet w/ the list of breaking changes but will be coming soon as soon as
I have a chance to put it together. :)

lib/index.js
charmander: Unconfigurable properties are rarely necessary, and with `enumerable: false` being the default:

```suggestion
    configurable: true,
```

lib/index.js
```diff
@@ -40,20 +40,25 @@ if (typeof process.env.NODE_PG_FORCE_NATIVE !== 'undefined') {
...
-    var native = null
-    try {
-      native = new PG(require('./native'))
-    } catch (err) {
-      if (err.code !== 'MODULE_NOT_FOUND') {
-        throw err
+  var native
```
charmander: To eliminate one subtle introduced difference:

```suggestion
  var native = null
```

lib/index.js
```diff
@@ -40,20 +40,25 @@ if (typeof process.env.NODE_PG_FORCE_NATIVE !== 'undefined') {
...
+    enumerable: false,
+    get () {
+      if (native) {
+        return native
       }
-      /* eslint-disable no-console */
-      console.error(err.message)
-      /* eslint-enable no-console */
+
+      try {
```
charmander: To avoid redoing this (and printing the error) every time it’s accessed, the original approach is good:

```js
Object.defineProperty(module.exports, 'native', {
  value: native,
})
```

(`value` replaces `get`/`set`)

gabegorelick: @charmander Not sure I follow. Wouldn't that immediately load `./native`, rather than lazily load it?

As far as I can tell, the original approach only works because it overwrites `module.exports.native` after the first invocation of the getter [1].

[1] https://github.com/brianc/node-postgres/blob/8f56b8c2fd97119f23215af543eec90bbcee959b/lib/index.js#L56

lib/index.js
gabegorelick: Updated

lib/index.js
gabegorelick: Updated

lib/index.js
charmander: A second `Object.defineProperty` inside the getter, I mean. `{ value }` will get merged with the existing descriptor to produce `{ configurable: true, value: native }`.

charmander: Oh, you already wrote a comment about that: https://github.com/brianc/node-postgres/pull/1992/files#r336177268. (`delete` should work when the property is configurable, but like you said, it’s not necessary.)

gabegorelick: @charmander Updated. Let me know if that's what you had in mind.

gabegorelick:
Rebased to fix merge conflicts, but otherwise no changes were made.

@brianc Let me know if there's anything I can help with to get this merged and released.

brianc:
Will do! I cut my fingertip off on NYE so cant really type (1 hand all
bandaged) - i expect to be back in service on Monday at the latest and will
review and merge in an 8.0 branch shortly! Wanna get 8.0 out asap so I can
start on some major refactoring for 9.0!! Thanks for the PR! :heart:

charmander:
Landed on the 8.0 branch with a test in #2065.

michaelbromley:
> I agree it's super bizarre to enumerate the pg instance but folks do bizarre things sometimes. https://github.com/brianc/node-postgres/pull/1992#issuecomment-552560292

I've just been chasing down this issue and I think in my case it is caused by TypeScript's implementation of ES module imports:

https://github.com/microsoft/tslib/blob/3cf8ae045d949fe7a636d8274d0974a4f30ede8d/tslib.js#L227-L233

```js
    __importStar = function (mod) {
        if (mod && mod.__esModule) return mod;
        var result = {};
        if (mod != null) for (var k in mod) if (Object.hasOwnProperty.call(mod, k)) result[k] = mod[k];
        result["default"] = mod;
        return result;
    };
```

My use-case is using `pg` in a Jest test file written in TypeScript. I look forward to the 8.0 release!

PR #1996, opened 2019-10-22
Support additional tls.connect() options

jgeurts:


jgeurts:
The build error seems unrelated. Please let me know if it is related and there is something for me to fix.

jgeurts:
@brianc any suggestions for how to get this merged?

jgeurts:
@brianc Can yo please help me with what needs to be done to get this merged? This allows for more secure connections to the db server

brianc:
I will take a look tomorrow and let you know

charmander:
Tests need to be fixed (#1946), then this should probably have tests.

How about copying all of the remaining properties from `self.ssl` instead of listing each one, for automatic compatibility with most future additions?

lib/connection.js
```diff
@@ -102,7 +102,20 @@ Connection.prototype.connect = function (port, host) {
...
+      dhparam: self.ssl.dhparam,
+      ecdhCurve: self.ssl.ecdhCurve,
+      honorCipherOrder: self.ssl.honorCipherOrder,
+      privateKeyEngine: self.ssl.privateKeyEngine,
+      privateKeyIdentifier: self.ssl.privateKeyIdentifier,
+      maxVersion: self.ssl.maxVersion,
+      minVersion: self.ssl.minVersion,
+      minDHSize: self.ssl.minDHSize,
       secureOptions: self.ssl.secureOptions,
+      ALPNProtocols: self.ssl.ALPNProtocols,
```
brianc: what if we just did 

```js
tls.connect(Object.assign({
  // all the non-ssl options here
}, self.ssl))
```

jgeurts: Works for me! I'll push an update shortly

jgeurts:
@brianc should be all set. Let me know if there are any changes needed

lib/connection.js
```diff
@@ -91,18 +91,10 @@ Connection.prototype.connect = function (port, host) {
...
-      rejectUnauthorized: self.ssl.rejectUnauthorized,
-      ca: self.ssl.ca,
-      pfx: self.ssl.pfx,
-      key: self.ssl.key,
-      passphrase: self.ssl.passphrase,
-      cert: self.ssl.cert,
-      secureOptions: self.ssl.secureOptions,
-      NPNProtocols: self.ssl.NPNProtocols
-    }
+      checkServerIdentity: tls.checkServerIdentity
```
charmander: This can be taken out – `|| tls.checkServerIdentity` was only needed because it was being explicitly specified in the first place (#1638).

```suggestion
```

brianc:
Awesome! Currently waiting to take off. Will likely be destroyed tired
tomorrow but i have some time on Wednesday booked for this and doing some
more work as well. Thanks again!

teleranek:
> This does change the behaviour for the existing properties being read from `self.ssl` when they aren’t own/enumerable, but it probably won’t affect anyone.

This already broke our server, I think there will be more -_-

charmander:
@teleranek Broke it how? #2009? If so: that it worked was a bug. You can get the previous behaviour back with `rejectUnauthorized: false`, but it’s best to actually make it validate the certificate.

sh-developers:
This broke our server(s) as well, I just forced the version to 7.12.
Not a great update to do in a minor version update, would have preferred breaking change like this to be done in Major versions like semver recommends.

teleranek:
Sorry @charmander have no time to investigate this, but it just suffice to pass only `connectionString` to proper ssl server with legit certificate and leave all options default. Most common way to connect. No matter if that's a bug or not, what matters that the default behaviour was changed in non-major version, not cool.

brianc:
Was an unintended mistake to change default behavior in a minor version. I'm going to fix this within the hour.  Thanks for the patience.

charmander:
> Sorry @charmander have no time to investigate this, but it just suffice to pass only `connectionString` to proper ssl server with legit certificate and leave all options default.

I doubt that it actually has a legit certificate, and that you think it does is why this is a dangerous bug. But yes, it was introduced a long time ago, so I guess a major version is helpful to people.

Issue #1860, opened 2019-03-21
Query streams break with `query_timeout`


twooster:
## Bug:

If a `query_timeout` is reached when using a `submit`-able query object that doesn't have a `callback` method on it, then the Client will throw a `queryCallback is not a function` error. General repro would be:

```javascript
const pg = require('pg')
const QueryStream = require('pg-query-stream')
const db = new pg.Client({ ..., query_timeout: 1000 })
db.connect(err => {
  if (err) throw new Error('Connection error')
  const qs = new QueryStream('SELECT pg_sleep(2)')
  db.query(qs)
    .on('data', d => console.log('data', d))
    .on('end', d => console.log('end', d))
    .on('error', d => (qs.close(), console.log('error', d)))
})
```

## Analysis:

Although the `node-postgres` `Query` object has a `callback` method for handling bulk results, other streaming objects, such as `pg-cursor` and `pg-query-stream`, do not have such a method. Looking at the code in `client.js`, we see: https://github.com/brianc/node-postgres/blob/master/lib/client.js#L431-L457

If the Query object has a `submit` method on it, and no `callback` method, and if no callback is passed in as the second (`values`) argument to `Client#query`, then `query.callback` will be undefined. Subsequently if a `readTimeout` is specified (either as a `query_timeout` on the Query object, or as part of the database configuration), then the read timeout assumes that `query.callback` is defined and attempts to call it, resulting in a `TypeError: queryCallback is not a function`. 

## Fix:

Since the `callback` attribute appears to be an implementation detail of the `node-postgres` `Query` object, and the documentation for `Client#query` does not indicate that this method should be present, or that a second parameter should be passed, it's unclear whether the fix is to make `callback` a required attribute on all `Query`-like objects, or if it's to fail gracefully if such a callback does not exist.

Furthermore, it's debatable whether a `query_timeout` is even appropriate on non-bulk query types (timeout to results ready? timeout until entire result set has been retrieved?), and whether that timeout code should move into the typical bulk `Query` object itself, dodging the issue entirely.

Because that's unclear, I'm leaving it to you to decide. Happy to submit a pull-request with a fix if you can tell me where/what you think the appropriate fix is.

emilioplatzer:
I think that `query_timeout` is useful in other situations. For example when you have a expensive query (like a gruoup by of joins on huge tables or a function that actually calculates some complicated thing). 

With the `query.on('row'` we can send results as they ocurr.

twooster:
[...]

PR #2000, opened 2019-11-05
Ensure query.callback is always valid

adam-nielsen:
Consider an object such as a `pgCursor` being passed to `query()`, and no callback is supplied as the caller is expecting a promise to be returned.  The code will go via `client.js:465` which results in `query.callback` being left undefined.  If the query takes too long and times out, `client.js:490` will try to use the callback to report the timeout, causing an error:
```
    at Timeout.setTimeout (node_modules/pg/lib/client.js:501:7)
    at ontimeout (timers.js:436:11)
    at tryOnTimeout (timers.js:300:5)
    at listOnTimeout (timers.js:263:5)
    at Timer.processTimers (timers.js:223:10)
```
This appears to be the cause of #1860.

This PR fixes the issue by ensuring that the callback is always valid, however I don't fully understand the implications of this change so would appreciate some expert opinions on whether this is the correct solution for this bug.

For one, if the database connection gets terminated, without this PR I immediately get an error:
```
 Error: Connection terminated
    at Connection.con.once (node_modules/pg/lib/client.js:251:9)
    at Object.onceWrapper (events.js:286:20)
```
Followed a few moments later by the above `queryCallback` error.  If I instead use the patch in the PR, the "Connection terminated" error never appears, and I get nothing for a few seconds until the promise fails nicely with the query timeout error.

It would be nicer if the promise failed immediately with the "Connection terminated" error instead, but I'm not sure if that's possible.  It appears that with or without this PR, an error arriving in the `pgCursor` object never makes it back up to reject the promise returned by `query()`.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -487,7 +487,7 @@ Client.prototype.query = function (config, values, callback) {
         query.handleError(error, this.connection)
       })
 
-      queryCallback(error)
+      if (queryCallback) queryCallback(error)
 
       // we already returned an error,
       // just do nothing if query completes


charmander:
`Cursor`s don’t have a `callback` property, so this will replace a cursor return value with a promise that usually doesn’t resolve, which isn’t correct.

The original query timeout change should only have called `query.handleError`, not `query.callback` directly.

adam-nielsen:
Thanks for the tip.  If I make this change in client.js `query()`:
```
-      queryCallback(error)
+      if (queryCallback) queryCallback(error)
```
Then it seems to work better, but I still get an additional error:
```
Error: Connection terminated
    at Connection.con.once (/src/node_modules/pg/lib/client.js:251:9)

pg-cursor handleError: Error: Query read timeout
    at Timeout.setTimeout (/src/node_modules/pg/lib/client.js:494:19)
```
The first one is expected, but the second one arrives somewhat later and appears to be ignored.  I guess it's harmless so I take it there's no way for the `pgCursor` instance to tell `query()` to clear the timeout?

adam-nielsen:
I've been using the fix in this PR for a few weeks now and it's working well.  The tests are only showing failure as they were cancelled and did not finish running.

Any chance of getting this merged?

charmander:
> The original query timeout change should only have called `query.handleError`, not `query.callback` directly.

I think the right fix starts by removing `queryCallback(error)` entirely.

adam-nielsen:
Closing, replaced with #2614

PR #2010, opened 2019-11-20
Revert "Support additional tls.connect() options"

brianc:
Reverts brianc/node-postgres#1996

This is causing backwards compatibility issues.  We don't have an tests in travis covering the expected behavior when connecting to a postgres server over SSL with a self-signed cert.  Until we can get some tests up for that behavior I'm going to revert this to not break all existing folks who do a minor version upgrade.

For v8.0 we'll need to nail this down and make sure it's behaving properly.  As this is the first I've heard of this I haven't dug in yet, but I will figure out what the proper behavior should be, do as much as we can in the next semver minor release (like supporting more ssl options), and make sure it behaves the same way a raw `tls.connect` call is made in node.  It kinda blows my mind that node would treat `self.ssl.rejectUnauthorized: undefined` as `false`...

### DIFF
--- a/lib/connection.js
+++ b/lib/connection.js
@@ -91,9 +91,18 @@ Connection.prototype.connect = function (port, host) {
         return self.emit('error', new Error('There was an error establishing an SSL connection'))
     }
     var tls = require('tls')
-    const options = Object.assign({
-      socket: self.stream
-    }, self.ssl)
+    const options = {
+      socket: self.stream,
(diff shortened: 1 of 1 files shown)


charmander:
Publishing this as a minor version is also a backwards compatibility break, though I suppose almost nobody’ll be using the feature yet. Still, how about:

```js
    const options = Object.assign({
      socket: self.stream,
      rejectUnauthorized: false,
    }, self.ssl)
```

PR #2014, opened 2019-11-28
Monorepo

brianc:
This will likely contain quite a few more commits...just opening it now so I have a quick place to check on travis.  I need to merge the most recent master in as it should help fix some of the travis things.  My bad on branching off a previous commit!


brianc:
I'm not planning on monorepoing everything in this one PR. I just want to get this main repo set up _as if it were multiple packages_ at first. I still need to do a security announcment for the SSL verification and add some deprecation warnings in code for things going away in 8.0.

First thing I want to move into this is `pg-pool` because it's so coupled w/ this repo and even used in test for this repo in places.

sehrope:
I had an idea about maintaining history for the other repos as part of this monorepoization. For the core `pg` module there's nothing special to do (i.e. just move things into `package/`).

For the rest of them how about adding the current state of the other repo (ex: pg-pool) into a branch (ex: local-pg-pool), moving its code to its final `package/` destination, and then having the merge commit into master have two parents commits (ex: master and local-pg-pool. I think that'd allow tracking back history across the file renames / moves.

brianc:
ohh good idea i'll do that!

brianc:
@sehrope I'll probably do each migration in a separate PR starting with the "small" ones (pg-cursor and pg-query-stream), then pg-pool and hopefully the docs too though I'm not suuuuper concerned about them being in their own repo.

sehrope:
Awesome. Yes that will lend itself nicely to separate follow-on PRs too as each would be just that one repo.

charmander:
I was going to suggest `git subtree` for preserving full history – it’s built into Git now.

brianc:
I'll check out `git subtree` today.  I'll do some experiments.  Probably wont add those sub-modules here until over the holiday break next week.

brianc:
@charmander I looked at git subtree. It's nice it preserves full history.  If I'm reading this correctly it seems like it does a merge commit which contains the commit messages of all existing commits in the old repo?  Is that what you had in mind?  I wouldn't want to _continue_ using git subtree and actually maintaining two separate repos, but it looks like a great way to merge an existing repo in and have some tracking of what happened in it.

charmander:
@brianc Yep, that’s exactly what I had in mind! :)

brianc:
sweet!

PR #2021, opened 2019-12-06
Warn when functions intended as constructors are called without new

charmander:
in preparation for #1541. `eval` is a bit awkward, but it’s more accurate than an `instanceof` check and will work on platforms new enough to support pg 8 (i.e. only not Node 4).

### DIFF
--- a/lib/compat/check-constructor.js
+++ b/lib/compat/check-constructor.js
@@ -1,0 +1,22 @@
+'use strict'
+
+const warnDeprecation = require('./warn-deprecation')
+
+// Node 4 doesn’t support new.target.
+let hasNewTarget
+
+try {
+  // eslint-disable-next-line no-eval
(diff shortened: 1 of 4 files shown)


charmander:
I should probably extend this to `Client` and `Query`, huh.

charmander:
Confirmed that `BoundPool` and `Query` are the only two functions that can work without `new` right now.

brianc:
great! You want me to merge this _before_ 8.0 right?

charmander:
@brianc Yep! Thanks for looking at it.

brianc:
totes! Okay gonna merge this & then work on the monorepo over the christmas break. I really want to ship 8.0 in January and start cranking on this again - the end of the year has been non-stop travel and crunch mode at work so looking forward to some FOSS gardening time.

PR #2022, opened 2019-12-06
'query' event at Client

oknesar:
Client fire 'query' event on Client.prototype.query calls


brianc:
I think adding more events is a nice thing to do - let's wait until I get pg-pool put into this repo so we can add consistent 'query' event to the pool and client and document it all at once.  Thanks!

PR #2030, opened 2019-12-18
Add pg cursor

brianc:
This moves the entire history of the [pg-cursor](https://github.com/brianc/node-pg-cursor) into the monorepo structure.  I used `git subtree` to preserve every commit from the prior repo.

@charmander is this what you had in mind w/ preserving history? You can optionally squash the exiting history, but I think it's nice to preserve it.

I had to update a few lint rules and change the formatting and linting on this module to bring it into conformance w/ the main repo lint rules.  Once I get all the repos consolidated here I can do cross-cutting lint changes much more easily.

I haven't yet added `pg` as a dependency to `pg-cursor` and set up learn to manage the dependencies between them.  This will be a semver major change for `pg-cursor` I think as it currently allows you to install it with __any__ version of `pg` which isn't good.  Since `pg-cursor` reaches into the internals of `pg` in some places it'll be nice to lock-step these versions so changes can be easily coordinated between the two.  I haven't researched yet exactly how to manage the versions so I'll research that a bit later.


charmander:
> @charmander is this what you had in mind w/ preserving history? You can optionally squash the exiting history, but I think it's nice to preserve it.

Yep! It should definitely be merged with a merge commit. Looks like this isn’t based off of the latest master, though.

brianc:
> Looks like this isn’t based off of the latest master, though.

latest master of pg-cursor? I think it is...? The most recent commit there is `492fbdb`

charmander:
Latest master of pg, but I was misled by GitHub. Never mind :D

brianc:
uggg...I think I did something wrong - I'm not seeing all the commits in history here after a merge commit.

brianc:
@charmander think I should revert & try again?

charmander:
@brianc They’re there, just sorted by date.

brianc:
ah okay I checked this and only saw two commits, thought I did something wrong...

https://github.com/brianc/node-postgres/commits/master/packages/pg-cursor/index.js

Maybe it's just because the file move stuff got lost due to `git subtree` it considers it a new file.

charmander:
Yeah, that’s exactly it :( I actually thought there would be a rename involved that Git could follow, but it seems like it’s a special kind of merge commit instead. Maybe I’ve done this other times by rewriting the commits of the subtree repo to point to their new destinations… but that loses a different set of information too.

Oh well – the history can be found by following the last commit in that list manually.

brianc:
😱 k thanks!!!

Issue #1993, opened 2019-10-20
Massive performance difference between pg and pg-native


gajus:
I have been trying to get to the bottom of this issue https://github.com/brianc/node-postgres/issues/1952 and along the way I have discovered that the latter issue is not present when using pg-native. However, what I also discovered is that there is massive performance difference between pg and pg-native. Without any change to the workload, and only by changing pg with pg-native, my database CPU usage dropped from an avg. 70% to an avg. 35%.

I suggest that it would be made more clear in the documentation that there is a considerable performance difference between using pg and pg-native. Switching from pg to pg-native will allow us to downgrade our server and save a significant amount of money.

In terms of what changes, as far as I can tell, the biggest difference is how much time is spent in ClientRead state. pg-native compared to pg spends a lot less time in ClientRead state. This appears to imply slow (?) parsing or some other client-side issue handling the response/ sending parameter values.

gajus:
<img width="1920" alt="Screenshot 2019-10-21 at 01 46 28" src="https://user-images.githubusercontent.com/973543/67169673-9f0b1300-f3a4-11e9-9130-fc376e812bad.png">
<img width="1920" alt="Screenshot 2019-10-21 at 01 46 21" src="https://user-images.githubusercontent.com/973543/67169674-9f0b1300-f3a4-11e9-9354-3545997ed097.png">

Here is what a switch from pg-native to pg looks over an hour's use.

vitaly-t:
This would depend a lot on type, size and frequency of data being exchanged. For example, if you are exchanging a lot of binary objects, it may have a significant impact. It would be difficult to generalize for  documentation.

gajus:
90%+ of the queries are simple primary key lookups returning couple of columns, e.g.

```sql
SELECT id FROM country WHERE code_alpha_2 = $1

```

I would think that this is the case for most applications.

There is no binary data being read/ written.

vitaly-t:
And what is the average number of rows being red?

gajus:
Looking at the top queries in the `pg_stat_statements`, most of them return 1 or none rows.

[`maybeOne`](https://github.com/gajus/slonik#maybeone) and [`maybeOneFirst`](https://github.com/gajus/slonik#slonik-query-methods-maybeonefirst)  are the most used query methods.

Are there any real downsides to using pg-native apart from:

[...]

Issue #1952, opened 2019-08-20
ClientRead statement_timeout


gajus:
I have been observing a strange behaviour where PostgreSQL backends created by node-postgres are hanging in ClientRead state.

```
cinema_data_task_executor=> SHOW statement_timeout;
 statement_timeout
-------------------
 10s
(1 row)

cinema_data_task_executor=> SELECT
cinema_data_task_executor->   psa1.backend_start,
cinema_data_task_executor->   psa1.pid,
cinema_data_task_executor->   now() - psa1.query_start duration,
cinema_data_task_executor->   psa1.query,
cinema_data_task_executor->   psa1.state,
cinema_data_task_executor->   psa1.application_name,
cinema_data_task_executor->   psa1.wait_event_type,
cinema_data_task_executor->   psa1.wait_event,
cinema_data_task_executor->   psa1.usename
cinema_data_task_executor-> FROM pg_stat_activity psa1
cinema_data_task_executor-> WHERE
cinema_data_task_executor->   psa1.pid = 29342 AND
cinema_data_task_executor->   psa1.state = 'active'
cinema_data_task_executor-> ORDER BY psa1.query_start ASC;
         backend_start         |  pid  |    duration     |                                                query                                                 | state  |     application_name     | wait_event_type | wait_event |          usename
-------------------------------+-------+-----------------+------------------------------------------------------------------------------------------------------+--------+--------------------------+-----------------+------------+---------------------------
 2019-08-20 20:25:00.282386+00 | 29342 | 00:08:02.920093 | UPDATE event_seating_lookup SET last_seen_at = now() WHERE id = $1 AND ended_at IS NULL RETURNING id | active | do-event-seating-lookups | Client          | ClientRead | cinema_data_task_executor
(1 row)

```

In the above example you can see a backend with `statement_timeout` set to 10 seconds. Meanwhile, the particular query has been running for 8 minutes. From Node.js perspective, the program is just hanging after sending the query.

There appears to be a case where node-postgres is stuck in ClientRead state.

gajus:
To someone knowledgeable about PostgreSQL in general, is there a way to enforce `statement_timeout` or another timeout even when client is in ClientRead state?

adam-nielsen:
[...]

PR #2031, opened 2019-12-19
Some intial performance work

brianc:
I did some experiments with sending flush more regularly & I saw query throughput on a very, very basic benchmark improve by about 10%.  I'm wondering if this has more impact or less impact on situations with more latency as my benchmark was running locally.  I'm particularly trying to shake out what's impacting things in #1993 and #1952 

I also removed some sort of micro-optimization I made a long time ago that isn't helping anymore.  I saw small speed increases, but I think it might help the ClienRead delays in certain cases.  I was buffering up a bunch of messages in the process (parse, bind, execute) before sending any of them.  Now I'm just writing them all to the wire when the come in, and following each w/ a sync to help the backend send it's responses as quickly as possible as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ build/
 node_modules/
 package-lock.json
 *.swp
+dist
+.DS_Store
--- a/packages/pg/bench.js
+++ b/packages/pg/bench.js
@@ -1,0 +1,43 @@
+const pg = require("./lib");
+const pool = new pg.Pool()
+
+const q = {
+  text:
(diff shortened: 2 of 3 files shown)


STRML:
Hey @brianc, we're doing some checking up on how node-pg sends anonymous prepared statements and perf compared to libpq in `binary_parameters` mode. From my read of the code, node-pg actually handles this well because it doesn't do `parse --- wait --- bind/describe/execute/flush`, it just sends them all in one go, as [`libpq` does in this mode](https://github.com/lib/pq/commit/2997d165a566810a3b9e4d16e7cff1300f744c12). Please correct me if I am mis-reading. The main point of interest is compatibility with transaction pooling adapters such as pgbouncer which won't be able to keep state across multiple round-trips.

That said, this PR and #1993 are of interest. Where did this investigation end up after you closed this PR? I see https://github.com/brianc/node-postgres/tree/bmc/make-faster 404s now as well.

PR #2033, opened 2019-12-19
Rewrite parser as transform stream

brianc:
I think there are some performance gains to be had by rewriting the parser.  Also, now that there's a monorepo I can experiment a bit more easily & start integrating typescript into modules outside of the core pg module.

Benchmarks locally for me show the new parser is currently at around 50 queries per second faster in the one really rudimentary benchmark I'm running.  There's a lot more work I need to do here, and this is 100% work in progress, but putting it up just so anyone interested can follow along!


packages/pg-packet-stream/src/BufferReader.ts
```diff
@@ -0,0 +1,44 @@
+const emptyBuffer = Buffer.allocUnsafe(0);
```
joelmukuthu: Is it intentional that the same Buffer instance be shared by all BufferReader instances?

brianc: yeah it's just a placeholder to keep typescript happy & reduce allocations since it's never used.  It gets overwritten with `BufferReader.setBuffer` as soon as the buffer reader starts to be used.  Since you can't resize a buffer, an `allocUnsafe(0)` is fine to share - it's a length of 0 so you can't put anything into it, and you can't resize it.  I'll reduce the noise a bit since it is confusing. Thanks for the feedback!

joelmukuthu: Thank you for the explanation!

PR #2035, opened 2019-12-20
Add pg-query-stream module

brianc:
This uses `git subtree` as awesomely recommended by @charmander to pull in the history of the [`pg-query-stream`](http://github.com/brianc/node-pg-query-stream) into this module. As was done with #2030 I haven't used lerna to manage sub-module dependencies between this & pg-cursor yet.  I'll do that in a follow on PR.

__important__: need to use merge commit button here to merge this, not squash.


packages/pg-packet-stream/dist/BufferReader.js
```diff
@@ -0,0 +1,48 @@
+"use strict";
```
charmander: pg-packet-stream’s `dist` seems to have been added unintentionally?

brianc: ah, boo - thought I put an ignore in for that. will fix, thanks for the catch!

brianc: i rebased that mistake out of the history on this branch.

brianc:
looks like there's a flaky test on node 8 I'll need to take a look at

brianc:
oh wasn't a flaky test it was just running tests from all packages in `parallel` mode isn't gonna work since they use a postgres instance and some of the tests in the main package cancel all running queries and the like.  I was _wondering_ why it was saying "command terminated due to administrative command."  Now I know...

brianc:
Thanks for the review!

packages/pg-query-stream/test/async-iterator.es6
```diff
@@ -0,0 +1,57 @@
...
+  })
+
+  it('can async iterate and then do a query afterwards', async () => {
+    const stream = new QueryStream(queryText, [])
+    const client = new pg.Client()
+    await client.connect()
+    const query = client.query(stream)
+    const iteratorRows = []
+    for await (const row of query) {
+      iteratorRows.push(row)
```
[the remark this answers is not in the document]
matthieusieben: Refer to https://github.com/brianc/node-pg-query-stream/issues/52#issuecomment-561644112 for an alternative

matthieusieben:
You really shouldn't have just copied the code from brianc/node-postgres as it is an invalid implementation of the Readable interface. Using the current implementation breaks the pg connection when used with async iterables.

brianc:
> You really shouldn't have just copied the code from brianc/node-pg-query-stream as it is an invalid implementation of the Readable interface. Using the current implementation breaks the pg connection when used with async iterables.

Ohhh yeah I wanted to keep the operations discrete: first merge the old repo _as is_ just to keep it contained in a single step, then implement your changes, add tests, release a new version, then convert to typescript, etc....versus doing it all in 1 big bang as it's harder to follow that way for me.  Thanks for reminding me though, I'll get to fixing the stream right now!

what do you mean by 
> a test that triggers two breaking loops?

matthieusieben:
```js
for await(const r of new stream) break
for await (const r of new stream) break
```
With each stream running a query that results in more than `batchSize` entries.

(I am on mobile, sorry for the poor example)

brianc:
no worries! appreciate the help! its my wifes birthday today so prolly wont
have time to finish this until tomorrow but i put up a PR and i think i
covered that case with a test!

PR #2049, opened 2019-12-27
Adding ability to pass through idle_in_transaction_session_tim…

aheuermann:
…eout

Allowing for passing in the `idle_in_transaction_session_timeout` option.

### DIFF
--- a/packages/pg/lib/defaults.js
+++ b/packages/pg/lib/defaults.js
@@ -59,6 +59,10 @@ module.exports = {
   // false=unlimited
   statement_timeout: false,
 
+  // Terminate any session with an open transaction that has been idle for longer than the specified duration in milliseconds
+  // false=unlimited
+  idle_in_transaction_session_timeout: false,
+
   // max milliseconds to wait for query to complete (client side)
   query_timeout: false,
 
--- a/packages/pg/lib/client.js
+++ b/packages/pg/lib/client.js
@@ -371,6 +371,9 @@ Client.prototype.getStartupConf = function () {
   if (params.statement_timeout) {
     data.statement_timeout = String(parseInt(params.statement_timeout, 10))
   }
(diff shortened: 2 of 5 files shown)


packages/pg/lib/defaults.js
```diff
@@ -59,6 +59,10 @@ module.exports = {
   // false=unlimited
   statement_timeout: false,
 
+  // Terminate any session with an open transaction that has been idle for longer than the specified duration in milliseconds
+  // false=unlimited
+  idle_in_transaction_session_timeout: false
```
brianc: you need a comma at the end of the line here

aheuermann: Doh, thanks! Was having trouble getting the tests running locally so relied on running on travis, thanks for the quick look!

brianc: yeah no problemo!  The tests are pretty antiquated since a lot were written before things like mocha and jest existed. If you have vscode & docker installed you can install the `remote - containers` extension & it'll boot you into an isolated environment where you can develop & test....hopefully that helps!  Using travis to run tests is also fine with me!

aheuermann:
Note: it looks like `idle_in_transaction_session_timeout` at the client-level was added in v10, is there a simple way to disable the tests for all versions < v10?

brianc:
no rush here - code changes look good...I'll merge it once the tests pass. 👍

brianc:
published with `pg@7.17.0` - thanks!

pSnehanshu:
This is confusing, there's also a `idle_in_transaction_session_timeout` parameter available in Postgres. Does it override that value at the connection level, or is this handled at `pg` level?

charmander:
> there's also a `idle_in_transaction_session_timeout` parameter available in Postgres

@pSnehanshu That is exactly what this option sets, and nothing else.

PR #2050, opened 2019-12-28
[pg-stream] Async iterable

matthieusieben:
- Does not emit `error` [events](https://nodejs.org/api/stream.html#stream_errors_while_reading)
- Properly destroy the stream in case of error while reading
- Does not emit `close` to aclively close. `close` avents should be used to notify that the stream *was* closed, and not that the stream must be closed.
- No need to keep a custom `this.batchSize`. The internal Readable machinery relies on a `highWaterMark` that does exactly that when reading more rows.
- No need to make an additional `cursor.read()` when we received fewer rows than requested (the cursor will always return exactly the requested number of rows, unless there are no more rows to be read from the query).

### DIFF
--- a/packages/pg-query-stream/index.js
+++ b/packages/pg-query-stream/index.js
@@ -1,14 +1,16 @@
 'use strict'
-var Cursor = require('pg-cursor')
-var Readable = require('stream').Readable
+
+const { Readable } = require('stream')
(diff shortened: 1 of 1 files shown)


matthieusieben:
Fixed by #2051

Issue #2058, opened 2020-01-09
pg-query-stream runs VERY slow


gabrieldodan:
Hi,

I have noticed that ``pg-query-stream`` runs very slow compared to ``pg`` , about 10x slower ...Tested on

```
pg 7.17.0
pg-query-stream 2.1.2
node 12.14.0
postgresql 10
```

The old versions of ``pg`` and ``pg-query-stream`` works fine , ``pg 6.1.2`` and ``pg-query-stream 1.0.0``

Any idea?

Thanks!
Gabriel

charmander:
How about the latest pg with an older pg-query-stream (2.0.0 or… 1.1.2?)? pg-query-stream had a fair number of changes made to it recently for correctness, although I’m not sure exactly how the versions worked out in the end.

gabrieldodan:
Better, tested with pg-query-stream 2.0.0   Still a performance difference but not that huge

brianc:
I'm going to release a new major version of pg-query-stream in about 1 hour.  The internals are substantially re-written.  Interested in if you see more performance improvements!  Also, I plan on working hard and long on performance for pg@9.0.  This module has collected some cruft and kept old, strange API design decisions around for a long time in the name of backwards compatibility or me not having enough time to get to things.  That's changing for 2020.  I had plans to already start on a bunch of this work, but cut the end of my finger off on NYE...now that I can (mostly) type okay again the work has resumed!  Stay tuned...just writing changelog entry for pg-query-stream@3.x and will push it.

mike-marcacci:
That sounds like quite a new year! Best of luck!

brianc:
my work day got a little out of control & I didn't get a chance to release - wanted to wait for a review on it & now its a bit late...would rather release in the AM when I get up so I'll do this w/ my morning coffee tomorrow.  Sorry for the delay!

brianc:
hmmm sorry for not following up here sooner...I released pg-query-stream@3.0 a while ago...hopefully the performance issues are resolved for you now!  If they're not, please feel free to repoen w/ more info.

Issue #2056, opened 2020-01-06
pg.Result.fields isn't always an array


songsanli:
After [this commit](https://github.com/brianc/node-postgres/commit/e5d46749c0e41498da8451622dd3ed2218f717bd#diff-16be60eca5ce2d08e46c4822f4f56606R20), `pg.Result.fields` will be an array only after `pg.Result.addFields()` is called.

However, `pg.Result.addFields()` is called only in the callback of `RowDescription` event, which is dispatched by pg in response to a SELECT, FETCH, etc query. If I send sqls like `create SCHEMA hello`, the `pg.Result.fields` will remain undefined. This behavior is different from [the doc](https://node-postgres.com/api/result):

> Every result will have a fields array.

Please heve a look. Thanks.

brianc:
ahh dang yeah I need to fix that, sorry about the hassle there.  I cut the end of my finger off on NYE  🤕 and juuust about healed up but haven't realy been able to type effectively up until yesterday...I'll get a change for backwards compatibility (+ a test for this) released ASAP (1-2 days 🤞 )

PR #2059, opened 2020-01-09
pg-query-stream@3.0 release

brianc:
As mentioned in #2058 and #2051...I delayed releasing this until I could type again after 🔪 my finger on new years eve 🎆 😢 .  I noticed when typing up the changelog that I unintentionally entirely removed the ability to control the pg-query-stream's highWaterMark through the "proper" node readablestream config value so I added that back in with tests.  

Now that I can (mostly) type okay....let's get this out!  Immediately following this release today I'm going to put up a patch to fix a regression mentioned in #2056.


CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ For richer information consult the commit log on github with referenced pull req
 
 We do not include break-fix version release in this file.
 
+### pg-query-stream@3.0.0
+
+- [Rewrote stream internals](https://github.com/brianc/node-postgres/pull/2051) to better conform to node stream semantics.  This should make pg-query-stream much better at respecting [highWaterMark](https://nodejs.org/api/stream.html#stream_new_stream_readable_options) and getting rid of some edge case bugs when using pg-query-stream as an async iterator.  Due to the size and nature of this change (effectively a full re-write) it's safest to bump the semver major here, though almost all tests remain untouched and still passing, which brings us to a breaking change to the API....
+- Changed `stream.close` to `stream.destroy` which is the [official](https://nodejs.org/api/stream.html#stream_readable_destroy_error) way to terminate a readable stream.  This is a __breaking change__ if you rely on the `stream.close` method on pg-query-stream...though should be just a find/replace type operation to upgrade as the semantics remain very similar (not exactly the same, since internals are rewritte, but more in line with how streams are "supposed" to behave).
```
charmander: Minor typo fix:

```suggestion
- Changed `stream.close` to `stream.destroy` which is the [official](https://nodejs.org/api/stream.html#stream_readable_destroy_error) way to terminate a readable stream.  This is a __breaking change__ if you rely on the `stream.close` method on pg-query-stream...though should be just a find/replace type operation to upgrade as the semantics remain very similar (not exactly the same, since internals are rewritten, but more in line with how streams are "supposed" to behave).
```

brianc:
Thanks for the reviews my friend! ❤️ Hope your new years was bright and all yer fingers stayed in 1 piece!

PR #2062, opened 2020-01-10
Drop support for EOL versions of node

brianc:
Starting to work on the 8.0 release now!  First things first: drop support for EOL versions of node.  This particular PR doesn't do much outside of removing node 8 from the test matrix, and removing a few places where we check the version of node and branch for earlier versions.  More importantly, this will allow us to start using newer features like `async/await` internally which is big!

I'm targeting all pull requests into the `bmc/8.0` branch.  Once that branch looks good we can release it, and then I'll get to work on 9.0 stuff which is a lot more significant internal changes in the name of performance & maintainability.


vitaly-t:
As per my comments from #2068...

> The reality is such, many companies are stuck with older versions of Node.js in production, and dropping support for the sake of chasing Node.js release cycle will lock out a huge user base, without good reason. Therefore, I would advise against that, and not to drop older Node.js unless there is a good reason for it.

> If you want to bring in `async`/`await`, it went official since Node.js 7.6.0, which is what I am supporting in pg-promise as the minimum version. And there wasn't any big change in Node.js that would be worth dropping older Node.js for.

> I believe that supporting at least Node.js v8 is very much worthwhile at this point.

brianc:
Continuing discussion from #2062 - should we continue to support node@8.x?  It's a few years old and end of life so not supported anymore by node.js itself...however, I don't want to lock out people on old versions of node from upgrades without a reason.  Node prior to 8 has definitely caused issues so that's going bye bye from the support matrix...but if all the tests still pass on 8...maybe we set the minimum version to 8 unless there's a compelling reason to drop it?  I think one of the ideas for dropping 8 was `BigInt` support but I'm not putting that in the 8.x branch of pg by default so....

vitaly-t:
>  I think one of the ideas for dropping 8 was `BigInt` support

`BigInt` started working only since Mode.js v10.4.0, which i think is too new at this point to make it a required minimum.

brianc:
> `BigInt` started working only since Mode.js v10.4.0, which i think is too new at this point to make it a required minimum.


Okay yeah...thanks for weighing in on this @vitaly-t - I think you've convinced me to keep supporting v8.x of node until there's a compelling reason not to but I'll wait over the weekend to let others weigh in as well.  ❤️  you're right it's a lot more common to run EOL versions of node in prod than it can seem when you're working at a bleeding edge startup or just on side projects using the newest version of everything.

vitaly-t:
The reason many companies are stuck with older Node.js is having to use a large number of packages that do not yet support newer Node.js versions, and you can use only one version of Node.js at a time, so it becomes a necessary compromise to carry on with an older Node.js

joelmukuthu:
I think the maintenance burden is a good enough reason to drop support for older Node.js versions, especially when a project is open-sourced. The more legacy patterns remain in this project the harder it is to maintain, not only for the core maintainers but also for community members wishing to submit PRs.

When a maintainer of an open source project is willing to keep working on a project, I think it's in our best interests to support that maintainer however we can. If, as in this case, dropping support for older Node.js versions makes maintenance easier for them then it is what it is. The benefits for new users are increased (with new features and "newer", perhaps easier to understand code), while for users on older versions, the code will still keep working.

That said, important bug fixes can still be backported to the older versions, which is something that can also be pioneered by the consumers of the older versions due to the open-source nature of it all.

vitaly-t:
> When a maintainer of an open source project is willing to keep working on a project, I think it's in our best interests to support that maintainer however we can. If, as in this case, dropping support for older Node.js versions makes maintenance easier for them then it is what it is

You confuse working for community with writing some code for your own amusement. Countless businesses depend on this library. It will either have support done right, or it will be abandoned and forgotten.

brianc:
Yeah im fine keeping support for node 8. At the end of the day _any_ amount
of maintenance can be viewed as a burden __or__ a labor of love. I choose
to see it as a joy and privilege to get to maintain this. So...I am fine
with keeping 8 support for now. I plan on transitioning to typescript for
pg@9 and that will help smooth over any JS inconsistency between node
versions too.

I will say to the countless businesses relying on this module...you should
sponsor me!

https://github.com/sponsors/brianc

:heart:

Ill change this PR to support node>=8 by adding it back to the test matrix
later today and then merge it!

brianc:
Okey dokey re-added node8.x to the test matrix....so it'll still be supported for maybe another year or so or until it becomes a blocker on shipping some new feature? We'll figure it out some time in the future!  🚀

charmander:
(There are still a few changes to .travis.yml that weren’t reverted)

brianc:
> (There are still a few changes to .travis.yml that weren’t reverted)

yeah I just left one of the carbon runs out - its testing against a diff minor version of postgres but we test against that version w/ another version of node...so I figured it was a bit redundant.  I can add it back if you think it's important...but in reality there are so few changes in minor versions of postgres anyway its prolly overkill.

charmander:
What that one exists to test, IIRC, is Ubuntu 12.04 with the most recent Postgres it supports. (The other `dist: precise` is to test PostgreSQL 9.2, which more recent versions of Ubuntu don’t support.)

brianc:
cool - i put it back!

packages/pg-pool/package.json
```diff
@@ -34,6 +34,6 @@
     "pg-cursor": "^1.3.0"
   },
   "peerDependencies": {
-    "pg": ">5.0"
+    "pg": ">8.0"
```
charmander: `semver.validRange` confirms that this means “at least 8.1.0”.

```suggestion
    "pg": ">=8.0"
```

karlbecker:
> I will say to the countless businesses relying on this module...you should
sponsor me!

+100 on this 😊 

If your business gets some value out of this, return the favor and provide some vale to @brianc ! 

Also, just adding a +1 that node 8 support is still useful for our team, so I'm glad you kept it around! We're in the process of transitioning up to node 10 or 12 on a few repos, but not there quite yet.

brianc:
Thanks @karlbecker ❤️ 

Glad to hear leaving node 8 support was the way to go!  It'll be supported in `pg@9.0` too.  I want to do minimal breaking changes in v9.0, but I will be gutting the internals in the name of performance!

PR #2063, opened 2020-01-10
Make `BoundPool` an es6 subclass of `Pool`

brianc:
This fixes subtle subclassing bugs using es6 classes.  It is a semver breaking change targeting 8.0 release...though it's not likely to cause issues in the general case.

I'm merging this into the `bmc/8.0` branch.  This closes #1541 as it had merge conflicts since it got quite old & doesn't have tests...but I still appreciate the thought and work that went into it!

### DIFF
--- a/packages/pg/lib/index.js
+++ b/packages/pg/lib/index.js
@@ -7,35 +7,30 @@
  * README.md file in the root directory of this source tree.
  */
 
-var util = require('util')
-var Client = require('./client')
-var defaults = require('./defaults')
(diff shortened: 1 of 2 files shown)


charmander:
#1541 was editable, so I brought it up to date with these tests and a smaller diff if you want ;)

Then the other post-ES5 cleanup can go in on top of it so each change is more self-contained.

brianc:
alright i'll close this one & merge that one

PR #2065, opened 2020-01-10
Make `native` non-enumerable

brianc:
This is as superset of #1992 which includes a test.  Targeted at `8.0` release branch.  The impact of this change is minimal, but it's technically breaking.

Closes #1992


brianc:
need to merge #2062 to fix travis

Issue #1568, opened 2018-02-08
Security issue: DB passwords possibly leaking to logs


igabesz:
**Version:** `"pg": "7.1.2"`

We've seen such lines in our logs (uncaught exception handler). 

    {"err":{
    "message":"Connection terminated unexpectedly",
    "stack":"Error: Connection terminated unexpectedly\n    at Connection.con.once (PATH/node_modules/pg/lib/client.js:182:21)\n    at Object.onceWrapper (events.js:313:30)\n    at emitNone (events.js:106:13)\n    at Connection.emit (events.js:208:7)\n    at Socket.<anonymous> (PATH/node_modules/pg/lib/connection.js:75:10)\n    at emitOne (events.js:116:13)\n    at Socket.emit (events.js:211:7)\n    at TCP._handle.close [as _onclose] (net.js:554:12)",
    "client":{
      "domain":null,
      "_events":{},
      "_eventsCount":1,
      "connectionParameters":{
        "user":"USER",
        "database":"DATABASE",
        "port":PORT,
        "host":"HOST",
        "password":"PASSWORD!!!111!!one!!"
        ...

Well. I hope I don't have to explain. 
Also it's kinda urgent for us.

charmander:
The `connectionParameters` property is only necessary for `.cancel()`, and that doesn’t work right now (#1392). Maybe it can be removed. However…

> Also it's kinda urgent for us.

Configure your error logging not to include the `password` property.

igabesz:
I highly recommend to reconsider this error management because right now I would not recommend this package for any serious industrial use-cases. I explain it in details. 

You can *never* know how your exceptions will be handled. The try-catch block is not necessarily near the database operations. See this actual example: something broke and the error was caught by the uncaught error handler. (Which logged the error and then killed the whole process.) In this very case we can (and will) check the exception in the uncaught error handler if it contains `client.connnectionParameters.password` property before logging. But shall we do this in each and every try-catch blocks just to make sure? What if a 3rd party lib has a try-catch and some logging as well? Shall I intercept the logged texts and search for `password` strings inside?

You don't know how and where your exceptions will be handled. And vice versa: one should not be afraid in their `catch` to log the exceptions. Furthermore: One should not be required to know if logging an error is safe or not. 

[...]

PR #2066, opened 2020-01-10
Remove password from stringified outputs

### LINKED ISSUE #2064: Ensure password is non-enumerable on client configuration
No amount of `console.log(client)` or `console.log(JSON.stringify(client))`  should print the password anywhere.

brianc:
There's a security concern where if you're not careful and you include your client or pool instance in console.log or stack traces it might include the database password.  To widen the pit of success I'm making that field non-enumerable.  You can still get at it...it just wont show up "by accident" when you're logging things now.

The backwards compatibility impact of this is very small, but it is still technically somewhat an API change so...8.0.

closes #2064 #1568


brianc:
Yeah...the formatting is kinda jacked as pg-pool, pg-cursor, and pg all had different lint rules...so my vscode autoformatting went a bit crazy.  I temporarily turned it off for this repo.  I want to do a single big "prettier + eslint reformat everything all at once" commit & then clamp down on 1 set of lint rules + prettier for the whole repo, but waiting until after 8.0 just so if anyone wants to bisect what changed between the releases there's less noise in the diff.

brianc:
> rebase-merged onto this one.

sorry i missed that - i did a sqash merge (was planning on doing a squash merge of this into bmc/8.0)

PR #2070, opened 2020-01-13
Remove password from stringified outputs

charmander:


packages/pg/lib/client.js
```diff
@@ -33,12 +33,11 @@ var Client = function (config) {
 
   // "hiding" the password so it doesn't show up in stack traces
   // or if the client is console.logged
-  const password = this.connectionParameters.password
   Object.defineProperty(this, 'password', {
+    configurable: true,
```
brianc: i didn't really understand the advantage or diff between configurable or not 🤷‍♂

charmander: Configurable is pretty much always the right option. It lets people monkeypatch, hook, freeze, and do all the other JavaScripty stuff when they want to, and it’s the default for most things built into the language.

Except `Object.defineProperty`. 🙃

charmander:
I just realized that 410b89b22bd72cbdb71db5de8c990d16a2cf6c04 fixes the problem for `Pool` when used from pg-pool, but not when used from pg (because its `BoundPool` extends the pool’s options with an `Object.assign`). Should the same fix go in there to support `new pg.Pool(otherPool.options)`, or does that seem like too much of a hack to avoid swapping out `Object.assign` with something that doesn’t care about whether properties are inherited or enumerable? @brianc

brianc:
I'm not following what you're saying exactly? If you do 

```js
const newPool = new pg.Pool(otherPool.options)
console.log(JSON.stringify(newPool))
```
the password shows up again? Dang...yeah I mean ideally there's no way to __accidentally__ log out a password...regardless of how objects are constructed.

charmander:
@brianc The password doesn’t show up, but that pattern will stop working, because the `Object.assign` that copies `otherPool.options` doesn’t include the `password` property.

brianc:
oh i see what you're saying - so we could special case it to be copied in if you do that?  ugg....hmmm I mean we could special case copying password over, mutate the `options` argument and append the `Client` directly to it, make the `pg.Pool` constructor take a second (undocumented) argument for the Client constructor to use (gross).  We could also do something like...

```
class BoundPool {
 constructor (options) { 
    super(options)
    this.options.Client = Client
  }
}
```

That's actually pretty clean IMO and isn't even more code or anything.  Should fix the problem too, right?

charmander:
@brianc Ah, nice solution! That would be perfect, except there’s this in pg-pool:

https://github.com/brianc/node-postgres/blob/ee8d32f97cd5e5907c8cd9d815c7fe57a7031f7f/packages/pg-pool/index.js#L65

which makes it do `require('pg')`, which is probably fine because of the peer dependency but still a little weird. However, I think `|| Client` might achieve exactly what’s needed:

```js
super(options, Client)
```

which also defaults to the `Client` being passed into `BoundPool` if one exists, instead of overwriting it. \o/

PR #2071, opened 2020-01-13
Fix tests skipped because of missing suffixes

charmander:
Mocha will happen eventually!


brianc:
> Mocha will happen eventually!

indeed it will!

.travis.yml
```diff
@@ -17,6 +17,17 @@ addons:
 
 matrix:
   include:
+    - node_js: lts/erbium
```
brianc: nit: would be nice to put a comment here just to explain why we're doing this for our future selves

brianc:
This looks good! Will merge once tests pass.

brianc:
beautiful. and the tests still pass! 😅

PR #2074, opened 2020-01-14
Update package.json

dhritzkiv:
Change the homepage URL's scheme to https

Tools like VSCode warn me about visiting http://github.com when accessing this repo from it


charmander:
Thanks!

Issue #2072, opened 2020-01-13
Issue with pg-query-stream dest.on is not a function


nhumrich:
Using the example code in pg-query-stream, you get an error that dest.on is not a function. 

Here is my code:
```node
const pg = require('pg')
const QueryStream = require('pg-query-stream')

let p = new pg.Pool({host: 'localhost', user: 'postgres', database: 'postgres'})
p.connect((err, client, done) => {
   if (err) throw err;
   const query = new QueryStream('SELECT * FROM generate_series(0, $1) num', [1000000])
   const stream = client.query(query)
   //release the client when the stream is finished
   stream.on('end', done)
   stream.pipe(console.log)
}
```

and I get the following error:
```
Uncaught TypeError: dest.on is not a function
    at PgQueryStream.Readable.pipe (_stream_readable.js:714:8)
    at PendingItem.callback (repl:7:10)
    at Pool._acquireClient (/home/nhumrich/project/node_modules/pg-pool/index.js:267:21)
    at /home/nhumrich/project/node_modules/pg-pool/index.js:230:21
    at Connection.<anonymous> (/home/nhumrich/project/node_modules/pg/lib/client.js:234:7)
    at Object.onceWrapper (events.js:411:26)
    at Connection.emit (events.js:316:22)
    at Connection.EventEmitter.emit (domain.js:545:15)
```

pg-query-stream version 3.0.0

brianc:
ah that should be an easy fix - I can take a look at it tomorrow.  In the mean time, this should work:

```
const query = new QueryStream('bla bla')
client.query(query)
query.on('end', done)
```

basicaly `pg.Client.query` isn't returning the thing you pass it...that function is incredibly overloaded and I might have missed this case.

brianc:
got a bit stuck at work today - will look @ this tomorrow!

brianc:
I am looking at this...I don't think it's very straight forward to support things other than query configs from `pool.query` - `pool.query` is a helper method that checks out a client, dispatches a query, and then releases the client.  If there's an error checking out a client `pool.query` needs to reject...so for the time being it needs to return a promise.  I'll update the documentation to reflect that if you need to dispatch a stream you need to check the client out and dispatch it.  There's some work I can do to make this work in a non-hacky way...(e.g. the stream could emit an error if you can't connect) but it requires solidifying the `Submittable` concept a bit more thats on the docket for 9.0 work.  For now I'm going to update the documentation and mark this as a feature request.

[...]

Issue #2009, opened 2019-11-19
self signed certificate error


eric-elsholz:
Hi. Sorry ahead of time, I don't have a lot of detail here. We updated our packages today and afterwards started seeing a connection error from Sequelize ORM: "self signed certificate". We do in fact use a self signed certificate but this has never been a problem for us in the past. We downgraded from pg@7.13.0 to pg@7.4.3 and then we were able to connect again.

ianwremmel:
If I’m reading https://github.com/brianc/node-postgres/commit/bf029c827049ca16add0a862d40f4e60dfd9e602 correctly, I think there may have been a bunch of options (including rejectUnauthorized) being inadvertently set to falsy values. They are now not set at all unless you pass them in. The last release was a breaking change, but it’s probably easiest to fix by passing rejectUnauthorized: false to pg.

ianwremmel:
Sorry, I think I actually meant checkServerIdentity.

charmander:
> I think there may have been a bunch of options (including rejectUnauthorized) being inadvertently set to falsy values

`tls.connect` treats `rejectUnauthorized: undefined` as `rejectUnauthorized: false`? … It does. Oh no. :(

Well, it’s a good thing that was fixed. We should probably add a test for it. It might also be worth a security announcement (@brianc).

powerc9000:
God I thought I was going crazy! Thanks for this info rolling back to 7.12 worked for me.

eldimious:
I am facing the same issue. Rolling back to `7.12.1` worked for me too.

znarf:
Spotted that in our staging system (heroku), happy to give more details.

BrandonZacharie:
I'm using Sequelize as well and I got the "self signed certificate in chain" error until I rolled back to 7.12.1

Kinetic-Pulse:
This one one of the most painful bugs I had to identify.
Probably shouldnt do too many updates at once but ye....
Downgrading is the way for now ~7.12.1

ptpt:
Same here. Very hard to identify. The only error message we got is:
```
Error: self signed certificate in certificate chain
    at TLSSocket.onConnectSecure (_tls_wrap.js:1321:34)
    at TLSSocket.emit (events.js:210:5)
    at TLSSocket.EventEmitter.emit (domain.js:476:20)
    at TLSSocket._finishInit (_tls_wrap.js:794:8)
[...]

PR #2075, opened 2020-01-15
Deprecate implicit TLS `rejectUnauthorized: false`

charmander:
Yes, it treats `undefined` as `false`. Discussion in #2009. Introduced unintentionally in pg 0.8.7.


brianc:
Awesome!

brianc:
this one goes in before 8.0, yeah? the work to throw instead of warn happens in 8.0, yeah?

charmander:
@brianc Yep!

brianc:
sounds good. Will get this out today or tomorrow!  wanna get a fix #2072 up and out too.

PR #2076, opened 2020-01-15
Continue support for creating a pg.Pool from another instance’s options

charmander:
by dropping the requirement for the `password` property to be enumerable.


charmander:
[proof of correctly failing test][1]


[1]: https://travis-ci.org/brianc/node-postgres/builds/637666188

brianc:
nice! Perfection. ❤️

Issue #2047, opened 2019-12-27
Anything else to add?


brianc:
@charmander thanks for putting the 8.0 milestone on a few of the issues.  Anything else that comes to mind you want to break in the newest release?  I want to keep churn _somewhat_ down on the 8.0 release...one of the main things I want to do is drop support for all versions of node earlier than any active LTS versions.  There was some talk about making `BigInt` support opt-in in 8.0...but since it'll be opt-in for now I don't think we need to land it directly on the major version bump.

It looks like 8.0 is going end of life in like...3 days!  So I think we could drop official support for anything earlier than node 10 which is the currently oldest LTS version...what do you think? [reference](https://nodejs.org/en/about/releases/)

Now that this is a monorepo with most of the main dependent modules included I can have a much easier time doing sweeping changes to the internal APIs and weird timing behaviors.  I want to hit the entire monorepo w/ a prettier format & enforce prettier formatting via eslint because the lint rules haven't been consistent between all the repos...and the tests inside `pg` (some of which are very old) haven't been linted at all...so...I'm going to fix all the lint stuff before 8.0 just to keep things slightly more organized. 

Thanks again for all your help & support.

uasan:
Remove callback interfaces, save only promise interfaces.
This clearing your code base and use async native JS functions.

vitaly-t:
I just wanted to add my main concern here...

There are literally thousands of examples and manuals out there, explaining how to use such external modules as `pg-query-stream`. And merging the module in has just invalidated all that. I predict the confusion over using it as external module will echo here accordingly, 1000-fold.

vitaly-t:
@brianc As to what to add, I will again sneak in with my [connection-string](https://github.com/vitaly-t/connection-string) module :smile:

brianc:
> merging the module in has just invalidated all that

[...]

PR #2081, opened 2020-01-23
Do not return broken or closed clients to the pool

brianc:
Building off of https://github.com/brianc/node-pg-pool/pull/119

This was slated to go into 8.0, but after working on this I'm not sure it's a breaking change...previously if you released a broken or ended client you'd basically put your app into a state where you'd crash or hit undefined behavior in the future.  Perhaps we just merge this into the existing 7.x branch so more folks can get the change sooner.  what you think @charmander?

Either way I think this closes the final thing mentioned in #2047 

The only thing really left for 8.0 is to get some proper tests in place for handling self-signed versus "real" certificates so we can check the code is doing the right thing in travis in all cases.  I need to reference the jdbc postgres repo and find how the configure all those different situations to test against them.  After that, I can write tests, make sure it works, and release 8.0.  Then the fun work of 9.0 begins. 🏃

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ env:
 node_js:
   - lts/dubnium
   - lts/erbium
-  - 13
+  - 13.6
 
 addons:
   postgresql: "10"
--- a/packages/pg-pool/index.js
+++ b/packages/pg-pool/index.js
@@ -25,6 +25,31 @@ class PendingItem {
   }
 }
 
+function throwOnRelease () {
+  throw new Error('Release called on client which has already been released to the pool.')
+}
+
+function release (client, err) {
+  client.release = throwOnRelease
+  if (err || this.ending || !client._queryable || client._ending) {
+    this._remove(client)
+    this._pulseQueue()
+    return
+  }
+
+  // idle timeout
+  let tid
+  if (this.options.idleTimeoutMillis) {
+    tid = setTimeout(() => {
+      this.log('remove idle client')
+      this._remove(client)
+    }, this.options.idleTimeoutMillis)
+  }
+
+  this._idle.push(new IdleItem(client, tid))
+  this._pulseQueue()
+}
+
 function promisify (Promise, callback) {
   if (callback) {
     return { callback: callback, result: undefined }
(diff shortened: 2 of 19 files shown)


.travis.yml
```diff
@@ -10,7 +10,7 @@ env:
 node_js:
   - lts/dubnium
   - lts/erbium
-  - 13
+  - 13.6
```
charmander: What’s this for?

packages/pg-pool/index.js
```diff
@@ -256,7 +260,7 @@ class Pool extends EventEmitter {
 
     client.release = (err) => {
       if (released) {
-        throw new Error('Release called on client which has already been released to the pool.')
+        throwOnDoubleRelease()
```
charmander: "

packages/pg-pool/test/releasing-clients.js
```diff
@@ -0,0 +1,54 @@
...
+    // make a pool w/ only 1 client
+    const pool = new Pool({ max: 1 })
+    expect(pool.totalCount).to.eql(0)
+    const client = await pool.connect()
+    expect(pool.totalCount).to.eql(1)
+    expect(pool.idleCount).to.eql(0)
+    // end the client gracefully (but you shouldn't do this with pooled clients)
+    client.end()
+
+    // try to return it to the bool
```
charmander:
```suggestion
    // try to return it to the pool
```

.travis.yml
brianc: the test to iterate a stream multiple times w/ async iterator stopped working in node 13.7....I've had occasional breaks before on single semver minor versions of non LTS node...if this doesn't resolve itself w/ 13.8 I'll have to track it down & see if it was something they intentionally changed w/ async iterators on streams.

brianc: btw...found the issue:

https://github.com/nodejs/node/pull/31508

brianc:
image


having a bit of a mental breakdown trying to rebuild the history in a way that cleanly merges into master.  I know it _can_ be done, but for the sake of my own sanity I'm just going to open a new PR that replaces this one, with the same commits, but targets `master` properly.

PR #2082, opened 2020-01-28
Close connection on SSL connection errors

### LINKED ISSUE #2079: jest reports open handlers when connecting with `ssl=true` to a local PostgreSQL (no SSL support)
**Expected behavior**

when testing a connection with ssl=true to a local PostgreSQL server (no SSL support),
jest should report OK

**Actual behavior**
when testing a connection with ssl=true to a local PostgreSQL server (no SSL support),
jest reports:
```
Jest did not exit one second after the test run has completed.

This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
Done in 61.39s.  
```

**Note:**
It works fine if the connection string contains:

- correct username and password
- correct username and incorrect password

If fails with :

- correct username and password but with ssl=true

**Steps to reproduce**
Clone and run minimum test case at:
https://github.com/cyrilleverrier/pgtest

**Environment**
Version of pg: 7.17.1
OS type (Linux/Windows/Mac): PostgreSQL 9.6.13 on x86_64-pc-linux-musl, compiled by gcc (Alpine 8.3.0) 8.3.0, 64-bit
Version of Node.js: v13.6.0

brianc:
Fixes #2079

### DIFF
--- a/packages/pg/lib/connection.js
+++ b/packages/pg/lib/connection.js
@@ -120,7 +120,9 @@ Connection.prototype.connect = function (port, host) {
     const buff = Buffer.alloc(8)
     buff.writeUInt32BE(8)
     buff.writeUInt32BE(80877103, 4)
-    self.stream.write(buff)
+    if (self.stream.writable) {
+      self.stream.write(buff)
+    }
 
     self.attachListeners(self.stream)
 
@@ -354,6 +356,10 @@ Connection.prototype.end = function () {
   // 0x58 = 'X'
   this.writer.add(emptyBuffer)
   this._ending = true
+  if (!this.stream.writable) {
+    this.stream.end()
+    return
+  }
   return this.stream.write(END_BUFFER, () => {
     this.stream.end()
   })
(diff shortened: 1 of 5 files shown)


packages/pg/lib/connection.js
```diff
@@ -85,11 +85,13 @@ Connection.prototype.connect = function (port, host) {
   this.stream.once('data', function (buffer) {
     var responseCode = buffer.toString('utf8')
     switch (responseCode) {
-      case 'N': // Server does not support SSL connections
-        return self.emit('error', new Error('The server does not support SSL connections'))
       case 'S': // Server supports SSL connections, continue with a secure connection
         break
+      case 'N': // Server does not support SSL connections
+        self.end()
```
charmander: It looks like this might make the connection emit both `'error'` and `'end'`. Is that okay? I don’t think it can in any other situations, but I might be wrong.

brianc: i think that's fine really, but I can change it

brianc:
> It’s probably fine to clean up event edge case consistency later. (It looks like `'end'` might be emitted twice normally anyway?)

ha yeah i just saw that - so yeah I'm punting on that.  Technically the stream _did_ "end" since it did also technically "connect" - but emitting twice isn't nice.

Issue #773, opened 2015-05-04
how is the way for cancel query?


emilioplatzer:
Can I cancel a query inside query.on('row'...? 

I want to do something like:

```js
query.on('row',function(row, result){
    if(row.field1=='ok'){
        result.addRow(row);
    }else{
        result.cancel(row, 'bla bla bla'); 
    }
});
```

The need is to not continue fetching rows if I detect an error condition.

emilioplatzer:
1) Is this the way? 
2) Why I need to pass client.connectionParameters?

```js
query.on('row',function(row, result){
    if(row.field1=='ok'){
        result.addRow(row);
    }else{
        pg.cancel(client.connectionParameters, client, query);
    }
});
```

https://github.com/brianc/node-postgres/wiki/pg#cancelconfig-client-query

emilioplatzer:
In #753 there is an example with `client.cancel(query);` but cancel isn't in the documentation.

joskuijpers:
The cancel code for the native client seems to be completely different from the non-native client. The native simply cancels the query if it is the active one (and removes from queue if not active). The non-native code does some weird stuff with making a connection just to cancel the connection...
See also #753

brianc:
When the PostgreSQL back-end is handling a query on a client it doesn't process any other incoming messages from the socket, so, you can't actually cancel a query from the client that's running the query.  Hence having to pass connection parameters & the client you'd like to cancel in.  

1) The api for canceling aquery here in node-postgres is still _very weird_ though, I agree.  It should probably be made easier to reason about...instead of auto-connecting the best thing is probably have something like this...

```js
var badClient = new Client()
badClient.connect()
badClient.query('pg_sleep(10000000000)')

var goodClient = new Client()
goodClient.cancel(badClient)

```

Something like that - I'll need to check the client/server protocol documentation to see the details - I know you need the client id of the client you'd like to cancel.  Anyways...that might be a bit more explicit and a bit less magical - still cumbersome having to keep around a reference to the old client...but that's how it goes.

2) The native driver doesn't need this because libpq internally does a bunch of this management for you.  I just call `pqcancel` which internally likely opens a new client & sends a cancel request.  Not 100% sure there.

robhaswell:
You should be able to cancel queries using `pg_cancel_backend()`. I think the pattern looks like this:

```js
const pool = pg.Pool()

/* Check out a client */
pool.connect()
.then((c) => {
  /* Get my PID */
[...]

PR #2084, opened 2020-01-28
Add ability to cancel active query

brianc:
This is a proof of concept to cancel active executing query on a client.  It behaves the same way with native bindings and pure JS in that it opens a new connection, bypasses all startup messaging, and dispatches the cancel call [as described here](https://www.postgresql.org/docs/9.3/protocol-flow.html#AEN100004).  Whether the backend is able to cancel the query or not it will terminate the connection which requested cancelation w/o any other messages.

### DIFF
--- a/packages/pg/test/integration/client/cancel-tests.js
+++ b/packages/pg/test/integration/client/cancel-tests.js
@@ -1,0 +1,25 @@
+"use strict"
+var helper = require('./../test-helper')
+var assert = require('assert')
+
+const suite = new helper.Suite()
+
+suite.testAsync('All queries should return a result array', async () => {
+  const pool = new helper.pg.Pool()
+  const client = await pool.connect()
+  const cancledPromise = client.query('SELECT pg_sleep(100000)')
+    .then(() => {
+      throw new Error('this should not resolve because query is cancled')
+    })
+    .catch(err => {
+      return err
+    })
+  await client.cancelActiveQuery()
+  const err = await cancledPromise
+  assert(err instanceof Error)
+  // make sure client is still usable
+  const { rows } = await client.query('SELECT 1 as "foo"')
+  assert.strictEqual(rows.length, 1)
+  assert.deepStrictEqual(rows[0], { foo: 1 })
+  await client.end()
+})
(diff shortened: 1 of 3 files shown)


packages/pg/test/integration/client/cancel-tests.js
```diff
@@ -0,0 +1,25 @@
...
+  const cancledPromise = client.query('SELECT pg_sleep(100000)')
+    .then(() => {
+      throw new Error('this should not resolve because query is cancled')
+    })
+    .catch(err => {
+      return err
+    })
+  await client.cancelActiveQuery()
+  const err = await cancledPromise
+  assert(err instanceof Error)
```
charmander: This will be true even in the `throw new Error('this should not resolve` case. One option:

```js
.then(
  () => {
    throw new Error('this should not resolve because query is cancelled')
  },
  err => err
)
```

sehrope:
I don't think this will work will SSL as nothing is invoking `requestSsl()`. The function body is on the `Connection` prototype but `Client` adds the `connect` event listener to request upgrading the connection:

https://github.com/brianc/node-postgres/blob/c26caa80d2289a50ee81bafb36f1768e95e9f4a4/packages/pg/lib/client.js#L114-L121

And if SSL is enabled in the config properties, `Connection` adds event handlers expecting someone to request the connection upgrade which will never happen: 

https://github.com/brianc/node-postgres/blob/c26caa80d2289a50ee81bafb36f1768e95e9f4a4/packages/pg/lib/connection.js#L81-L94

How about we move all the SSL handling into`Connection`? Could also / instead create an internal `CancelClient` that follows similar init steps to `Client` but then just sends the cancel message.

packages/pg/test/integration/client/cancel-tests.js
brianc: oh yeah your right

brianc:
yah the ssl stuff needs some work for sure - i just pounded this out as a proof of concept to discuss the API a bit yesterday - I circled back to tidy this up in the evening and my internet went out so I was like "meh I'll just look at that later, need to get 8.0 out first." But i do want to deprecate `client.cancel` so when we release 8.0 we can remove that...in favor of a better API.

mhkeller:
Was any canonical way to cancel a query implemented? I see there are a few opinions in the [referenced issue above](https://github.com/brianc/node-postgres/issues/773) but curious why this wasn't included in the library. If there is any guidance on how users should implement this functionality that would be great! Thanks!

PR #2090, opened 2020-01-30
Make notice messages not an instance of Error

### LINKED ISSUE #1982: Notices are Errors, not Messages
(related to #1735)

I think `parseN`/`parseE` are broken https://github.com/brianc/node-postgres/blob/e4578d2c7bca60cfb02ec004daa22c8ff810140e/lib/connection.js#L635-L640 https://github.com/brianc/node-postgres/blob/e4578d2c7bca60cfb02ec004daa22c8ff810140e/lib/connection.js#L592-L604

It looks like the one should create a `Message` and the other should create an `Error`, depending on the `.name`. However, `parseE` does always construct the `input` message with `.name === 'error'`, so the `else` path is never hit.

brianc:
Slight API cleanup to make a notice instance the same shape as it was, but not be an instance of error.  This is a backwards incompatible change though I expect the impact to be minimal.

Closes #1982

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ env:
 node_js:
   - lts/dubnium
   - lts/erbium
-  - 13.6
+  - 13
 
 addons:
   postgresql: "10"
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ env:
 node_js:
   - lts/dubnium
   - lts/erbium
-  - 13
+  - 13.6
 
 addons:
   postgresql: "10"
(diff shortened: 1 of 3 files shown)


brianc:
yeah....I was afraid the tests wouldn't pass.  There was a comment in there from past Brian saying they don't work in travis.  Travis must have some config setting disabling notice messages or something.  I'll see if I can make it skip in travis but still run locally so it'll be covered when I do local dev at least.

Gobie:
You have to call `SET SESSION client_min_messages=notice` on the connection before raising notice.

brianc:
> You have to call `SET SESSION client_min_messages=notice` on the connection before raising notice.

ohh nice I'll try that

PR #2104, opened 2020-02-13
Little typo fix, and add GitHub Sponsors

karlbecker:
Since I believe GitHub Sponsors is your preferred way to donate now, maybe you want to totally get rid of the mention of Patreon, or mention that GitHub Sponsors is preferred?


brianc:
Fantastic! Thanks! ❤️

PR #2106, opened 2020-02-19
Implement handleEmptyQuery for pg-query-stream.

KyleLilly:
`handleEmptyQuery()` wasn't implemented for pg-query-stream.


brianc:
doh! thanks!

PR #2109, opened 2020-02-20
Call callback when end called on unconnected client

### LINKED ISSUE #2108: Different behavior calling .end() on a non connected Client between 7.17 and 7.18
Calling `client.end()` on a `Client` not connected in `pg@7.17.1` fails with an error message

```plain
Socket is closed
```

However in `pg@7.18.1` the same call just hangs indefinitely. I'm using the promise-based call (without the callback) and `await`ing the promise to fulfil.

Is this an expected change?

If the context matters, I have a wrapper around a `Client` and I'm "unit testing" that a call to its closing function throws if the wrapped client is not connected. Before I was just calling `.end()` on the inner client but now the test fails since the promise never fulfilled.

If the change is expected, is there a way of knowing if a given `Client` is connected?

brianc:
Closes #2108

Previously the callback would never fire if one didn't first connect the client.  This fixes the issue & brings the behavior in line w/ the native client.

### DIFF
--- a/packages/pg/lib/client.js
+++ b/packages/pg/lib/client.js
@@ -563,10 +563,10 @@ Client.prototype.end = function (cb) {
   }
 
   if (cb) {
-    this.connection.stream.once('close', cb)
+    this.connection.once('end', cb)
   } else {
     return new this._Promise((resolve) => {
-      this.connection.stream.once('close', resolve)
+      this.connection.once('end', resolve)
     })
   }
 }
--- a/packages/pg/lib/client.js
+++ b/packages/pg/lib/client.js
@@ -546,7 +546,7 @@ Client.prototype.end = function (cb) {
   this._ending = true
 
   // if we have never connected, then end is a noop, callback immediately
-  if (this.connection.stream.pending && !this.connection.stream.connecting) {
+  if (this.connection.stream.readyState === 'closed') {
     if (cb) {
       cb()
     } else {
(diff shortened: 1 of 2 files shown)


brianc:
cc @InExtremaRes  😄

InExtremaRes:
Thank you, this should work! I will test this ASAP.

However I still wonder if could be a way of knowing if a client is connected. Is there a reason that isn't there?

brianc:
> However I still wonder if could be a way of knowing if a client is connected. Is there a reason that isn't there?

I just haven't added it "officially" yet.  I'll be adding this somewhere after 8.0 ships.

packages/pg/lib/client.js
```diff
@@ -545,6 +545,15 @@ Client.prototype.query = function (config, values, callback) {
 Client.prototype.end = function (cb) {
   this._ending = true
 
+  // if we have never connected, then end is a noop, callback immediately
+  if (this.connection.stream.readyState === 'closed') {
+    if (cb) {
+      cb()
+    } else {
+      return Promise.resolve()
```
charmander: For consistency while this feature exists:

```suggestion
      return this._Promise.resolve()
```

InExtremaRes:
I can confirm the change work for me, either with `Promise.resolve()` as with `this._Promise.resolve()`.

BTW, now that this is a monorepo, is there an easy way of testing a packaged version from a PR or branch? I mean, one can `npm install` from a _commit-ish_ but ASFAIK that isn't possible for a subfolder. Do you have a way to install and test and individual branch locally without publishing the package? I guess have a local copy of the branch and `npm link` could work, but I'm asking if you have an "official" o recommended way to do that. Thank you!

packages/pg/lib/client.js
brianc: yeah good call!

PR #2113, opened 2020-02-25
Allow any tls.connect() options

charmander:
#1996 for 8.0. This reverts commit 510a273ce45fb73d0355cf384e97ea695c8a5bcc.


brianc:
ah sweet I think this is basically the last thing to change for 8.0!  I'll try to release it this weekend 😎 .  I'm gonna write up an announcement on the docs so there can be a single place to point to for all the breaking changes, and plans post 8.0 which are mostly performance related.

PR #2115, opened 2020-02-25
Enable old commented out ssl test, add test for self-signed cert failure

brianc:
Just add a few more tests.  A bit redundant, but better than dead code.


brianc:
closing this, taking a different approach

PR #2117, opened 2020-02-25
8.0 Release

brianc:
This bundles up a bunch of breaking changes.  Most of them are small.  I'm going to put up proper announcement documentation on the docs, so I'll keep this PR description pretty short.

probably the biggest breaking change is:
- default to `rejectUnauthorized` to __true__ on ssl connections. 

other breaking changes:
- support __all__ `tls.connect` options as passed on the `ssl` parameter.
- drop support for node older than 8.0
- remove password from being printed out accidentally
- make `native` non-enumerable
- make `Pool` a proper es6 class
- make `Notice` messages _not_ be instances of a JavaScript `Error`

I'm going to really start to focus on performance after this lands, as I've done some benchmarks w/ experimental code here and seeing speed increases up to 50% in some cases.


brianc:
https://github.com/brianc/node-postgres-docs/pull/62 related PR to update the docs.

maximelkin:
Hello!
What do you think about bumping also pg-types?

abrahem79:
`You've provided some excellent examples of how to connect to a PostgreSQL database using `node-postgres` in various scenarios! This covers a lot of ground, from environment variables to programmatic configuration, dynamic passwords, Unix domain sockets, and connection URIs.

While I don't have access to external tools or the ability to run code, I can definitely discuss how these methods relate to Google Cloud and best practices for connecting to databases like Cloud SQL.

Here's how these `node-postgres` connection methods can be applied and what to consider when working with Google Cloud:

### Connecting to Google Cloud SQL with `node-postgres`

Google Cloud SQL is a fully managed relational database service. When connecting to Cloud SQL instances from applications, security and reliability are key. Let's break down your examples in the context of Google Cloud.

#### 1. Environment Variables (`PGUSER`, `PGPASSWORD`, etc.)

**Google Cloud Context:**
This is a very common and recommended approach for deploying applications on Google Cloud, especially with services like Google Kubernetes Engine (GKE) or Cloud Run.

*   **How it works:** You define database credentials as environment variables in your deployment configuration (e.g., a Kubernetes manifest, Cloud Run service definition, or App Engine `app.yaml`). `node-postgres` automatically picks them up.
*   **Security:** Avoid hardcoding sensitive credentials directly in your code. Instead, store them securely. For Google Cloud, this often means using:
    *   **Secret Manager:** A robust service for storing and managing secrets. You can then inject these secrets as environment variables into your application containers.
    *   **Kubernetes Secrets:** If using GKE, you can create Kubernetes Secrets and mount them as environment variables.
*   **Best Practice:** Always use environment variables for sensitive information. This keeps your code clean and prevents accidental exposure of credentials.

#### 2. Programmatic Configuration

**Google Cloud Context:**
While less common for direct database credentials due to security concerns, programmatic configuration is useful when your application needs to dynamically construct connection details or integrate with other configuration management systems.

*   **How it works:** You pass an object with connection properties (like `user`, `password`, `host`, `database`) directly to the `Pool` or `Client` constructor.
*   **Security:** If you go this route, ensure that the values for `user`, `password`, etc., are still sourced from secure locations (e.g., Secret Manager at runtime, not hardcoded in your source code).
*   **Use Cases:** Could be useful if you're building a tool that connects to many different databases and needs to build connection strings based on user input or internal logic.

#### 3. Dynamic Passwords (e.g., `getPassword` callback)

**Google Cloud Context:**
This is highly relevant for Cloud SQL, though the example uses AWS RDS. Cloud SQL offers a secure way to connect using **IAM database authentication**.

*   **How it works (Cloud SQL equivalent):** Instead of a static password, you can configure Cloud SQL to authenticate users based on their Google Cloud IAM credentials. This involves:
    1.  Enabling IAM database authentication on your Cloud SQL instance.
    2.  Granting IAM roles (e.g., `Cloud SQL Client`) to service accounts or user accounts.
    3.  When your application connects, it obtains a **short-lived access token** using its IAM credentials (often automatically handled by Google Cloud client libraries or the Cloud SQL Auth Proxy). This token is then used as the password.
*   **Security Benefits:**
    *   **No long-lived passwords:** Tokens expire, reducing the risk if compromised.
    *   **Centralized access control:** Manage database access through IAM, leveraging Google Cloud's robust identity and access management.
    *   **Auditability:** IAM provides clear audit trails of who accessed what.
*   **How to implement with `node-postgres`:** You'd adapt the `getPassword` function to call a Google Cloud library (or use a helper function) that fetches an access token for the connecting IAM entity.

#### 4. Unix Domain Sockets

**Google Cloud Context:**
This is the **recommended and most secure way to connect to Cloud SQL from services running on the same Google Cloud project**, especially from services like App Engine Standard, Cloud Functions, Cloud Run, or Compute Engine.

*   **How it works:** Instead of connecting over TCP/IP to a `host:port`, you connect to a Unix domain socket path. This connection is typically brokered by the **Cloud SQL Auth Proxy**.
    *   The Cloud SQL Auth Proxy establishes a secure, encrypted connection to your Cloud SQL instance using IAM credentials.
    *   It then exposes a Unix domain socket (or a local TCP port) on the application's host.
    *   Your `node-postgres` client connects to this local socket, and the proxy forwards the traffic securely to Cloud SQL.
*   **`node-postgres` Configuration:** As shown in your example, the `host` parameter is set to the socket path, typically `/cloudsql/PROJECT_ID:REGION:INSTANCE_NAME`.
*   **Benefits:**
    *   **Encryption:** All traffic through the proxy is encrypted with TLS 1.2.
    *   **Authorization:** The proxy uses IAM permissions to control who can connect to which instance.
    *   **No IP allowlists needed:** You don't need to configure public IP addresses or allowlists, simplifying network configuration and enhancing security.
    *   **Automatic IP management:** Handles instance IP address changes transparently.
*   **Best Practice:** Always use the Cloud SQL Auth Proxy for connections within Google Cloud, especially for production environments.

#### 5. Connection URI (`postgresql://...`)

**Google Cloud Context:**
Very useful for quickly configuring connections, particularly when environment variables provide the full URI.

*   **How it works:** A single string encapsulates all connection details.
*   **Security:** Similar to environment variables, ensure this URI is stored and injected securely (e.g., from Secret Manager). Avoid hardcoding it.
*   **Use Cases:** Common in PaaS (Platform as a Service) environments or when you need to pass a single connection string around.

### Summary for Google Cloud

For most applications on Google Cloud, particularly those connecting to Cloud SQL, the following are the most common and recommended approaches:

1.  **Unix Domain Sockets with the Cloud SQL Auth Proxy:** This is the gold standard for secure, internal connections within Google Cloud.
2.  **IAM Database Authentication with dynamic passwords:** Provides passwordless authentication using Google Cloud IAM.
3.  **Environment Variables:** For storing general configuration, especially when combined with Secret Manager for sensitive data.

Your provided `node-postgres` examples perfectly illustrate the flexibility available to integrate with these Google Cloud best practices.`

PR #2120, opened 2020-02-26
fix: Prevent closing the portal twice

### LINKED ISSUE #2119: Cursor closed twice when used in async iterable
Running the following code produces a race condition causing the pg-cursor to be closed twice:

```js
const { Pool } = require('pg')
const PgQueryStream = require('pg-query-stream')

const connectionString = "XXX"
const pool = new Pool({ connectionString })

async function* stream(query, values, { highWaterMark = 1 } = {}){
    const stream = new PgQueryStream(query, values, { highWaterMark })
    this.query(stream)
    return yield* stream
}

const main = async () => {
  const connection = await pool.connect()
  try {
    for await(const item of stream.call(connection, 'SELECT TRUE')) break
    for await(const item of stream.call(connection, 'SELECT TRUE')) break
    for await(const item of stream.call(connection, 'SELECT TRUE')) break
    for await(const item of stream.call(connection, 'SELECT TRUE')) break
    for await(const item of stream.call(connection, 'SELECT TRUE')) break
    for await(const item of stream.call(connection, 'SELECT TRUE')) break
    for await(const item of stream.call(connection, 'SELECT TRUE')) break
  } finally {
    await connection.release()
  }
}

main()

```

Running this code (multiple times) with a fast connection (local pg server) results in the following error:

```
~/node_modules/pg/lib/client.js:337
    if (self.activeQuery.name) {
                         ^

TypeError: Cannot read property 'name' of null
    at Connection.<anonymous> (~/node_modules/pg/lib/client.js:337:26)
    at Connection.emit (events.js:321:20)
    at Socket.<anonymous> (~/node_modules/pg/lib/connection.js:134:12)
    at Socket.emit (events.js:321:20)
    at addChunk (_stream_readable.js:297:12)
    at readableAddChunk (_stream_readable.js:273:9)
    at Socket.Readable.push (_stream_readable.js:214:10)
    at TCP.onStreamRead (internal/stream_base_commons.js:186:23)
```

This error is caused by the fact that the `_closePortal()` will get called twice.
1) first because NodeJs will try and `destroy()` the steam when the loop is `break`en. This will cause `close()` to be called.
2) because `handleCommandComplete()` will be triggered by the pg connection listener

matthieusieben:
Fixes #2119

### DIFF
--- a/packages/pg-cursor/index.js
+++ b/packages/pg-cursor/index.js
@@ -82,6 +82,12 @@ Cursor.prototype._shiftQueue = function () {
 }
 
 Cursor.prototype._closePortal = function () {
+  // This method might get called twice in case handleCommandComplete is
+  // triggered after the user called stream.destroy().
+  if (this.state === 'done') return
+
+  this.state = 'done'
+
   // because we opened a named portal to stream results
   // we need to close the same named portal.  Leaving a named portal
   // open can lock tables for modification if inside a transaction.
@@ -120,9 +126,6 @@ Cursor.prototype._sendRows = function () {
 
 Cursor.prototype.handleCommandComplete = function (msg) {
   this._result.addCommandComplete(msg)
-  // The cursor might have been .close()'d right before this listener gets
-  // called. Let's not close the portal twice.
-  if (this.state === 'done') return
   this._closePortal()
 }
 
@@ -132,7 +135,7 @@ Cursor.prototype.handlePortalSuspended = function () {
 
 Cursor.prototype.handleReadyForQuery = function () {
   this._sendRows()
-  this.state = 'done'
+  this._closePortal()
   this.emit('end', this._result)
 }
 
@@ -193,7 +196,6 @@ Cursor.prototype.close = function (cb) {
     }
   }
   this._closePortal()
-  this.state = 'done'
   if (cb) {
     this.connection.once('readyForQuery', function () {
       cb()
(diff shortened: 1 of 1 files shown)


matthieusieben:
Not sure this is the right way to fix this (might be better to check in `_closePortal()` itself)

brianc:
ah dang - w/ the race condition situation I'm guessing there's no good way to write a test for this?

matthieusieben:
Not sure indeed. Maybe we can find a way to force a call to close() right after we get a first row (e.g. by monkey matching the listener?), and then wait for the complete command listener to kick in.

I am not too familiar with this lib's internals so I don't feel too confortable fixing it myself.

brianc:
Yeah I spent some time this morning looking at this, trying to force the error to happen...I am not able to simulate the race condition here.  I'm definitely down w/ you giving it a shot through any level of monkey patching you think would accurately recreate it.  If you can't, I can just merge this as it the behavior itself makes sense...though I do think moving `this.state = 'done'` _into_ `_closePortal` and having that function also check the state and no-op out if it's already switched into done is probably more "clean" in that the state is checked & updated in the same place that way.

Sorry for the delay. Thoughts?

matthieusieben:
I guess you could monkey patch by adding a `setTimeout` around everything inside `Cursor.prototype.close`.

I made the suggested change though I am not sure about calling `_closePortal` from `handleReadyForQuery`. I have no idea what the internal state machine of a cursor is inside of PG (where wan I find this documentation?).

matthieusieben:
Hey @brianc did you get time to look into this ?

brianc:

matthieusieben:
closing in favor of https://github.com/brianc/node-postgres/pull/2609

Issue #2112, opened 2020-02-24
7.18.2: "Connection terminated unexpectedly" when using client.query with a pool when pool has been idle for 10 minutes


jcollum:
Code is running in a lambda. When we run the function below it runs fine. But if we run it and then wait 10 minutes then run it again, we get an error every time. We are querying Aurora Postgres on AWS. 

Code: 

```
const {Pool} = require('pg');

// Global Connection, can be re-used!!
const pgPool = new Pool({
    user: process.env.PG_USER,
    host: process.env.PG_HOST,
    database: process.env.PG_DATABASE,
    password: process.env.PG_PASSWORD,
    port: process.env.PG_PORT,
    max: process.env.MAX_CLIENTS
});

pgPool.on('error', (err, client) => {
    console.error('Unexpected error in Postgress connection pool', err);
});

async function performQuery(event) {

    let queryString = null;
    let args = null;

    try {
        // some setup code here...  
        const client = await pgPool.connect();
        try {
            const res = await client.query(queryString, args);
            return res.rows;
        } finally {
            client.release();
        }
    } catch (err) {
        console.error('Problem executing export query:');
        console.error(err); // <-- this line is in the log below  
        throw err;
    }
}
```

This is what I see in the cloudwatch logs: 

```
{
    "errorType": "Error",
    "errorMessage": "Connection terminated unexpectedly",
    "stack": [
    "Error: Connection terminated unexpectedly",
    "    at Connection.<anonymous> (/var/task/node_modules/pg/lib/client.js:255:9)",
    "    at Object.onceWrapper (events.js:312:28)",
    "    at Connection.emit (events.js:223:5)",
    "    at Connection.EventEmitter.emit (domain.js:475:20)",
    "    at Socket.<anonymous> (/var/task/node_modules/pg/lib/connection.js:78:10)",
    "    at Socket.emit (events.js:223:5)",
    "    at Socket.EventEmitter.emit (domain.js:475:20)",
    "    at TCP.<anonymous> (net.js:664:12)"
]
}
```

I've tried a few variations on this but the constant is the 10 minutes and the use of the Pool. To me this code is almost identical to the code in https://node-postgres.com/features/pooling. 

So far it looks like the problem has been solved by using a Client instead: 

```
const client = new Client({
            user: process.env.PG_USER,
            host: process.env.PG_HOST,
            database: process.env.PG_DATABASE,
            password: process.env.PG_PASSWORD,
            port: process.env.PG_PORT
        });

[...]

PR #2121, opened 2020-02-26
Attempt to handle a stream being silently killed in the background

brianc:
I _think_ this will fix the issue outlined in #2112 

I just pounded out this initial version pretty quickly as I'm not 100% sure it'll fix the issue yet....as I'm not sure my test is _exactly_ replicating what's going on in lambda.  In the test I'm "cleanly" severing the connected stream to the backend which doesn't emit an error anywhere so the client/pool still thinks its okay.  

Could someone from #2112 try applying this patch to their code in lambda and see if it fixes the issue? If so I'll polish this up, write a couple more tests, and get a patch version fix out.

### DIFF
--- a/packages/pg-pool/index.js
+++ b/packages/pg-pool/index.js
@@ -116,11 +116,20 @@ class Pool extends EventEmitter {
     const pendingItem = this._pendingQueue.shift()
     if (this._idle.length) {
       const idleItem = this._idle.pop()
-      clearTimeout(idleItem.timeoutId)
       const client = idleItem.client
-      const idleListener = idleItem.idleListener
+      clearTimeout(idleItem.timeoutId)
 
-      return this._acquireClient(client, pendingItem, idleListener, false)
+      // TODO(bmc): we need a better state represetation on the client itself
+      // to indicate if it's connecting, idle, running a query, closed, etc...
+      // but for now we can look at it's stream.
+      // remove this client, it's died in the background - this can happen
+      // in aws lambdas - they go idle and the streams are closed without an error
+      if (client.connection.stream.readyState !== 'open') {
+        this._remove(client)
+      } else {
+        const idleListener = idleItem.idleListener
+        return this._acquireClient(client, pendingItem, idleListener, false)
+      }
     }
     if (!this._isFull()) {
       return this.newClient(pendingItem)
--- a/packages/pg-pool/index.js
+++ b/packages/pg-pool/index.js
@@ -121,10 +121,12 @@ class Pool extends EventEmitter {
 
       // TODO(bmc): we need a better state represetation on the client itself
       // to indicate if it's connecting, idle, running a query, closed, etc...
-      // but for now we can look at it's stream.
+      // as right now this hacky fix will only work for the JS client
+      // since the native client doesn't _have_ a connection property
+
       // remove this client, it's died in the background - this can happen
       // in aws lambdas - they go idle and the streams are closed without an error
-      if (client.connection.stream.readyState !== 'open') {
+      if (client.connection && client.connection.stream.readyState !== 'open') {
         this._remove(client)
       } else {
         const idleListener = idleItem.idleListener
--- a/packages/pg-pool/test/background-socket-terminiation.js
+++ b/packages/pg-pool/test/background-socket-terminiation.js
@@ -1,0 +1,31 @@
+'use strict'
+const net = require('net')
+const co = require('co')
+const expect = require('expect.js')
+
+const describe = require('mocha').describe
+const it = require('mocha').it
+
+const Pool = require('../')
+
+describe('a socket disconnecting in the background', () => {
+  it('should leave the pool in a usable state', async () => {
+    const pool = new Pool({ max: 1 })
+    // just run any arbitrary query
+    const client = await pool.connect()
+    await client.query('SELECT NOW()')
+    // return the client
+    client.release()
+
+    // now kill the socket in the background
+    client.connection.stream.end()
+
+    // now try to query again, it should work
+    await pool.query('SELECT NOW()')
+    await pool.query('SELECT NOW()')
+    await pool.query('SELECT NOW()')
+    await pool.query('SELECT NOW()')
+
+    await pool.end()
+  })
+})


packages/pg-pool/index.js
```diff
@@ -116,11 +116,20 @@ class Pool extends EventEmitter {
     const pendingItem = this._pendingQueue.shift()
     if (this._idle.length) {
       const idleItem = this._idle.pop()
-      clearTimeout(idleItem.timeoutId)
       const client = idleItem.client
-      const idleListener = idleItem.idleListener
+      clearTimeout(idleItem.timeoutId)
```
briandamaged: Just curious: was there any particular reason why this line needed to be moved?

brianc: not really, no

packages/pg-pool/test/background-socket-terminiation.js
```diff
@@ -0,0 +1,31 @@
...
+
+    // now kill the socket in the background
+    client.connection.stream.end()
+
+    // now try to query again, it should work
+    await pool.query('SELECT NOW()')
+    await pool.query('SELECT NOW()')
+    await pool.query('SELECT NOW()')
+    await pool.query('SELECT NOW()')
+
```
briandamaged: Perhaps you should also add something like:

```
const client2 = await pool.connect();
try {
  // Verify that `client` was removed from the pool and replaced w/ a new client
  expect(client2).to.not.equal(client);
} finally {
  await client2.release();
}
```

briandamaged: (Otherwise, it's possible that `client` just magically fixed itself through some other means.  So, adding this check helps to rule out that possibility)

brianc: yah that's not a bad idea!  I'd like to hold off on merging this until I get some confirmation its actually fixing things in aws lambda.  It also unfortunately doesn't work w/ the native bindings yet so I'll need to somehow figure that out.

packages/pg-pool/index.js
briandamaged: np -- just wanted to confirm that there wasn't some important reason that I completely overlooked.  😅

brianc: haha yeah nah just moving code slightly & bumped that line!

packages/pg-pool/test/background-socket-terminiation.js
briandamaged: Sounds good.

As I mentioned in [this thread](https://github.com/brianc/node-postgres/issues/2112#issuecomment-591040368), I suspect that both `"pg"` and `"knex"` were encountering the same issue within their respective pool implementations.  So, we'll most likely need to implement a similar fix on the `"knex"` side.

packages/pg-pool/index.js
```diff
@@ -116,11 +116,22 @@ class Pool extends EventEmitter {
...
+      // TODO(bmc): we need a better state represetation on the client itself
+      // to indicate if it's connecting, idle, running a query, closed, etc...
+      // as right now this hacky fix will only work for the JS client
+      // since the native client doesn't _have_ a connection property
 
-      return this._acquireClient(client, pendingItem, idleListener, false)
+      // remove this client, it's died in the background - this can happen
+      // in aws lambdas - they go idle and the streams are closed without an error
+      if (client.connection && client.connection.stream.readyState !== 'open') {
+        this._remove(client)
```
briandamaged: In the current logic: if it notices that the next-available `client` is already dead, then it removes it from the pool.  Afterwards, it instantiates a new client to replace it.

Would there be any advantage to looping on the `client` eviction step?  In pseudo-code:

```
while(this._idle.length) {
  const client = this._idle.pop();

  // ... Other important details I'm ignoring for the moment...

  if(isDead(client)) {
    this.remove(client);
  } else {
    return client
  }
}

if(!this._isFull()) {
  return this.newClient(...);
}

throw new Error('unexpected condition');
```

This seems like it helps in 2 ways:

1. It clears out the dead clients more quickly, and
2. It avoids the need to wait for `this.newClient(..)` when there is already another client available.

brianc: probably, yeah - good suggestions!

592da: hi guys, I want to test this branch, but npm says its missing the version... tried to fork it, put the version (https://github.com/592da/node-postgres) and then compiling failed : `Cannot find module 'pg' `

@brianc

briandamaged: Hmm... I suspect that's because this is a mono-repo that is built via [Lerna](https://github.com/lerna/lerna) .  You might need to build it locally first and then link to the build artifacts.  (@brianc might know another approach.  I haven't actually worked with this codebase before).

I don't remember the exact build steps for Lerna off the top of my head, but I think it's something like:

1. Install Lerna
2. Navigate to the root of the repo
3. Run the command:  `lerna bootstrap`

brianc: > hi guys, I want to test this branch, but npm says its missing the version... tried to fork it, put the version (https://github.com/592da/node-postgres) and then compiling failed : `Cannot find module 'pg' `

if you wanna test it it's probably easiest to just apply the diff of the 1 file i changed inside your `node_modules/pg-pool` folder to `index.js` (or just copy/paste the changes).

592da: thanks for your reply @briandamaged, unfortunately, it didn't work, maybe because its also imported from `typeorm`

592da: @brianc , having the same issue...
```
2020-02-26T21:38:37.302Z	da5bbf66-06b7-4c69-9c85-ab84b1b777a1	INFO	Error: Connection terminated unexpectedly
    at Connection.<anonymous> (/var/task/src/rest/webpack:/node_modules/pg/lib/client.js:255:1)
    at Object.onceWrapper (events.js:312:28)
    at Connection.emit (events.js:223:5)
    at Connection.EventEmitter.emit (domain.js:475:20)
    at Socket.<anonymous> (/var/task/src/rest/webpack:/node_modules/pg/lib/connection.js:139:1)
    at Socket.emit (events.js:228:7)
    at Socket.EventEmitter.emit (domain.js:475:20)
    at endReadableNT (_stream_readable.js:1185:12)
    at processTicksAndRejections (internal/process/task_queues.js:81:21)```

File: packages/pg-pool/index.js [hunk 4]
```diff
@@ -116,11 +116,22 @@ class Pool extends EventEmitter {
...
+
+      // TODO(bmc): we need a better state represetation on the client itself
+      // to indicate if it's connecting, idle, running a query, closed, etc...
+      // as right now this hacky fix will only work for the JS client
+      // since the native client doesn't _have_ a connection property
 
-      return this._acquireClient(client, pendingItem, idleListener, false)
+      // remove this client, it's died in the background - this can happen
+      // in aws lambdas - they go idle and the streams are closed without an error
+      if (client.connection && client.connection.stream.readyState !== 'open') {
```
briandamaged: I'm seeing some mention online that `stream.readyState` is deprecated/unsupported.  It might be safer to do this instead:

```
(client.connection.stream.readable && client.connection.stream.writable)
```

When experimenting w/ things locally, I'm finding that these properties change as soon as you call `stream.end()`.

brianc: yeah those properties don't exist in all supported versions of node that pg supports.  stream.readyState _does_ exist in all supported versions.

brianc:
I put this up as a proof of concept fix.  If this doesn't fix the issue please lemme know & I'll close it.  If it _does_ fix the issue, let me know & I'll clean it up and land it for real.

If this doesn't end up fixing it I hope this can be a starting point for someone to do more investigation and eventually submit a pull request.  As I don't have a lambda or a way to reproduce the problem I don't think my time is best spent trying to track this particular issue down.  So...I'm open to a PR to fix the issue & supplying ideas and guidance but I'm likely not going to push on fixing this issue myself until I take care of other higher priority issues for the project.

packages/pg-pool/index.js
brianc: > When experimenting w/ things locally, I'm finding that these properties change as soon as you call `stream.end()`.

you should check if `readyState` changes as well (I think it does) - reason is any official patch is gonna need to work in node 8, 10, 12, and 13.  So if its "officially deprecated" in node but still works, I'd use it (as it probably _wasn't_ deprecated in older versions of node).  Typically I do that, write a bunch of tests for the feature, and then if a test breaks on an edge release of node I either decide to drop old version support at that point or make a compatibilty branch in the code that will handle both old and new versions of node.

briandamaged: Sure -- I'll give that a try after the initial experiment.  Speaking of which: you can find the experiment within the `knex` pooling here:

https://github.com/briandamaged/knex/tree/experiment/aws-disconnections

And, here's the link to the specific commit:

https://github.com/briandamaged/knex/commit/1490611ad9fca11e42ccfeb5fdfe4e482ce6bcbc

So, I think this roughly aligns w/ your implementation.  (Aside from the `readyState` piece)

brianc: yah totally actually I don't wanna burden you w/ backwards compatiblity - if you find a way to make it work in lambda w/ a version of node, i'll do my best to back port it and "production-ify" it for this library.  Mostly it's just hard for me to isolate & test since I don't have a lambda nor much experience w/ them so me setting all that testing infrastrucutre up is a huge time sync.  But if you figure out the root cause, I'll help any way I can getting a fix out

briandamaged: Ah -- yeah, I currently don't have a way to recreate the issue, either.  I just know that Knex's current implementation relies upon Events to detect the closed connections.  So, this just seems like a possible way to work around that issue.

Hopefully we'll hear some good news from this experiment!  If so, then we can figure out how to formulate the concepts into something that is production-worthy.  (One thought: we might even need to swap out implementations depending upon Node version.  Ex: use `readable` / `writable` if the Node version supports those properties; otherwise, fall back to `readyState`)

michaelseibt:
I'll give this a try. Since we're not able to reproduce the issue reliably, it could take a while to come back to you. Thanks for the patience!

michaelseibt:
Here is what I tried: I've added a couple of logging statements around the lines of code touched by this PR:

```
console.log("[PG-NODE] Client state is", client.connection.stream.readyState);
      if (client.connection && client.connection.stream.readyState !== 'open') {
        console.warn("[PG-NODE] Removing faulty client");
        this._remove(client)
      } else {
        const idleListener = idleItem.idleListener
        console.warn("[PG-NODE] Returning client");
        return this._acquireClient(client, pendingItem, idleListener, false)
      }
```

Built it, shipped it, waited. After a while, these logs came up:

```
08:47:13 INFO	[PG-NODE] Client state is open
08:47:13 WARN	[PG-NODE] Returning client
08:56:14 INFO	[PG-NODE] Client state is open
08:56:14 WARN	[PG-NODE] Returning client
08:56:14 ERROR	Uncaught Exception [...] Connection terminated unexpectedly
08:59:18 INFO	[PG-NODE] Client state is open
08:59:18 WARN	[PG-NODE] Returning client
... repeats a couple of times until ...
09:09:24 INFO	[PG-NODE] Client state is open
09:09:24 WARN	[PG-NODE] Returning client
09:09:24 ERROR	Uncaught Exception [...] Connection terminated unexpectedly
```

Is there something else I can provide to help?

592da:
any news regarding ?

osdiab:
looks like this PR is stalling?

brianc:
> looks like this PR is stalling? any way to help?

> any news regarding ?

I don't really use lambda - the best way to help would be to contribute a PR to fix the issue - without me having access to lambda I'm likely not going to be very helpful in finding a fix.  Will definitely be down to review & merge any fix if you find it!

dgobaud:
we are having this same problem in Google Cloud it seems

bryantbiggs:
bump - hoping for some resolution 😬 🙏🏽

mikelax:
I am also seeing the same issues reported in issue #2112 so I would love to help out however I can if this is a possible fix for it. Given more than a year has elapsed, I was wondering if this fix becomes easier if you don't need to support older versions of nodejs that are now out of support. 

You should now be able to say the minimum supported version is v12, and drop support for v8 and v10 (lts/dubnium).

PR #2126, opened 2020-02-28
Convert Query to an ES6 class

charmander:
The last missing `new` deprecation warning for pg 8.


sehrope:
LGTM.

Why just this one file though and not also Result and Connection?

charmander:
@sehrope #2021 added the warning to constructors that worked without `new`. `Result` and `Connection` already break when used without `new`, so they can be upgraded at any time without backwards compatibility issues.

sehrope:
Ah makes sense. Thanks for the explanation.

brianc:
Ah nice - I'll review this one (it's a big bigger) toady & then merge it.  Outside of this - anything left u know of outstanding for 8.0? I really want to tackle this performance stuff for a few weeks.

charmander:
> anything left u know of outstanding for 8.0?

Not that I can think of!

brianc:
sorry for the delay - corona virus panic and economic downturn causing a lot of disruption and general anxiety for me.  Will be getting on this today & will try to get 8.0 out tomorrow! Stay healthy. ❤️

PR #2127, opened 2020-02-28
Split message properties from error and notice objects

charmander:
This avoids exposing an unusual `name` on errors and the oddly isolated `length` on both objects and removes the special handling of the “error” message name.

The TypeScript probably still needs some work.


brianc:
> oddly isolated `length` on both objects

what do you mean by oddly isolated length?

charmander:
@brianc Oh, that the `length` property isn’t really related to anything the user can use. It’s for internal use, right?

charmander:
Hmm, hiding them might not be worth the effort though.

brianc:
Yeah it's a bit weird to have them on there I guess? As in they're not really _useful_ outside the parser.  OTOH I wanted the parser to actually deliver back the _entire_ shape of every backend message because it didn't really hurt anything & it's a bit more comprehensive.

PR #2147, opened 2020-03-27
Add maxUses pool config option

chriskchew:
This PR is intended to help with balancing connections to a growing read replica cluster.  Specifically, this scenario came up when adopting AWS Aurora in an environment where the read replica cluster expands and contracts based on weekly traffic cycles.

The PR adds an optional pool configuration `maxUses` and adds logic to the Pool that removes a connection upon release when it has been acquired and released `maxUses` number of times.

I expanded the existing "item wrapper" pattern used for IdleItem and added a `ClientItem` wrapper for the items put into the `_clients` array.  The ClientItem wrapper holds the use count for the client instance so that the release function can have the data it needs in order to determine when the connection has reached the end of its run.

I also added a few changes for issues I encountered while setting up my local dev environment:

* Added a little logic to `checkType` in the bring-your-own-promise to explicitly fail when unexpected error occurs

* Added some instructions in the main README listing the steps for dev setup

Thank you for considering these changes!

### DIFF
--- a/packages/pg-pool/test/bring-your-own-promise.js
+++ b/packages/pg-pool/test/bring-your-own-promise.js
@@ -8,29 +8,34 @@ const BluebirdPromise = require('bluebird')
 
 const Pool = require('../')
 
-const checkType = promise => {
+const checkType = (promise, expectError) => {
   expect(promise).to.be.a(BluebirdPromise)
-  return promise.catch(e => undefined)
+  return promise.catch(e => {
+    if (expectError) {
+      return undefined
+    }
+    expect().fail(e)
+  })
 }
 
 describe('Bring your own promise', function () {
   it('uses supplied promise for operations', co.wrap(function * () {
     const pool = new Pool({ Promise: BluebirdPromise })
-    const client1 = yield checkType(pool.connect())
+    const client1 = yield checkType(pool.connect(), false)
     client1.release()
-    yield checkType(pool.query('SELECT NOW()'))
-    const client2 = yield checkType(pool.connect())
+    yield checkType(pool.query('SELECT NOW()'), false)
+    const client2 = yield checkType(pool.connect(), false)
     // TODO - make sure pg supports BYOP as well
     client2.release()
-    yield checkType(pool.end())
+    yield checkType(pool.end(), false)
   }))
 
   it('uses promises in errors', co.wrap(function * () {
     const pool = new Pool({ Promise: BluebirdPromise, port: 48484 })
-    yield checkType(pool.connect())
-    yield checkType(pool.end())
-    yield checkType(pool.connect())
-    yield checkType(pool.query())
-    yield checkType(pool.end())
+    yield checkType(pool.connect(), true)
+    yield checkType(pool.end(), true)
+    yield checkType(pool.connect(), true)
+    yield checkType(pool.query(), true)
+    yield checkType(pool.end(), true)
   }))
 })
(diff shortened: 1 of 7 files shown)


packages/pg-pool/test/bring-your-own-promise.js
```diff
@@ -8,29 +8,34 @@ const BluebirdPromise = require('bluebird')
 
 const Pool = require('../')
 
-const checkType = promise => {
+const checkType = (promise, expectError) => {
   expect(promise).to.be.a(BluebirdPromise)
-  return promise.catch(e => undefined)
+  return promise.catch(e => {
```
chriskchew: Note for reviewer:  It took me a while to figure out why this test kept failing with an error about the value being undefined, so I added the new argument so we can differentiate an unexpected error scenario and subsequently report it as a test failure.

chriskchew:
@brianc I'm wondering if there is something I can do to make it easier for you to review this PR?  Let me know if you have any questions or initial reactions to the proposed changes.  Thank you!

brianc:
sorry for the delay - I'll take a look at this tomorrow morn!

packages/pg-pool/test/bring-your-own-promise.js
brianc: nice - thanks.  As an aside, I think it's probably time I deprecate BYOP for pg@9.0 as all current supported versions of node include a native promise.

chriskchew:
I appreciate your thoughtful review!

> Question though...how do you feel about just adding a _useCount variable or something directly to the client instance versus wrapping it?

This is a great question.  My initial thought was similar to yours, that I would add _useCount to the Client and simplify the Pool changes to simply increment that number and close it when the Client is deemed to be "expended".

What drew me away from that approach was partially that I didn't understand if it was safe to simply "patch" that value into the Client instance when it is instantiated in `Pool.newClient()`, and then also because the "max uses" seemed mostly to be a _Pool_ concern, and thus would be data that is held on the Pool side of the relationship.

In terms of adding that property to client, would the recommended approach simply be to patch that property on the Client instance inside `newClient()` and increment it in `release()`, or do I need to more explicitly define a field on the Client and NativeClient prototypes, probably along with accessor methods `incrementUseCount()` and `getUseCount()`?

brianc:
> What drew me away from that approach was partially that I didn't understand if it was safe to simply "patch" that value into the Client instance when it is instantiated in `Pool.newClient()`

Yeah...should be safe!  Maybe call the prop something slightly longer like `_poolUseCount` or something so it's obviously coming from the pool.

> the "max uses" seemed mostly to be a _Pool_ concern

That's true, so is monkey patching `.release()` on a client to some extent.  I think its preferable in this case over possibly breaking things by not having `pool._clients` not actually be clients.



> In terms of adding that property to client, would the recommended approach simply be to patch that property on the Client instance inside `newClient()` and increment it in `release()`

Yeah that's what I'd do.  Just patch it in `newClient` and increment it in `release`.  You could actually both patch and increment it in `release`.  Something like near on this line:

https://github.com/brianc/node-postgres/blob/master/packages/pg-pool/index.js#L265
```js
client._poolUseCount = (client._poolUseCount || 0) + 1
```
💭

chriskchew:
Thanks for the direction.  I will probably start with a fresh branch just to keep a cleaner commit log since the bulk of the changes need to be unwound.

brianc:
haha okay up to you! I usually squash merge anyway so whatever is easiest
for you!

chriskchew:
Closing this in favor of #2157

PR #2149, opened 2020-03-27
add finish function

mslatt:
similar to `end`, `finish` closes the connections and pool but only after the queue has processed all currently pending queries, instead of cutting out pending queries like `end`

### DIFF
--- a/packages/pg-pool/index.js
+++ b/packages/pg-pool/index.js
@@ -77,6 +77,7 @@ class Pool extends EventEmitter {
     this._idle = []
     this._pendingQueue = []
     this._endCallback = undefined
+    this.finishing = false
     this.ending = false
     this.ended = false
   }
@@ -91,7 +92,7 @@ class Pool extends EventEmitter {
       this.log('pulse queue ended')
       return
     }
-    if (this.ending) {
+    if (this.ending && (!this.finishing || !this._pendingQueue.length)) {
       this.log('pulse queue on ending')
       if (this._idle.length) {
         this._idle.slice().map(item => {
@@ -368,6 +369,11 @@ class Pool extends EventEmitter {
     this._endCallback = promised.callback
     this._pulseQueue()
     return promised.result
+  }
+
+  finish (cb) {
+    this.finishing = true
+    return this.end(cb)
   }
 
   get waitingCount () {


mslatt:
@brianc wondering what you think about this, change the default behavior of `end` to finish the pending queue, and add a `force` argument that will not process the pending queries (how end currently works). This matches how node-mongodb works by finishing pending queries by default and allow a force argument ([link](https://mongodb.github.io/node-mongodb-native/3.0/api/MongoClient.html#close)). I updated this PR with that commit above, that simplifies the change and then dont need a separate function

mslatt:
Can someone add tests for `pool.end` for this PR as requested by @brianc, I have limited experience with these Mocha tests they dont mean anything to me, maybe someone who uses these could update them https://github.com/brianc/node-postgres/blob/master/packages/pg-pool/test/ending.js

mslatt:
@brianc I tried running the existing tests on master branch for pg-pool and it is throwing require errors so Im not sure the existing tests are even working, last commit was 3 years ago?

brianc:
> @brianc I tried running the existing tests on master branch for pg-pool and it is throwing require errors so Im not sure the existing tests are even working, `pg-pool/test/ending.js` is a single commit from 3 years ago?

the tests run in CI on every commit...I am preeeetty sure they're working.  What commands exactly did you run & what was the output?

mslatt:
@brianc what am I supposed to be running? you dont have docs on how to set up or run testing. I installed mocha globally and ran `mocha` in the `node-postgres/packages/pg-pool` directory and it is throwing `DeprecationWarning: Configuration via mocha.opts is DEPRECATED and will be removed from a future version of Mocha. Use RC files or package.json instead` and doesnt appear to complete any tests. I'd rather someone who actually uses these tests update them to something more recent, rather than me who doesnt need the test to mess around with something I dont know

mslatt:
Ok I see some in your Readme to use `yarn test`

mslatt:
@brianc ok I added test, hopefully that is good enough to get this merged, really need this PR asap, all these poor coders trying to finish their pools without cutting short their queries need this

mslatt:
@brianc I am trying to write better test for you here but I am running into this bug I mentioned earlier https://github.com/brianc/node-pg-pool/issues/134 it is throwing off my test to have pool.end complete before the last pool.query https://github.com/brianc/node-pg-pool/issues/134 any plans on fixing that?

mslatt:
@brianc ok I wrote better tests. Only problem is that bug https://github.com/brianc/node-pg-pool/issues/134 is causing this line to be 9 instead of 10 https://github.com/brianc/node-postgres/pull/2149/files#diff-57867471f03f2f65dd6fde482c6f9ea2R59

fix that bug then I will update this to 10 instead of 9

mslatt:
Anyone have any thoughts on this PR? Anyone else think this should be the default behavior or is it just me? Also is this issue https://github.com/brianc/node-postgres/issues/2163 considered a bug that needs to be fixed? there are no comments on this PR or issue @brianc @charmander

charmander:
Sure, I think it makes sense for this to be the default behaviour. Is `force` even necessary? I’m not sure when it would be the best solution to a problem.

mslatt:
I don't think I would ever use `force` but best to give the option of dropping pending queries to preserve the same functionality as previous pool.end for those that want it. Anyone feel free to update this PR with any changes

mslatt:
does this line also need to be updated? by changing `this.ending` to `(this.ending && (!this._pendingQueue.length || this.force))` https://github.com/brianc/node-postgres/pull/2149/files#diff-4686c3e65ee230d720d4096cf8bdfc84R300

charmander:
Since the author has deleted their account and can’t update the PR anymore, I’ll propose a version of this in a new one.

PR #2158, opened 2020-04-09
Make several small speed tweaks

brianc:
I used to use a while loop a while ago.  I think some microbenchmark i ran on some version of node showed it made no difference in speed.  It makes a small difference now, which adds up when reading a lot of rows.

### DIFF
--- a/packages/pg-protocol/src/b.ts
+++ b/packages/pg-protocol/src/b.ts
@@ -2,11 +2,15 @@
 
 import { Writer } from './buffer-writer'
 import { serialize } from './index'
+import { BufferReader } from './buffer-reader'
 
 const LOOPS = 1000
 let count = 0
 let start = Date.now()
 const writer = new Writer()
+
+const reader = new BufferReader()
+const buffer = Buffer.from([33, 33, 33, 33, 33, 33, 33, 0])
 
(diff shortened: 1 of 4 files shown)


packages/pg-protocol/src/b.ts
```diff
@@ -2,21 +2,25 @@
 
 import { Writer } from './buffer-writer'
 import { serialize } from './index'
+import { BufferReader } from './buffer-reader'
```
chriskchew, line 5: Does this make serialize an unused import?

brianc: ah yeah this file is trash I'm just using during development, good thinking though.  After I merge your pool PR I'm going to turn on significant linting & prettier on the whole codebase.  

1) some of the code, particularly tests, in the test folder, is older than the hills.  It's mostly valuable but hasn't been linted or prettiered as both eslint and prettier didn't exist when it was written, so been putting it off
2) this repo recently became a mono-repo from several other projects, each w/ slightly different lint rules
3) been busy w/ covid anxiety!

But anyways, yeah huge PR incomming tomorrow (I think tomorrow) to lint everything. until then...chaos reigns (so long as tests pass)

PR #2161, opened 2020-04-10
Prettier the codebase

brianc:
Bring all the formatting under one roof.  I set up eslint w/ a pretty bare-bones config file that only enforces prettier run on everything.  I'll likely add more rules when I find something lacking; however, some rules require manual fixing.  I am keeping this PR isolated to just adding eslint and auto-formatting the codebase w/ the prettier plugin.  This way all code changes here (outside of eslint config changes) were done automatically. 

I'm going to merge this quick as it's a timebomb of merge conflicts.


brianc:
mmm good times, prettier@2.x doesn't support node@8.x so downgrading prettier to 1.x.

charmander:
Hmm, what does prettier 2.x’s output try to use that’s incompatible with Node 8?

brianc:
`String.prototype.trimEnd` for starters. It also specifies "engines >= 10" in their package.json so installs fail.

https://travis-ci.org/github/brianc/node-postgres/jobs/673478849

I could try making it an optional dependency and only lint in that situation.  Lemme let the current test suite run and then I'll try that.  I'd rather use newer versions.

brianc:
Allllrighty, re-upgraded, make prettier an optional dep, and conditionally bypass lint if it doesn't exist.  That's fine since code really only needs to be linted once in the build cycle, not every build.

brianc:
@charmander thanks for pushing me to just use newer prettier & not lint in node8! Much better. ❤️

Issue #2170, opened 2020-04-21
NodeJS v14 support


vitaly-t:
Having plenty of troubles running tests, and thus not sure how well this library works under the newly released NodeJS v14. It seems to have some problems there.

gbrunacci:
`pg` package was not trowing any error for me, it just ends execution. I was able to spot a `Segmentation Fault` while trying to connect with `pg-native`

This piece of code
```
const { Client } = require('pg').native
const client = new Client({
    host: 'myhost',
    user: 'threetier',
    database: 'threetier',
    password: 'threetier',
    port: 5432
});

client.on('error', err => {
  console.error('something bad has happened!', err.stack)
})

client.on('notice', msg => console.warn('notice:', msg))


const f = async () => {
    try {
        console.log('connecting...');
        await client.connect();
        console.log('querying');
        const res = await client.query('SELECT NOW() as t')
        console.log(res.rows[0].t);
        await client.end()
    } catch (e) {
        console.log('failed e', e);
    }
}

console.log('calling');
f();
```

Throw the following error:

```
[ec2-user@ip-10-0-100-72 p]$ gdb node
GNU gdb (GDB) Red Hat Enterprise Linux 8.0.1-30.amzn2.0.3
Copyright (C) 2017 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-redhat-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from node...done.
(gdb) set args test.js
(gdb) run
Starting program: /home/ec2-user/.nvm/versions/node/v14.0.0/bin/node test.js
Missing separate debuginfos, use: debuginfo-install glibc-2.26-34.amzn2.x86_64
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
[New Thread 0x7ffff6d33700 (LWP 22770)]
[New Thread 0x7ffff6532700 (LWP 22771)]
[...]

PR #2171, opened 2020-04-22
Replace uses of private/undocumented `readyState` API

charmander:
The `readyState` of a newly-created `net.Socket` changed from `'closed'` to `'open'` in Node 14.0.0, so this makes the JS driver work on Node 14.

`Connection` now always calls `connect` on its `stream` when `connect` is called on it.

#2170

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ node_js:
   # if 13.8 still has this problem when it comes down I'll talk to the node team about the change
   # in the mean time...peg to 13.6 
   - 13.6
+  - 14
 
 addons:
   postgresql: "10"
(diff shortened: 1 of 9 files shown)


charmander:
Oh, CI’s passing, so it looks like the native driver wasn’t broken?

brianc:
Thank you again! I can roll out a patch version for this here in an hour or so. 👯

brianc:
released fix:

```
Successfully published:
 - pg-cursor@2.1.10
 - pg-pool@3.1.1
 - pg-protocol@1.2.2
 - pg-query-stream@3.0.7
 - pg@8.0.3
```

Romstar:
I'm still experiencing this with node 14.16.1 and pg 8.6.  Any advice?

PR #2194, opened 2020-05-05
Fix repository field in package.json files after packages were moved into monorepo

papandreou:
Noticed this because of an automated PR for updating pg-connection-string: 2.2.0 → 2.2.1 linked me to the wrong repository.


brianc:
Thanks!

PR #2198, opened 2020-05-05
Add no-verify ssl option

brianc:
This adds no-verify ssl option to

1) connection string parsing.  You can pass `?ssl=no-verify` on the pg connection string & it will properly use `{ rejectUnauthorized: false }` in the ssl config passed to `tls.connect`.  Connecting with connection strings is common in heroku, and without this patch the connection info cannot be entirely contained within the connection string for heroku environments.

2) environment variables.  added support for `PGSSLMODE=no-verify`.

3) manually passing `new Client({ ssl: 'no-verify' })`.  The 3rd option isn't very common, but it _will_ technically work.

This is a backwards compatible, semver minor change.

Thanks to @benhjames for takin' the lead on this one 😄

### DIFF
--- a/packages/pg/lib/connection-parameters.js
+++ b/packages/pg/lib/connection-parameters.js
@@ -25,9 +25,11 @@ var val = function (key, config, envVar) {
   return config[key] || envVar || defaults[key]
 }
 
-var useSsl = function (modeFromConfig) {
+var normalizeSSLConfig = function (modeFromConfig) {
   // if the ssl parameter passed to config is not a string, just return it
   // directly (it will be passed directly to tls.connect)
+  // this way you can pass all the ssl params in via constructor:
+  // new Client({ ssl: { minDHSize: 1024 } }) etc
   if (modeFromConfig !== undefined && typeof modeFromConfig !== 'string') {
     return modeFromConfig
   }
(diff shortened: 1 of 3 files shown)


packages/pg/lib/connection-parameters.js
```diff
@@ -25,15 +25,31 @@ var val = function (key, config, envVar) {
...
+
+  switch (mode) {
     case 'disable':
       return false
     case 'prefer':
     case 'require':
     case 'verify-ca':
     case 'verify-full':
       return true
+    // no-verify is not standard to libpq but allows specifying
```
boromisp, line 46: If the intent is to follow the libpq behavior, then `prefer` should also skip certificate validation, and `require` only needs to validate the certificate if the root CA file is available.
Also, the `checkServerIdentity` should be disabled (set to no-op) if the mode is not `verify-full`.

brianc: Yeah totally...but I'm probably not gonna change these - it's a semver major to change these and they've behaved this way since...kinda forever w/ `pg`.  Slight inconsistencies w/ the native bindings aren't ideal, but they're less not-ideal than breaking changes.  I will do a better job of documenting these options though in the docs and what they mean and how they diverge.

PR #2208, opened 2020-05-09
Add some SCRAM tests and enable them on travis

sehrope:
First commit adds another test helper. Engines is node 8+ so async all the way!

Second commit adds a test for SCRAM auth. It's disabled by default. To run the test export SCRAM_TEST_PGUSER and SCRAM_TEST_PGPASSWORD env vars. The rest of the vars will inherit from the default PG% config.

Third commit enables the SCRAM test in it's own matrix build. Could probably clean this up later to piggy back on an existing combo but for now I didn't want to break anything. Also, to check if we're actually using SCRAM it adds listeners to the underlying connection so test only runs on pure-JS for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ matrix:
     # Run tests/paths that require password authentication
     - node_js: lts/erbium
       env:
-        - CC=clang CXX=clang++ npm_config_clang=1 PGUSER=postgres PGDATABASE=postgres PGPASSWORD=test-password
+        - CC=clang CXX=clang++ npm_config_clang=1 PGUSER=postgres PGDATABASE=postgres PGPASSWORD=test-password SCRAM_TEST_PGUSER=scram_test SCRAM_TEST_PGPASSWORD=test4scram
       before_script: |
         sudo -u postgres sed -i \
           -e '/^local/ s/trust$/peer/' \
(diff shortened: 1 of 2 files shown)


packages/pg/test/test-helper.js
```diff
@@ -125,6 +125,28 @@ assert.throws = function (offender) {
...
+/**
+ * Executes a function that should return a Promise that will reject.
+ *
+ * If the Promise resolves then fail.
+ * If the Promise does not reject within timeoutMillis then fail.
+ * If the Promise rejects then __return__ the Error
+ *
+ * NOTE: The error is returned so the caller can perform additional validation
+ */
+assert.rejects = function (offender, timeoutMillis) {
```
charmander, line 137: Since `assert.rejects` is an existing builtin but behaves differently, I think this should be called something else. Or `assert.rejects(() => client.connect, {code: '28P01'})`.

sehrope: Ha! I knew I couldn't have come up with a name that good on my own.

I'll take a look tomorrow at which is available going back to v8.x and adjust accordingly. Thanks for the review!

sehrope:
Rebased and updated to remove the extra test helper commit as, per @charmander, it already exists in built-in asserts package.

brianc:
this is wonderful, thank you!

Issue #2204, opened 2020-05-08
Encoding situation


Ygarasab:
Hey, I'm working on a litle api that is using node-postgres, but i'm having some encoding issues

I've made a query through a Client, and got the following response:

```
[
  { value: 2, text: 'cent¡metros' },
  { value: 1, text: 'metros' },
  { value: 5, text: 'pacotÆos' },
  { value: 4, text: 'pacotes' },
  { value: 3, text: 'unidades' }
]
```
So, i've got ¡ instead of í and Æ instead of ã.

I wondered if it was a pg issue but the database cli showed the correct carachters.

I'm brazilian, so this sort of character will be quite common to enter the database, and I really don't know how to solve that :/

vitaly-t:
Looks like your client encoding isn't set right, which is not relevant to this library.

Tubskleyson:
Oh, i tried set it like this:

```
new Client({
            user: 'myuser',
            host: 'localhost',
            database : 'mydb',
            password : 'topsecretpassword',
            client_encoding: 'utf8',
            port : 5432
        })
```

Isn't utf8 the right encoding? Or should I set it up other way?

Tubskleyson:
Sorry for that, got a little bit confused with the mobile aplication buttons

Still need help c:

Ygarasab:
Hey, I figured ou what happened

So, it turns out that the problem was in the windows prompt encoding, you see, I did some entries there, and they looked just fine in the terminal, but when i looked at it anywhere else, the encoding just seemed outta place. So I tried making some entries with especial carachter through the API I was doing, and when I tried to get then the encoding was OK.

So yeah, no issue with the module and power shell == bad

Issue #2203, opened 2020-05-07
Select not returning when using accentuation


LuizGMX:
I'm trying to use a word with accentuation in my WHERE clause but it's not working.
And when I retrieve data that cotains a word with accentuation i always get "�" in place.

charmander:
Please show the relevant code and output, and, if possible, a minimal script to reproduce the problem.

LuizGMX:
`router.get("/filtraEspecialidade", function (req, res) {    

    async function listaEspecialidades() {

        await db.connect()

        var result;
        result = await db.query(`
    SELECT te.des_especialidade 
    FROM arq_medico am
    INNER JOIN arq_medico_especialidade ame ON ame.id_medico = am.id_medico
    INNER JOIN tab_especialidade te ON te.id_especialidade = ame.id_especialidade
    WHERE am.ind_status = 'Ativo' AND EXISTS (SELECT 1 FROM arq_agendap WHERE am.id_medico = arq_agendap.id_medico LIMIT 1) 
`)        
        res.send(result.rows);

    }
    listaEspecialidades();

});`

charmander:
Where are the accents? Where is the output?

Anyway, respond with the encoding you’re sending:

```js
res.set('Content-Type', 'text/plain; charset=utf-8')
```

LuizGMX:
Here's the output
![image](https://user-images.githubusercontent.com/44658199/81438761-ab3d9e80-9143-11ea-9757-482d0b41e043.png)

LuizGMX:
![image](https://user-images.githubusercontent.com/44658199/81440916-34a2a000-9147-11ea-8d51-42b31a8f577e.png)

LuizGMX:
Any help?

charmander:
Can you show the results of that query using another tool, like psql? Are they correct in that case?

LuizGMX:
yes it's correct, see: 
![image](https://user-images.githubusercontent.com/44658199/81482905-b27eae00-9210-11ea-9f08-9727f0f2afcc.png)

LuizGMX:
Hi, can someone help me?

boromisp:
Please execute the following queries both in node.js, and in the environment that returns the expected characters (pgAdmin?) and show us the results.

```sql
SHOW client_encoding;
SHOW server_encoding;
```

LuizGMX:
[...]

PR #2215, opened 2020-05-13
Latin 1 client encoding support for the pure JavaScript client

boromisp:
This PR fixes the handling of the `client_encoding` parameter in the pure JavaScript client.
The intersection of the natively supported text encodings of Node.js and Postgres is UTF-8 and Latin 1.

The main (only?) use-case for this is compatibility with existing `SQL_ASCII` databases containing `LATIN1` encoded text.

I've could not get the native client working.

Related issues: #2203 (#2204 unconfirmed)


packages/pg-protocol/src/serializer.ts
```diff
@@ -197,40 +198,40 @@ type PortalOpts = {
   name?: string
 }
 
-const cstringMessage = (code: code, string: string): Buffer => {
+const cstringMessage = (code: code, string: string, encoding: TextEncoding): Buffer => {
```
brianc, line 201: nit: would be nice to have ` = 'utf8'` as the default value here for consistency with the other functions.

packages/pg/test/integration/connection/test-helper.js
```diff
@@ -6,7 +6,11 @@ var utils = require(__dirname + '/../../../lib/utils')
 var connect = function (callback) {
   var username = helper.args.user
   var database = helper.args.database
-  var con = new Connection({ stream: new net.Stream() })
+  var client_encoding = helper.args.client_encoding
```
brianc, line 9: I see you're passing this in here which is cool, but were you intending on setting this in CI to test both permutations?  AFAIK this will always still default to `utf-8` for now?

brianc:
whoah that was super fast! thanks!  LMK if there's anything else you want me to take a look at or any additional tests you'd like me to write.  I left a touch of feedback on there but this is fabulous. 👯

packages/pg/test/integration/connection/test-helper.js
boromisp: I haven't considered how to test it. Would it be better to write a few targeted integration and unit tests instead of running all the tests with different client encodings?

boromisp:
@brianc I've made a couple of improvements:
- Validate that the chosen `client_encoding` is supported by the library (`UTF8` and `LATIN1` with [the JavaScript](https://github.com/boromisp/node-postgres/blob/ec2da254d4612c1fda06c1284924c98d3bbb906d/packages/pg/lib/connection.js#L29) and only `UTF8` with [the native client](https://github.com/boromisp/node-postgres/blob/ec2da254d4612c1fda06c1284924c98d3bbb906d/packages/pg/lib/native/client.js#L47:L53))
- Detect `client_encoding` changes and use the new encoding for parsing and serializing messages
- pg-protocol encoding unit tests

There is a known bug in the current implementation:

If the user sets the `client_encoding` with a query to an unsupported value (eg. `'SET client_encoding TO LATIN2'`), then when the `Parser` intercepts the `ParameterStatus` message and tries to [change its encoding](https://github.com/boromisp/node-postgres/blob/ec2da254d4612c1fda06c1284924c98d3bbb906d/packages/pg-protocol/src/parser.ts#L255), it will throw an error.

The simplest way to handle this error would be to terminate the connection, but I'm not sure how to propagate this error to the `Connection`.


As an aside, since the only use-case for this code seems to be supporting misconfigured databases (`SQL_ASCII` `server_encoding` with non-ASCII text) it might be worth considering dropping all this code and simply enforcing UTF8 encoding with errors and documentation. That's what the JDBC driver is doing.

brianc:

packages/pg/test/integration/connection/test-helper.js
brianc: > Would it be better to write a few targeted integration and unit tests instead of running all the tests with different client encodings?

yeah i think so - running the whole test suite again just to test a different encoding is overkill IMO

Issue #2103, opened 2020-02-12
native driver ignores statement_timeout


gajus:
```js
const pool = new pg.Pool({
  statement_timeout: 1000,
});

```

Passing `statement_timeout` to node-postgres when using `pg-native` has no effect on the `statement_timeout` configuration.

brianc:
This might be a pretty easy fix if you're interested in doing a PR.  Otherwise I'll get to it soon-ish but I've got a bunch of other stuff in the pipeline too. 🙏

gajus:
Switching to the JavaScript driver now throws:

```
      error: unsupported startup parameter: statement_timeout
          at Connection.parseE (/Users/gajus/Documents/dev/rayroute/rayman/node_modules/pg/lib/connection.js:614:13)
          at Connection.parseMessage (/Users/gajus/Documents/dev/rayroute/rayman/node_modules/pg/lib/connection.js:413:19)
          at TLSSocket.<anonymous> (/Users/gajus/Documents/dev/rayroute/rayman/node_modules/pg/lib/connection.js:129:22)
          at TLSSocket.emit (events.js:321:20)
          at TLSSocket.EventEmitter.emit (domain.js:485:12)
          at addChunk (_stream_readable.js:297:12)
          at readableAddChunk (_stream_readable.js:273:9)
          at TLSSocket.Readable.push (_stream_readable.js:214:10)
          at TLSWrap.onStreamRead (internal/stream_base_commons.js:186:23)

```

boromisp:
@gajus Could the `unsupported startup parameter` error be related to this: https://www.pgbouncer.org/config.html#ignore_startup_parameters?

shayonj:
At @loomhq we are looking to see if we can using this in our sequelize. I noticed `statement_timeout` doesn't get applied with the native driver. proposed a PR to fix in https://github.com/brianc/node-postgres/pull/2323. Added support for `idle_in_transaction_session_timeout` as well

shayonj:
Re: 
>       error: unsupported startup parameter: statement_timeout

This is expected from pgbouncer, because `statement_timeout` is not part of `ignore_startup_parameters` (pgbouncer's config). However, when the fix ships it will complain on `options` (since we do `options=-c statement_timeout=...`). 

The issue here is pgbouncer ignores startup parameter once added to `ignore_startup_parameters` (it only does so that pgbouncer doesn't hard fail), so passing `statement_timeout` via `options` to `pgbouncer` won't work I believe. 

[...]

PR #2216, opened 2020-05-13
Support options connection parameter

### LINKED ISSUE #2214: Support options/PGOPTIONS connection parameter/environment variable
Postgres describes this parameter here: https://www.postgresql.org/docs/9.1/libpq-connect.html#LIBPQ-CONNECT-OPTIONS. This driver should support passing through the parameter, or reading it from an environment variable.

This was previously requested in this issue: https://github.com/brianc/node-postgres/issues/1482 I believe that issue was closed mistakenly. Here is [the Postgres source code](https://github.com/postgres/postgres/blob/7dd777938bbeae8113f73849920a5b19bef723d9/src/interfaces/libpq/fe-connect.c#L253) that looks at this parameter.

It looks like the fix would be in the `ConnectionParameters` object. I am happy to create a PR to addrss this issue.
https://github.com/brianc/node-postgres/blob/70cf4dc6edbe1e4d5b6e4ddbe9ebaabc656ff738/packages/pg/lib/connection-parameters.js#L121-L126

rafiss:
This supports the connection parameter documented here:
https://www.postgresql.org/docs/9.1/libpq-connect.html#LIBPQ-CONNECT-OPTIONS

fixes #2214

### DIFF
--- a/packages/pg-connection-string/index.d.ts
+++ b/packages/pg-connection-string/index.d.ts
@@ -11,4 +11,5 @@ export interface ConnectionOptions {
 
   application_name?: string
   fallback_application_name?: string
+  options?: string
 }
--- a/packages/pg-connection-string/test/parse.js
+++ b/packages/pg-connection-string/test/parse.js
@@ -188,10 +188,10 @@ describe('parse', function () {
     subject.fallback_application_name.should.equal('TheAppFallback')
   })
 
-  it('configuration parameter fallback_application_name', function () {
-    var connectionString = 'pg:///?fallback_application_name=TheAppFallback'
+  it('configuration parameter options', function () {
+    var connectionString = 'pg:///?options=-c geqo=off'
     var subject = parse(connectionString)
-    subject.fallback_application_name.should.equal('TheAppFallback')
+    subject.options.should.equal('-c geqo=off')
   })
 
   it('configuration parameter ssl=true', function () {
--- a/packages/pg/lib/client.js
+++ b/packages/pg/lib/client.js
@@ -391,6 +391,9 @@ Client.prototype.getStartupConf = function () {
   if (params.idle_in_transaction_session_timeout) {
     data.idle_in_transaction_session_timeout = String(parseInt(params.idle_in_transaction_session_timeout, 10))
   }
+  if (params.options) {
+    data.options = params.options
+  }
 
   return data
 }
--- a/packages/pg/lib/connection-parameters.js
+++ b/packages/pg/lib/connection-parameters.js
@@ -70,6 +70,7 @@ var ConnectionParameters = function (config) {
   })
 
   this.binary = val('binary', config)
+  this.options = val('options', config)
 
   this.ssl = typeof config.ssl === 'undefined' ? readSSLConfigFromEnvironment() : config.ssl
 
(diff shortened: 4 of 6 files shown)


brianc:
wow that was fast! Do you know of a param I could send to the backend & then query the backend to verify it's taken effect?  I'd like to add a test that does something like...

```js
const client = new Client({ options: '--foo=bar' })
await client.connect()
const { rows } = await client.query('SELECT session_options')
assert(rows[0].key === 'foo')
assert(rows[0].value === 'bar')
```

or something like that.  The specifics aren't really ironed out there I just always like to test "round tripping" the data when I can.  I'll write this test if you know of how it could work?

sehrope:
@brianc How about `default_transaction_isolation`? I have a feeling that one won't be going away:

```
$ psql \
    -c 'SHOW default_transaction_isolation'
 default_transaction_isolation 
-------------------------------
 read committed
(1 row)

$ PGOPTIONS='--default_transaction_isolation=serializable' psql \
    -c 'SHOW default_transaction_isolation'
 default_transaction_isolation 
-------------------------------
 serializable
(1 row)
```

boromisp:
With recent postgres versions you can use custom namespaced variables:

```bash
PGOPTIONS="-c example.foo=bar" psql -c "SHOW example.foo"
 example.foo
-------------
 bar
(1 row)

psql -c "SHOW example.foo"
ERROR:  unrecognized configuration parameter "example.foo"
```

rafiss:
Doing a round-trip test definitely sounds good to me. And thanks for taking on the task of writing it!

I'd agree with using a custom namespaced variable like how @boromisp posted. I think the `-c name=value` syntax is needed to set it.

boromisp:
I think you will have to parse these parameters and [include them in the startup message](https://github.com/brianc/node-postgres/blob/bf469399b88bcdf86eff096fd0dd05684adc1117/packages/pg/lib/client.js#L373) for the JavaScript client.

rafiss:
> I think you will have to parse these parameters and [include them in the startup message](https://github.com/brianc/node-postgres/blob/bf469399b88bcdf86eff096fd0dd05684adc1117/packages/pg/lib/client.js#L373) for the JavaScript client.

Ah thanks, I'd missed that earlier. Just updated.

boromisp:
The https://github.com/brianc/node-postgres/issues/2103 issue could be fixed by setting the `statement_timeout` for the native driver in the `options` parameter:

```js
var escapeOptionValue = function (value) {
  return ('' + value).replace(/\\/g, '\\\\').replace(/ /g, '\\ ')
}

var addOption = function (options, config, optionName) {
  var value = config[optionName]
  if (value !== undefined && value !== null) {
    options.push('-c ' + optionName + '=' + escapeOptionValue(value))
  }
}

ConnectionParameters.prototype.getLibpqConnectionString = function (cb) {
  // ...
  var options = []
  if (this.options) {
    options.push(this.options)
  }
  addOption(options, this, 'statement_timeout')
  if (options.length > 0) {
    params.push('options=' + quoteParamValue(options.join(' ')))
  }
  // ...
}
```

An other (undocumented) config parameter that doesn't work with the native driver is `idle_in_transaction_session_timeout`. This could also be fixed the same way.

(The `unsupported startup parameter: statement_timeout` error in #2103 is probably related to this: https://www.pgbouncer.org/config.html#ignore_startup_parameters)

brianc:

brianc:
@rafiss I have included a patch file for an integration test

https://cloudup.com/cHnXN0tuoD6

If you apply that patch to your branch I can merge this & release a new minor version w/ the support added.  Thanks for your work here & sorry for the delay!

brianc:
or actually i can just merge this & then add the test myself. :)

brianc:
Thanks for the PR! I'll release this tomorrow morning so I can make sure it doesn't cause weird regressions in anyone's systems.  (it shouldn't...but you never know)

rafiss:
Thank you @brianc! Let me know if anything comes up.

brianc:
Released as `pg@8.3.0`

PR #2217, opened 2020-05-14
Send the `client_encoding` startup parameter value with more typical formatting

charmander:
All non-alphanumerics are ignored, but `'utf-8'` implies a different syntax. `UTF8` is the canonical name, and is what libpq sends.


brianc:
👍

Issue #2240, opened 2020-06-17
Major performance issues with bytea performance


regevbr:
Hi,

I encountered the exact same issue as #1286 wheres a simple query to retrieve a 100mb bytea column from a one table, one row DB takes forever and causes the cpu to spike at 100%, rendering my node instance unresponsive.

After investigating, I think I found the cause in the `pg-protocol` package https://github.com/brianc/node-postgres/blob/ea6ac2ad2313af57159b10a0292c0c178e8e0923/packages/pg-protocol/src/parser.ts#L91-L93

Which suffers from the same thing as the original issue - too many allocations and buffer copies going around. I believe the fix introduced back then in https://github.com/brianc/node-packet-reader/pull/3 should solve the problem here.

lognaturel:
Just started running into this and feeling very grateful to see it’s been fixed for a while now already! Thank you, @regevbr, @brianc. 😊

Issue #1286, opened 2017-05-10
Problems with bytea performance


mramato:
I originally opened this as part of a `knex` issue: https://github.com/tgriesser/knex/issues/2052 but was directed here instead.

I have a simple query:

```SQL
SELECT glb FROM models where assets_id = 16 LIMIT 1;
```

glb is a `bytea` column and the value I'm retrieving is `28630077` bytes (~27MB) (models contains a single row in this example).  The query takes `13305 ms` to run and the Node process (not the DB process) maxes out the CPU while the query is running). If I query for the `assets_id` column instead of the `glb` column, it only takes `2 ms`.

Running the same query with the same data from the psql command line completes almost immediately:

```bash
time psql -A -c "SELECT glb FROM models where assets_id = 16 LIMIT 1;" master postgres > out.glb

real    0m0.679s
user    0m0.000s
sys     0m0.031s
```

I also tested the same query in `pg-native` and it completed in `~450ms`, but using `pg-native` isn't an option for me at this time (though I might have to re-evaluate that depending on where this issue goes).

Here's the table definition for completeness.

```SQL
CREATE TABLE public.models
(
  assets_id integer NOT NULL,
  glb bytea NOT NULL,
  CONSTRAINT models_pkey PRIMARY KEY (assets_id),
  CONSTRAINT models_assets_id_foreign FOREIGN KEY (assets_id)
      REFERENCES public.assets (id) MATCH SIMPLE
      ON UPDATE NO ACTION ON DELETE CASCADE
)
WITH (
  OIDS=FALSE
);
ALTER TABLE public.models
  OWNER TO postgres;
```

Finally, I thought maybe it was a performance issue in the type parser, but all of the time is taken up by the query and then the typeParser completes almost instantly at the end.

Am I doing something wrong?  Or is there a performance issue with `bytea` issues?  I'd be happy to debug this further myself if someone can point me in the correct direction.  Thanks in advance.

vitaly-t:
> I'd be happy to debug this further myself if someone can point me in the correct direction

Please do, many people would want to know :wink:

vitaly-t:
This caught my interest, so I did some quick local testing, and I can confirm the following results...

Reading a single `bytes` field that contains a 16MB file:

* With JavaScript binding: 4ms
* With Native Binding: 200ms

mramato:
> On second thought though, since use of the pg-native made such a huge difference, means it is not a pg-types issue, more like inside the Connection object.

[...]

PR #2241, opened 2020-06-18
fix: major performance issues with bytea performance #2240

regevbr:
fix: #2240

a simple query to retrieve a 100mb bytea column from a one table, one row DB takes forever and causes the cpu to spike at 100%, rendering my node instance unresponsive.

The issue is exactly like #1286 and the fix is +- the same as  https://github.com/brianc/node-packet-reader/pull/3

The code at:
https://github.com/brianc/node-postgres/blob/ea6ac2ad2313af57159b10a0292c0c178e8e0923/packages/pg-protocol/src/parser.ts#L91-L93
suffers from too many allocations and buffer copies going around.

The fix is a bit messy but it gets the job done.

Before the fix, fetching 100mb of bytea column took 3 minutes and 10 seconds and consumed on average 95% cpu usage during the entire time

After the fix the query took only! 40 seconds and consumed on average 3% cpu

### DIFF
--- a/packages/pg-protocol/src/parser.ts
+++ b/packages/pg-protocol/src/parser.ts
@@ -96,11 +96,20 @@ export class Parser {
       const newLength = this.remainingBufferLength + combinedBufferLength
       const newFullLength = newLength + this.remainingBufferOffset
       if (newFullLength > this.remainingBuffer.byteLength) {
-        let newBufferLength = this.remainingBuffer.byteLength * 2
-        while (newLength >= newBufferLength) {
-          newBufferLength *= 2
+        // We can't concat the new buffer with the remaining one
+        let newBuffer: Buffer
+        if (newLength <= this.remainingBuffer.byteLength && this.remainingBufferOffset >= this.remainingBufferLength) {
+          // We can move the relevant part to the beginning of the buffer instead of allocating a new buffer
+          newBuffer = this.remainingBuffer
+        } else {
+          // Allocate a new larger buffer
+          let newBufferLength = this.remainingBuffer.byteLength * 2
+          while (newLength >= newBufferLength) {
+            newBufferLength *= 2
+          }
+          newBuffer = Buffer.allocUnsafe(newBufferLength)
         }
-        const newBuffer = Buffer.allocUnsafe(newBufferLength)
+        // Move the remaining buffer to the new one
         this.remainingBuffer.copy(
           newBuffer,
           0,
@@ -110,6 +119,7 @@ export class Parser {
         this.remainingBuffer = newBuffer
         this.remainingBufferOffset = 0
       }
+      // Concat the new buffer with the remaining one
       buffer.copy(this.remainingBuffer, this.remainingBufferOffset + this.remainingBufferLength)
       combinedBuffer = this.remainingBuffer
       combinedBufferLength = this.remainingBufferLength = newLength
@@ -134,16 +144,18 @@ export class Parser {
         break
       }
     }
-
     if (offset === fullLength) {
+      // No more use for the buffer
       this.remainingBuffer = emptyBuffer
       this.remainingBufferLength = 0
       this.remainingBufferOffset = 0
     } else {
       if (reuseRemainingBuffer) {
+        // Adjust the cursors of remainingBuffer
         this.remainingBufferLength = combinedBufferLength - offset
         this.remainingBufferOffset += offset
       } else {
+        // To avoid side effects, copy the remaining part of the new buffer to remainingBuffer
         this.remainingBuffer = combinedBuffer.slice(offset)
         this.remainingBufferLength = this.remainingBuffer.byteLength
         this.remainingBufferOffset = 0
--- a/packages/pg-protocol/src/parser.ts
+++ b/packages/pg-protocol/src/parser.ts
@@ -73,10 +73,18 @@ const enum MessageCodes {
 
 export type MessageCallback = (msg: BackendMessage) => void
 
+interface CombinedBuffer {
+  combinedBuffer: Buffer
+  combinedBufferOffset: number
+  combinedBufferLength: number
+  combinedBufferFullLength: number
+  reuseRemainingBuffer: boolean
+}
+
 export class Parser {
-  private buffer: Buffer = emptyBuffer
-  private bufferLength: number = 0
-  private bufferOffset: number = 0
+  private remainingBuffer: Buffer = emptyBuffer
+  private remainingBufferLength: number = 0
+  private remainingBufferOffset: number = 0
   private reader = new BufferReader()
   private mode: Mode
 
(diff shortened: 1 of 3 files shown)


brianc:
ah wow thanks - i'll check this out now & get it pushed out

regevbr:
@brianc no prob, a review now will be welcome, but some tests are failing to it is not ready to be merged just yet. I will ping you when it is ready

brianc:
at first glance this may have regressed performance in other places - travis is failing on 'big-simple-query-tests'

brianc:
ah gotcha - i'll let you pound on it a bit more than & check back l8r!

regevbr:
@brianc that sounds great!

regevbr:
@brianc I fixed the bug and the CI is passing now :-)
Can you please review?

regevbr:
I made some more improvements and refactoring and added a few comments. It is now really ready for review

packages/pg-protocol/src/parser.ts
```diff
@@ -102,19 +113,94 @@ export class Parser {
...
+      this.remainingBufferLength = 0
+      this.remainingBufferOffset = 0
     } else {
-      this.remainingBuffer = combinedBuffer.slice(offset)
+      this.remainingBufferLength = combinedBufferFullLength - combinedBufferOffset
+      if (reuseRemainingBuffer) {
+        // Adjust the cursors of remainingBuffer
+        this.remainingBufferOffset = combinedBufferOffset
+      } else {
+        // To avoid side effects, copy the remaining part of the new buffer to remainingBuffer with extra space for next buffer
```
regevbr, line 199: @brianc this part is the only thing I think I can still improve. The question is if I can use the input buffer for the next run without anyone modifying it out of scope. What do you think?

regevbr:
I uses a nice tool called [clinic.js](https://clinicjs.org/) to visualize the before and after of this PR.
Before:
before-doctor
After:
after-doctor

Before:
before-flames
After:
after-flames

regevbr:
@brianc did you have a chance to review my PR?

regevbr:
@brianc @charmander can you please CR?

brianc:
I'm sorry for the delay. I've been having some anxiety attacks related to covid (I live in Austin, the #1 worst city in the US for new cases) & I've been really slacking on this work trying to keep mental health intact as best I can.  I'll take a look at this now...I really appreicate the work.

regevbr:
Oh bummer, I'm sorry you have to go through that! I hope you will overcome it soon :-)
I don't want to impose any more pressure on you so don't feel obliged to work on it if you don't feel like it.
I trust that when everything will be back to normal for you, you will find the time to look it over.

packages/pg-protocol/src/parser.ts
brianc: what do you mean by anyone modifying it out of scope? The only place in the whole codebase where we deal w/ raw input buffers from the socket is here in `parser.ts` and its not exposed to the user ever so any type of modifications you do here are safe in that no one else should (or can AFAIK) ever modify input buffers from the socket.

brianc:
I looked over this - it looks pretty good!  I ran my crude benchmarks too & perf doesn't seem to be impacted.  Is there any way to write any kind of test for this?  It's a great change, I just worry w/o tests it might regress in the future (as this currently already was kind of a regression when I moved away from buffer-reader since this lib didn't contain any tests that stress dealing w/ big bytea returns)

packages/pg-protocol/src/parser.ts
```diff
@@ -102,19 +113,94 @@ export class Parser {
...
         break
       }
     }
+    this.consumeBuffer({
+      combinedBuffer,
+      combinedBufferOffset: offset,
+      combinedBufferLength,
+      reuseRemainingBuffer,
+      combinedBufferFullLength,
+    })
```
brianc, line 130: I have a question for you....a long time ago it was more efficient to make "backing classes" for things you'd return frequently as v8 can optimize them much more aggressively.  You are using a type `CombinedBuffer` and constructing it w/ the object shorthand here & returning it w/ the object shorthand on line `172`.  Do you think it'd make a difference to actually make this a class?
something like this...
```
class CombinedBuffer {
  constructor(public combinedBuffer: buffer = combinedBuffer, public combinedBufferOffset: number = offset, public combinedBufferLength: number = length, public reuseRemainingBuffer: boolean = reuse, public combinedBufferFullLength: number = fullLength) {}
}
```
and then return it with `this.consumeBuffer(new CombinedBuffer(combinedBuffer, offset, combinedBufferLength, reuseRemainingBuffer, combinedBufferFullLength))`

? 

Staying on top of how v8 optimizes things is a moving target so I'm not sure if you have any experience around the perf diff of using dynamically created objects versus proper "classes."  I wouldn't normally sweat it but this is a white code code path & it might be a small perf win here.

brianc: You might also be able to allocate a single instance of `CombinedBuffer` on the parser class itself & just set the properties on it versus allocating new ones every time.   Again, a micro-optimization but those things can add up when the code is called so frequently. 

I can play around w/ this too & push to your branch (I think github lets me do that?) if you're cool w/ it & think it's an okay idea.

packages/pg-protocol/src/parser.ts
regevbr: It will be better here to use the existing input buffer for the next run instead of copying it over. What you say is that it is safe for me to do (as far as you know) which is great! I will change it then.
I was worried that who ever passes that buffer, can decide themselves to resuse that buffer and then it will mess up with my code...

packages/pg-protocol/src/parser.ts
regevbr: That is an interesting solution, why don't you let me research it (as I'm not too familiar with v8 bits) and try to enhance it

regevbr:
> I looked over this - it looks pretty good! I ran my crude benchmarks too & perf doesn't seem to be impacted. Is there any way to write any kind of test for this? It's a great change, I just worry w/o tests it might regress in the future (as this currently already was kind of a regression when I moved away from buffer-reader since this lib didn't contain any tests that stress dealing w/ big bytea returns)

I'm glad to hear that! thanks!
Do you have a test scenario in place? I believe it will be a bit complicated as you will need to stress test it and make sure that the cpu doesn't get hogged.
I will try to add a test and see if I manage to get it to fail with the current code

packages/pg-protocol/src/parser.ts
brianc: the buffer is passed in [here](https://github.com/brianc/node-postgres/blob/master/packages/pg/lib/connection.js#L104) as the whole `net.Socket`

The only place I could imagine this might cause issues is w/ some of the tests which may incorrectly be reusing buffers though nothing comes to mind.  Some of the tests are old and kinda wonky so you could just try & run the tests...as long as they pass we're definitely good. The raw buffers of the socket are never exposed anywhere outside of the parser.

regevbr: Great will do

brianc:
> I will try to add a test and see if I manage to get it to fail with the current code

Yeah if you can't then I understand.  It's incredibly difficult to test how much CPU some code is using within a unit test.  It'd be ideal to have some way to catch a performance regression.  You could add another benchmark to [this crude benchmark file](https://github.com/brianc/node-postgres/blob/master/packages/pg/bench.js).  I run that before any merge & manually check for perf regressions.

regevbr:
> > I will try to add a test and see if I manage to get it to fail with the current code
> 
> Yeah if you can't then I understand. It's incredibly difficult to test how much CPU some code is using within a unit test. It'd be ideal to have some way to catch a performance regression. You could add another benchmark to [this crude benchmark file](https://github.com/brianc/node-postgres/blob/master/packages/pg/bench.js). I run that before any merge & manually check for perf regressions.

Ok will check it out, thanks!

regevbr:
@brianc I started reading about the performance difference between object literal and classes (or more specifically constructor functions) and indeed back in 2014 there was a difference, but it doesn't exist anymore...
For example you can look at [this](https://stackoverflow.com/questions/22057579/why-is-calling-a-method-on-an-object-literal-slower-on-v8) post but if you run the [jsperf test](https://jsperf.com/function-call-on-js-objects) he used there today, there is no difference.
But all of that doesn't matter for us, since I retained the input buffer for the next run, the code got much simpler and there is no need to pass objects around anymore.

I also added a bench test. Running it in the current code takes 144954 ms! and after the fix it only runs for 1107 ms which is an insane improvement.

Can you please CR again and run the bench tests locally to verify the fix?

I appreciate your time looking into it and hope things are better back home!

brianc:
Fantastic! Thanks for your work on this! ❤️ I'll merge this tomorrow in the AM & do a patch release.

regevbr:
Thanks @brianc for merging!
Please let me know when there is an official release with the fix :-)

brianc:
@regevbr Thanks for your persistence and great work on this.  I apologize for the delays on my end.  I like to release in the AM (CST) So I can be around all day in case some weird incompatibility goes out.  I've released your changes with `pg@8.2.2`.  Going to put them into my prod app today!

regevbr:
@brianc sure not a problem! always happy to help.
Your delays are more than justifiable and in any case it is not my place to judge as this is your (awesome) library :-) 
I will update my repos as well and let you know if everything works.
Thanks again!

PR #2271, opened 2020-07-10
Fix: use types configured on pg client in pg-query-stream

aravindanve:


brianc:
Yo! this is rad! Sorry for the delay in getting to this! Thank you tons! ❤️

PR #2273, opened 2020-07-10
Pass host to tls.connect for certificate validation

### LINKED ISSUE #2263: IP hosts are not validated correctly against certificate altnames
For example, connecting to IP 1.2.3.4 yields the following error:

```
Error [ERR_TLS_CERT_ALTNAME_INVALID]: Hostname/IP does not match certificate's altnames: Host: localhost. is not in the cert's altnames: IP Address:1.2.3.4
       at Object.checkServerIdentity (tls.js:250:17)
       at TLSSocket.onConnectSecure (_tls_wrap.js:1098:27)
       at TLSSocket.emit (events.js:198:13)
       at TLSSocket._finishInit (_tls_wrap.js:666:8)
```

Since TLS support was added to pg, it has [passed a socket](https://github.com/heap/node-postgres/blob/a1d00919b0b45845824767e05ae032db17167210/lib/connection.js#L46) to [`tls.connect`](http://nodejs.org/api/tls.html#tls_tls_connect_options_callback), meaning the host should be passed separately (it isn't). It passed `servername`, which is not valid for IP addresses and was removed in #1890.

The reason that the error message above uses localhost can be found in [`_tls_wrap.js`](https://github.com/nodejs/node/blob/8b4d0967ffa1d53dbceae1736d67b418ae479102/lib/_tls_wrap.js#L1503-L1508).

I found a previous issue (#2178) about this but it wasn't fixed. The correct fix is to always pass `host`. The documentation for `options.socket` says:

> If this option is specified, path, host and port are ignored, except for certificate validation.

I can submit a PR but I will need help if you'd like a test for this.

RazerM:
Fixes #2263

### DIFF
--- a/packages/pg/lib/connection.js
+++ b/packages/pg/lib/connection.js
@@ -77,6 +77,7 @@ class Connection extends EventEmitter {
       var tls = require('tls')
       const options = {
         socket: self.stream,
+        host,
       }
 
       if (self.ssl !== true) {


charmander:
Someone’ll need to write a test for this. I wonder what it does when `host` is a Unix domain socket?

RazerM:
> I wonder what it does when `host` is a Unix domain socket?

I think from the links I put in #2263 it's clear that host is only used for checking the server identity, and that for non-IP addresses this is currently set by servername. So nothing has changed for unix domain sockets in that regard.

RazerM:
What can I do to get this merged?

vitaly-t:
When indeed? :) I got [another issue open about the same](https://github.com/vitaly-t/pg-promise/issues/790), it seems.

charmander:
@vitaly-t That doesn’t sound like the same issue; the host in that one isn’t an IP address.

@RazerM It needs a test.

vitaly-t:
>  That doesn’t sound like the same issue; the host in that one isn’t an IP address, and the claimed behaviour is missing SNI

Yeah, sorry, I didn't look up close. But what's valid there is something else in the configuration here, since `pg-promise` just passes the config object on to this driver, without any modification.

charmander:
@vitaly-t It looks like pg sets `servername` unconditionally, but we want to allow it to be overridden; working on a fix for that. (That should mean the SNI is the same as the host instead of missing, but that’s probably what the issue reporter means.)

RazerM:
@charmander #1890 was merged without a test despite the consequences for IP addresses so I get the impression the test setup wouldn't easily support such a test. The existing tests all pass for this PR too.

I think it should be clear from the description in the issue what is wrong. It's quite frustrating because it's fairly easy to see why IP addresses don't work and how pg got like this:

1. pg passes a socket to `tls.connect`.
2. `tls.connect` has no idea which hostname it needs to check the identity for when it has a socket.
3. `tls.connect` uses [`options.servername || options.host || 'localhost'`](https://github.com/nodejs/node/blob/8b4d0967ffa1d53dbceae1736d67b418ae479102/lib/_tls_wrap.js#L1503-L1508) to verify server identity.
4. pg used to always pass `servername`, which as noted should not be done for IP addresses.
5. PR #1890 changed this to only do it for non-IP addresses, therefore the `tls.connect` identity verification uses 'localhost'. **This is bad!**
6. The snippet above shows why removing servername for IP addresses results in 'localhost' being used. You need to pass `host`!

charmander:
Of course it’s clear what’s wrong. That doesn’t mean it doesn’t need a test. It broke in the first place because there was no test.

penous:
Hey all, any updates on this?

jhnns:
Just stumbled upon this issue. I can confirm that this PR would fix the problem. Node's docs are quite clear in that regard that ["`path`, `host`, and `port` are ignored, except for certificate validation"](https://nodejs.org/docs/latest/api/tls.html#tlsconnectoptions-callback).

@charmander Would you still merge it if it had tests?

charmander:
@jhnns Yep!

jhnns:
@charmander I created a PR with the implementation and the tests: https://github.com/brianc/node-postgres/pull/3756

PR #2284, opened 2020-07-17
type(pg-connection-string): fix typings in ConnectionOptions

ssut:
- The `ssl` option can be either boolean or its own type consisting of cert, key, and ca.
- Because as [the documentation here](https://www.npmjs.com/package/pg-connection-string) says, any other query params should be preserved intact as well as types so I added `[key: string]: any` in `ConnectionOptions`.

Because without this I had to cast the result parsed as any type, or add `{ schema?: string }` in everywhere with such conditions that require `schema`, I think this PR will help those in need.

### DIFF
--- a/packages/pg-connection-string/index.d.ts
+++ b/packages/pg-connection-string/index.d.ts
@@ -7,9 +7,19 @@ export interface ConnectionOptions {
   port?: string | null
   database: string | null | undefined
   client_encoding?: string
-  ssl?: boolean | string
+  ssl?:
+    | string
+    | boolean
+    | {
+        cert: string
+        key: string
+        ca: string
+      }
 
   application_name?: string
   fallback_application_name?: string
   options?: string
+  schema?: string
+
+  [key: string]: any
 }


Macavirus:
This is correct, the `parse` function provided by `pg-connection-string` returns a value that is a valid `PoolConfig`, whereas the `ConnectionOptions` type currenly provided by the library incorrectly states that `ssl` key is ONLY a `boolean | string`.

Macavirus:
@brianc please take a look this currently requires ugly hacks like:
`parse(connectionStringFoo) as unknown as PoolConfig`

packages/pg-connection-string/index.d.ts
```diff
@@ -7,9 +7,19 @@ export interface ConnectionOptions {
   port?: string | null
   database: string | null | undefined
   client_encoding?: string
-  ssl?: boolean | string
+  ssl?:
+    | string
+    | boolean
+    | {
+        cert: string
```
CG-ANY, line 14: This object is still missing `rejectUnauthorized: boolean` property as documented [here](https://github.com/brianc/node-postgres/tree/master/packages/pg-connection-string#usage). Connecting to a host with a self-signed ssl certificate and PGSSLMODE=required will have ssl set to true and will fail if `rejectUnauthorized` is not set to `false`. We ran into issues on Heroku with our application running with self signed certificates and had to pass `rejectUnauthorized: false` to resolve those issues.

PR #2286, opened 2020-07-17
Speed up `bind` functionality

brianc:
Move from 3 loops (prepareValue, check for buffers, write param types, write param values) to a single loop.  This speeds up the insert benchmark by around 100 queries per second.  Performance improvement depends on number of parameters being bound.


packages/pg-protocol/src/serializer.ts
```diff
@@ -106,50 +106,53 @@ type BindOpts = {
   binary?: boolean
   statement?: string
   values?: any[]
+  // optional map from JS value to postgres value per parameter
+  valueMap?: (param: any) => any
```
sehrope, line 110: Rather than "valueMap", how about calling this "mapper" or "transformer"? The "Map" suffix sounds like it's an actual map but it's a function.

Also, this should accept more arguments, including the parameter index being transformed.

packages/pg-protocol/src/serializer.ts
```diff
@@ -106,50 +106,53 @@ type BindOpts = {
...
+      const mappedVal = valueMap ? valueMap(val) : val
+      paramWriter.addInt32(mappedVal.length)
+      paramWriter.add(mappedVal)
     } else {
-      buffer.addInt32(Buffer.byteLength(val))
-      buffer.addString(val)
+      writer.addInt16(0)
+      const mappedVal = valueMap ? valueMap(val) : val
+      paramWriter.addInt32(Buffer.byteLength(mappedVal))
+      paramWriter.addString(mappedVal)
```
sehrope, line 130: My read of this is that if a Buffer is supplied as an arg, it will be transformed and the result will be expected to still be a Buffer. Ditto for non-Buffers (expected to be string-like).

Instead, how about adding `const mappedVal = valueMap ? valueMap(val, i) : val` prior to the if-else-if-else block? That would allow transforming null or undefined into something else on a per-field basis. It also would allow the transform to return null, a string, or Buffer on a per-field basis and transform from a Buffer type bound as binary to a string one bound as text.

brianc: Great ideas. Will do, and write a couple more tests covering it too.

packages/pg-protocol/src/buffer-writer.ts
```diff
@@ -5,7 +5,7 @@ export class Writer {
   private offset: number = 5
   private headerPosition: number = 0
   constructor(private size = 256) {
-    this.buffer = Buffer.alloc(size)
+    this.buffer = Buffer.allocUnsafe(size)
```
kblcuk, line 8: Curious -- is changing to `allocUnsafe` here also contributes to performance improvements? The original PR comment doesn't mention anything about this change.
Also how certain we are that this doesn't cause any nasty side-effects of being, well, unsafe?

jantaylor: Here's a link to the [NodeJS documentation](https://nodejs.org/docs/latest-v12.x/api/buffer.html#buffer_static_method_buffer_allocunsafe_size) on `Buffer.allocUnsafe()`.

I'm not an expert on Buffers, though the documentation's last sentenance says this can increase a performance (subtle) over `Buffer.alloc()`.

> Allocates a new Buffer of size bytes. If size is larger than buffer.constants.MAX_LENGTH or smaller than 0, ERR_INVALID_OPT_VALUE is thrown.
>
> The underlying memory for Buffer instances created in this way is not initialized. The contents of the newly created Buffer are unknown and may contain sensitive data. Use Buffer.alloc() instead to initialize Buffer instances with zeroes.
>
> [...]
>
> Use of this pre-allocated internal memory pool is a key difference between calling `Buffer.alloc(size, fill)` vs. `Buffer.allocUnsafe(size).fill(fill)`. Specifically, `Buffer.alloc(size, fill)` will never use the internal Buffer pool, while `Buffer.allocUnsafe(size).fill(fill)` will use the internal Buffer pool if size is less than or equal to half `Buffer.poolSize`. The difference is subtle but can be important when an application requires the additional performance that Buffer.allocUnsafe() provides.

PR #2294, opened 2020-07-27
Fix most SSL negotiation packet tests being ignored

charmander:
`tc` was only one variable and the tests are asynchronous, so every test was writing 'E'.


brianc:
Thanks my friend ❤️

PR #2299, opened 2020-07-31
Fix 'Error: invalid sslmode value: "true"' in native mode

sundbry:
When using PGSSLMODE env in native mode, an invalid database URL string
is constructed with sslmode=true/false.

### DIFF
--- a/packages/pg/lib/connection-parameters.js
+++ b/packages/pg/lib/connection-parameters.js
@@ -123,7 +123,11 @@ class ConnectionParameters {
     add(params, this, 'connect_timeout')
     add(params, this, 'options')
 
-    var ssl = typeof this.ssl === 'object' ? this.ssl : this.ssl ? { sslmode: this.ssl } : {}
+    var ssl = (typeof this.ssl === 'object')
+      ? this.ssl
+      : ((this.ssl && typeof this.ssl === 'string')
+        ? { sslmode: this.ssl }
+        : {})
     add(params, ssl, 'sslmode')
     add(params, ssl, 'sslca')
     add(params, ssl, 'sslkey')


brianc:
Hey thanks for this!  Is there a way to write a test for this? If not, if you could show an example in code that causes the problem I can write a test to exercise the fix.  Either way I'm good - but need to get this under test so we don't regress on it! 😄

Andrei-Stepanov:
Hello, please merge and release this.
This is important.
It is necessary to have this merged.

```
[Error: invalid sslmode value: "true"
]
```

@sundbry thank you for the fix!

Andrei-Stepanov:
This bug is present in     "pg": "^8.7.1",

charmander:
@Andrei-Stepanov There’s still no test. What configuration causes this error?

Andrei-Stepanov:
@charmander : configuration/testcase scenario:

1. Have a postgresql-server **without** SSL transport.
2. Try to connect to this server using `pg` module enforcing/requiring SSL transport with env variable:
`PGSSLMODE=require`.

Connection should not be established, but it is estalished.  `pg` overrides this parameter. 

Related docs: 
https://www.postgresql.org/docs/current/libpq-envars.html
https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLMODE

charmander:
@Andrei-Stepanov Which is it? Connection is established when it should not be, or *Error: invalid sslmode value: "true"*?

Andrei-Stepanov:
I think this: `Connection is established when it should not be`.

charmander:
@Andrei-Stepanov That sounds like an issue separate from what this PR claims to address (and what I was requesting a test case for), then. Please open an issue for it.

Issue #2307, opened 2020-08-12
Uncaught exception while creating a SSL connection


chris--young:
I found an error in `Client.connect()` that isn't passed to the callback and can't be caught. If one of the `ca`, `key`, or `cert` strings you use while creating an SSL connection is not a valid certificate your program will crash.

This error should be properly handled and recoverable. An easy fix is to add a try catch around the `tls.connect()` call on line 92 of `lib/connection.js` and return while emitting the caught error. Then the error will gracefully be passed to the connection callback and the user can recover without their program crashing.

I'm happy to open a pr with the fix if you want.

This is the error...
```
_tls_common.js:129
      c.context.setCert(cert);
                ^

Error: error:0909006C:PEM routines:get_name:no start line
    at Object.createSecureContext (_tls_common.js:129:17)
    at Object.connect (_tls_wrap.js:1567:48)
    at Socket.<anonymous> (/Users/chris/Desktop/pg_error/node_modules/pg/lib/connection.js:92:23)
    at Object.onceWrapper (events.js:422:26)
    at Socket.emit (events.js:315:20)
    at addChunk (_stream_readable.js:302:12)
    at readableAddChunk (_stream_readable.js:278:9)
    at Socket.Readable.push (_stream_readable.js:217:10)
    at TCP.onStreamRead (internal/stream_base_commons.js:186:23) {
  library: 'PEM routines',
  function: 'get_name',
  reason: 'no start line',
  code: 'ERR_OSSL_PEM_NO_START_LINE'
}
```

This code will reproduce it...
```
const pg = require('pg');

const config = {
  host: 'redacted',
  database: 'redacted',
  user: 'redacted',
  password: 'redacted',
  ssl: {
    ca: 'foo',
    key: 'foo',
    cert: 'foo'
  }
};

const client = new pg.Client(config);

function callback(err) {
  if (err)
    return console.error('CALLBACK ERROR:', err); // this line should log the error but doesn't

  client.end();
}

try {
  client.connect(callback);
} catch (err) {
  console.error('CAUGHT ERROR:', err); // this line wont catch the error either
}
```

Lewiscowles1986:
Surely the server in the case of SSL should be configured for only SSL for data-security? Why would you recover rather than fix?

Lewiscowles1986:
Also, is this note similar to https://github.com/brianc/node-postgres/issues/2004

chris--young:
Yea this is the same issue. I didn't see that other issue when I searched before.

[...]

Issue #2004, opened 2019-11-08
Incorrect SSL ca, key or cert data causes uncaughtException


ianfairbairn:
Description:

I can successfully connect to a postgreSQL database using SSL if the certs (as strings) are correctly entered.  However, if I alter the ca, key or cert strings (to replicate a user fat-fingering them as they cut and paste them in the UI) I get an uncaughtException and my service falls over because the error doesn't bubble up to me to handle it.

Env:
Docker container
nodejs v11.6.0
pg v^7.12.0

Code:

const postgreSQL = require('pg');

exports.execSQL = (databaseConnectionModel, sql) => new Promise((resolve, reject) => {
	
	try {

		let pool;
		let config = {};

		config = {
			host: databaseConnectionModel.host,
			port: databaseConnectionModel.port,
			user: databaseConnectionModel.user,
			password: databaseConnectionModel.password,
			database: databaseConnectionModel.database
		}

		if (databaseConnectionModel.useSSL) {
			config.ssl = {
				ca: databaseConnectionModel.ca,
				key: databaseConnectionModel.key,
				cert: databaseConnectionModel.cert
			}
		}

		pool = new postgreSQL.Pool(config);

		pool.query(sql, (error, result) => {

			if (error) {           
				pool.end();
				reject(error); 
			} else {    
				pool.end();
				resolve(result);
			}
		});

	} catch (error) {
		reject(error);
	}
});

error:

error: uncaughtException: error:0906D064:PEM routines:PEM_read_bio:bad base64 decode
Error: error:0906D064:PEM routines:PEM_read_bio:bad base64 decode
    at Object.createSecureContext (_tls_common.js:134:17)
    at Object.connect (_tls_wrap.js:1221:48)
    at Socket.<anonymous> (/usr/src/app/node_modules/pg/lib/connection.js:95:23)
    at Object.onceWrapper (events.js:276:13)
    at Socket.emit (events.js:188:13)
    at addChunk (_stream_readable.js:288:12)
    at readableAddChunk (_stream_readable.js:269:11)
    at Socket.Readable.push (_stream_readable.js:224:10)
[...]

PR #2309, opened 2020-08-12
Prevents bad ssl credentials from causing a crash

chris--young:
This fixes a bug in `Client.connect()` where invalid ssl credentials would cause an error that was not passed to the callback and could not be caught.

Fixes: https://github.com/brianc/node-postgres/issues/2307
Fixes: https://github.com/brianc/node-postgres/issues/2004

### DIFF
--- a/packages/pg/lib/connection.js
+++ b/packages/pg/lib/connection.js
@@ -85,7 +85,11 @@ class Connection extends EventEmitter {
       if (net.isIP(host) === 0) {
         options.servername = host
       }
-      self.stream = tls.connect(options)
+      try {
+        self.stream = tls.connect(options)
+      } catch (err) {
+        return self.emit('error', err)
+      }
       self.attachListeners(self.stream)
       self.stream.on('error', reportStreamError)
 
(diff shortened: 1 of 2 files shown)


brianc:
Thanks!  This is fab - sorry for the delay...releasing a patch version this morning.

charmander:
Does the stream get cleaned up when this happens?

brianc:
> Does the stream get cleaned up when this happens?

according to the test you still need to call `client.end` (since the error was emitted on `.connect` and not when you construct the client) - it could be argued the client should just auto-end itself if the tls params cause an exception.

chris--young:
I can add a `stream.end()` in the catch block if thats preferable. It does seem to make sense that it should work that way based on how some of the other errors are being handled in `client.connect()`

brianc:
I don't have strong feelings about it either way - this is already better & calling `client.end()` on an ended client is a noop so we can change it to auto-end in the future if it starts causing problems.

chris--young:
Ok, up to you guys. I'm happy to make more improvements if necessary.

PR #2315, opened 2020-08-21
Fix context (this) in _checkPgPass.

Pimm:
In the specific scenario in which I was testing (`this.password` is `null`, and perhaps there are other factors at play) the following error is thrown:
> Cannot read property 'connectionParameters' of undefined

### DIFF
--- a/packages/pg/lib/client.js
+++ b/packages/pg/lib/client.js
@@ -215,7 +215,7 @@ class Client extends EventEmitter {
     } else if (this.password !== null) {
       cb()
     } else {
-      pgPass(this.connectionParameters, function (pass) {
+      pgPass(this.connectionParameters, (pass) => {
         if (undefined !== pass) {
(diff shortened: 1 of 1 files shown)


Pimm:
It turns out you can test it like so:
```javascript
const createClient = require('./test-helper').createClient

var client = createClient()
client.password = null
client.connection.emit('authenticationCleartextPassword')
```
The catch is that `~/.pgpass` has to contain a password which `pgpass` will read and return. So maybe we should mock `pgpass`.

Do you have any clever ideas after reading this?

vitaly-t:
Could we get this fix out, please?

brianc:
sure will work on this right now sorry for the delay

brianc:
I added a test which failed before your patch was applied [here](https://github.com/brianc/node-postgres/pull/2321)

brianc:
published fix in pg@8.3.3!

PR #2323, opened 2020-08-30
Support idle_in_transaction_session_timeout and statement for native driver

shayonj:
Right now when using `pg` with `pg-native` the config options
`idle_in_transaction_session_timeout` and `statement_timeout`
do not get applied. This is due to, when building the connection
string for native driver these options aren't pasted

I took advantage of `options` conn parameter and passed both the config options
into it using `-c` which postgres supports. This builds on top of the work happened in https://github.com/brianc/node-postgres/pull/2216.

Also fixes https://github.com/brianc/node-postgres/issues/2103

Also added some tests. Very much open to feedback and comments :).


vhmth:
An expedient (and better review than what I could offer myself) would be super appreciated here. 🙏  Supporting good timeouts is critical for us to run the native driver in production at Loom. ❤️

packages/pg/test/unit/connection-parameters/creation-tests.js
```diff
@@ -172,6 +172,26 @@ test('libpq connection string building', function () {
     )
   })
 
+  test('builds conn string with options', function () {
```
brianc, line 175: What happens if both `options` and `statement_timeout` is included in the config?

brianc:
Yo - sorry I was late to the party here.  Left a question on here....could we include a test when both `options` and `statement_timeout` is included?  Once we got that I can merge & release a new minor ver.

packages/pg/test/unit/connection-parameters/creation-tests.js
shayonj: ah! thats good catch, it would overridden the second the second `options=` attr. Updated the logic and added some more test.

shayonj:
late to the part myself this time, thanks @brianc for the review, updated with some more spec 🙌

machuga:
Hi, @brianc!

Is this PR still under consideration? Having this fully supported would be great. We are moving more of our libraries at work toward using `pg-native` (and I submitted a PR to add `pg-native` to `knex) due to running into the same performance issues here https://github.com/brianc/node-postgres/issues/1993.

Thanks for all your fantastic work on these libraries!

PR #2343, opened 2020-09-10
Fix metadata for pg-connection-string

benjie:
When you visit the [pg-connection-string module on npmjs.com](https://www.npmjs.com/package/pg-connection-string) it links to the wrong place.


brianc:
thanks!

PR #2352, opened 2020-09-25
Replace semver with optional peer dependencies

TrySound:
See example https://github.com/sindresorhus/gulp-chown/blob/bb74168c957b3a94f122aafcecf7ebc87088ec46/package.json#L42-L49

This feature is supported by both npm and yarn.

### DIFF
--- a/packages/pg/lib/native/client.js
+++ b/packages/pg/lib/native/client.js
@@ -3,15 +3,10 @@
 // eslint-disable-next-line
 var Native = require('pg-native')
 var TypeOverrides = require('../type-overrides')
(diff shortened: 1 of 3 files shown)


TrySound:
cc @brianc

charmander (approved):
Seems super under-documented in npm, but usable now: https://github.com/npm/rfcs/issues/221#issuecomment-695127530

Thanks!

TrySound:
Yes, this feature was first added in yarn and not that long ago in other package managers.

brianc:
Wow this is awesome, thank you.

Issue #2340, opened 2020-09-10
DatabaseError class should be public


garaboncias:
We need to handle Database errors in higher level, and we wan't to check if error come from db, and in that case error code is a specific code. Please export DatabaseError class from `node-postgres` to able to achieve that.

```typescript
// Current workaround but its hacky.
import { DatabaseError } from 'pg-protocol/dist/messages';

/** https://www.postgresql.org/docs/current/errcodes-appendix.html */
enum postgresErrorCodes {
    read_only_sql_transaction = '25006'
}

   if (error instanceof DatabaseError && error.code == postgresErrorCodes.read_only_sql_transaction) {
      // handle write modification tries
   }
```

tcarrio:
@garaboncias just a heads up, the fix for this was just merged. Keep an eye out for the next release 😁

brianc:
Release went out! pg@8.4.0

tcarrio:
Crazy fast @brianc . Thanks for updating us here too ❤️

garaboncias:
Thank you for your support and kindness!

PR #2359, opened 2020-10-04
Public export of DatabaseError

tcarrio:
## Description

When handling errors from this package it would be better to have the errors publicly exported

## Change Summary

- Updated root exports of 'pg-protocol' to include DatabaseError

Ref: #2340


brianc:
👍 thanks for the PR!

Issue #2363, opened 2020-10-07
Type parsers conflict


vitaly-t:
When your application uses multiple third-party modules, each initialized from reusable `Pool`, is there any provision as to how to deal with each module trying to set its own type parsers?

From what I see, we are getting nothing but conflicts in this scenario, as parsers are set on a singleton. Is there any solution to this that I'm not aware of?

charmander:
Pass `types` into the constructor. [Option documented](https://node-postgres.com/api/client/#constructor), [value needs documentation](https://node-postgres.com/api/types/).

charmander:
Duplicate of #1838

vitaly-t:
Type `TypeOverrides`  appears to be highly reusable. How come it is not exposed by the library's root? It really should be.

Issue #1838, opened 2019-02-18
How to configure types only for a particular instance of node-postgres?


gajus:
It appears that the only way to configure type parser is by doing so globally, i.e.

```js
var types = require('pg').types
types.setTypeParser(20, function(val) {
  return parseInt(val)
});

```

– https://github.com/brianc/node-pg-types

I would like to restrict type parsing to a particular connection.

Is there an API at the moment to do so?

gajus:
This could be as simple as allowing to override `types`.

https://github.com/brianc/node-postgres/blob/fcd0f02210f532bf2d823ea05105f4aaade389eb/lib/index.js#L34

charmander:
`pg.types` is a convenience reexport; pg itself doesn’t use that property. You can override type parsing by configuration (i.e. per client or pool) by passing something that satisfies the necessary interface in the `types` property, the most convenient something [I’m aware of][1] being a `TypeOverrides` from [pg/lib/type-overrides.js][2]:

```js
const TypeOverrides = require('pg/lib/type-overrides');

const OID_INT8 = 20;

const types = new TypeOverrides();
types.setTypeParser(OID_INT8, BigInt);

const pool = new pg.Pool({
    ⋮
    types,
});

const client = new pg.Client({
    ⋮
    types,
});
```


[1]: https://github.com/charmander/pg-auto/blob/eb91aed645e19073fc5428bd67d5c57393b543c4/pool.js#L5
[2]: https://github.com/brianc/node-postgres/blob/fcd0f02210f532bf2d823ea05105f4aaade389eb/lib/type-overrides.js

gajus:
Is there a way to override `TypeOverrides` after pool has been instantiated?

```js
const client = new pg.Client({});

client._types = types;

```

The latter does not seem to do the trick.

Will debug this further on my own (seems like I could definitely override the query getTypeParser, https://github.com/brianc/node-postgres/blob/fcd0f02210f532bf2d823ea05105f4aaade389eb/lib/client.js#L459), but if there is a better way, please let me know.

charmander:
> Is there a way to override `TypeOverrides` after pool has been instantiated? … `new pg.Client`

On a pool or a client? What you wrote should work fine for a *client* for queries after the assignment. `Client.prototype.setTypeParser`/`Client.prototype.getTypeParser` exist too. On a pool, you can assign arbitrary client options after instantiation, so set `pool.options.types`.

gajus:
Thank you

[...]

PR #2369, opened 2020-10-08
Exposing TypeOverrides from the library's root

vitaly-t:
`TypeOverrides` is too important not to expose it. It should be available from the root.

It is essential part of custom type parsing, for which singleton `types` is exposed, while `TypeOverrides` is not.

One of the big issues the current implementation has - impossible to properly declare the type in TypeScript, since it is a class, which we have to use as `any` interface instead (it's awkward).

Related: #1838, #2363

### DIFF
--- a/packages/pg/lib/index.js
+++ b/packages/pg/lib/index.js
@@ -3,6 +3,7 @@
 var Client = require('./client')
 var defaults = require('./defaults')
 var Connection = require('./connection')
+var TypeOverrides = require('./type-overrides')
 var Pool = require('pg-pool')
 
 const poolFactory = (Client) => {
@@ -20,6 +21,7 @@ var PG = function (clientConstructor) {
   this.Pool = poolFactory(this.Client)
   this._pools = []
   this.Connection = Connection
+  this.TypeOverrides = TypeOverrides
   this.types = require('pg-types')
 }
 


brianc:
Thanks for this - this seems reasonable, but could you write some tests which will both prevent this from regressing & also demonstrate how this would be useful?

vitaly-t:
It's just type exposure through variable, why do you need tests for that?

And as for how this is help ful, I explained it from start:

> One of the big issues the current implementation has - impossible to properly declare the type in TypeScript, since it is a class, which we have to use as any interface instead (it's awkward).

i.e. the lack of the type makes it very awkward to use it from TypeScript, with direct file inclusion and all. Especially because that extension is a class, and needs proper TypeScript class declaration.

brianc:
> It's just type exposure through variable, why do you need tests for that?

Otherwise in the future it might get unexposed & then we wouldn't know we're breaking anything until we release a version w/o it exposed.  So just a simple check to make sure it's there is fine.

gabegorelick:
It would be great to get this merged. `@types/pg` has had to add this path to their type definitions: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/51175

> but could you write some tests which will both prevent this from regressing & also demonstrate how this would be useful

Are there existing tests for exports that this can be added to?

mesqueeb:
@vitaly-t can you update this PR so the new esm index files also expose `TypeOverrides`.

@brianc this one could also get some love now that the esm index files are in! 🎉

brianc:
@mesqueeb - see [new PR](https://github.com/brianc/node-postgres/pull/3433) 😄

brianc:
closed in https://github.com/brianc/node-postgres/pull/3433

PR #2383, opened 2020-10-18
refactor(): Test refactor

chyzwar:
This is on top of my another branch. (pg-query-stream-typescript)

Convert tests to ts
user assert insstead of expect.js
use async/await instead of co
remove lodash
When I did initrial investigation I saw that project use nodejs assert in most cases. Based on that I converted usage of expect.js to assert. Later I discovered that chai.assert is also used. What do you plan o use longe term ?


chyzwar:
I will wait for pg-stream-typescript to be merged first. this is building on top of that.

chyzwar:
This is quite big, most 99% changes are only to tests. It is now rebase on master and you can start looking into this. Only test in PG packages are left but once this is merged I can start looking int o them.

charmander:
One change per PR might be better?

chyzwar:
@brianc @charmander  This PR refactor tests, convert them to TS, remove legacy test framework and stadarize on assert + mocha. There no changes to actual packages code beside one line to packages/pg-connection-string/index.d.ts

I can split this pr per package or you guys want to take it head on?

charmander:
I’m not sure what brianc would prefer, but I’d be able to review these faster as separate PRs. Not per package, but per logical change. (The list in this PR’s description – use assert instead of expect.js, use async/await instead of co, remove lodash – is a pretty good example.)

chyzwar:
It growed out of hand. I can possibly slice these by package but now it would be difficult to replay these by change. 
I still think there is a lot of value in this PR.

chyzwar:
Build is now green. I would suggest if no changes were made to source code of packages and verify tests quickly. Some test would lose history but that is more because git cannot track renames very well.

PR #2394, opened 2020-10-28
Forward options’ ssl.key even when non-enumerable

### LINKED ISSUE #2392: Cert-based authentication broken
Starting yesterday with the upgrade to 8.4.2, connections now fail with the error message: `error: connection requires a valid client certificate`.

It appears that the switch to making the [ssl key not enumerable](https://github.com/brianc/node-postgres/commit/fd2c3563a57f19ca49cefa6b1de999d9aaa9b5f5) means that [the key is no longer copied](https://github.com/brianc/node-postgres/blob/master/packages/pg/lib/connection.js#L79) because Object.assign() only works with enumerable properties.

This became much trickier to debug since the pg dependency on pg-pool allows for minor version upgrades. We downgraded to 7.17.0 to fix things for the short term.

Thanks

charmander:
Fixes #2392.


brianc:
Hey @charmander - thank you so much for diving in to this & trying to get it working.  I know working with travis can be pretty incredibly painful...yesterday it took >2 hours for some tests to run.  Would me upgrading to a paid plan help w/ the turn-around time there?  Also, wondering if it'd be worthwhile for me to take the functional change to the code & write a unit test (and mock the ssl socket call) to check it's working and we can work on the larger travis integration tests in another PR?  That way we can close the actual bug fast?

charmander:
> Would me upgrading to a paid plan help w/ the turn-around time there?

Hmm… well, how would you feel about a PR that switched the CI to GitHub Actions?

> Also, wondering if it'd be worthwhile for me to take the functional change to the code & write a unit test (and mock the ssl socket call) to check it's working and we can work on the larger travis integration tests in another PR? That way we can close the actual bug fast?

I’ve decided to write just one integration test that runs in its own CI job. When the tests are refactored to read configuration consistently, it won’t be needed anymore :)

Sorry it took so long to get back to this.

brianc:
> Hmm… well, how would you feel about a PR that switched the CI to GitHub Actions?

I would feel great about that.

> I’ve decided to write just one integration test that runs in its own CI job. When the tests are refactored to read configuration consistently, it won’t be needed anymore :)

dope



> Sorry it took so long to get back to this.

100% no worries at all.  Take all the time you want! Sorry travis was blocking you.

brianc:
release pg@8.5.0

PR #2420, opened 2020-11-25
Fix double readyForQuery

brianc:
This is fixing a double `readyForQuery` message being sent from the backend (because we were calling `sync` after an error, which I already fixed in the main driver).  I also pulled in the reproduction case for another error from #2333, and I'm going to work on fixing that one as well.

### DIFF
--- a/packages/pg-cursor/index.js
+++ b/packages/pg-cursor/index.js
@@ -37,6 +37,7 @@ Cursor.prototype._rowDescription = function () {
 }
 
 Cursor.prototype.submit = function (connection) {
+  this.state = 'submitted'
   this.connection = connection
   this._portal = 'C_' + nextUniqueID++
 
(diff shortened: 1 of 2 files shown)


brianc:
Included a test case for #2333 to reproduce the issue and then turned the test green.  The issue was a queued stream which the client had not submitted was having an error triggered on it when `.end` was called on the client. The client errors out all waiting queries if you `.end` the client while there are queued queries.  The entire query queue concept is an unfortunate API decision I made like 9 years ago & it causes all sorts of confusing problems like this.  I'd like to remove the ability to queue queries at some point in the future.  It doesn't make a lot of sense to me as you're queuing up a bunch of stuff that may or may not work in the future depending on future states & makes the API a bunch more complicated.  A lot of this was before there were things like async/await (or even before the async library was widely used) as a "convenience" thing which...yeah...bad idea.  The problem w/ removing it is it's a subtle & large-ish breaking change and also a ton of my early unit and integration tests rely on the behavior. So, will be a big undertaking to remove it.  Anyways - this fixes another weird edge cases w/ queued query canceling.

PR #2345, opened 2020-09-10
Add support for ?sslmode connection string param

benjie:
Fixes #1949

If you're using connection strings, you can't use the `PGSSLMODE=no-verify` envvar because that's only used when `config.ssl` is not set. I use `?ssl=true&sslmode=no-verify&sslrootcert=/app/data/amazon-rds-ca-cert.pem` which does set `config.ssl`, so we need to also support parsing sslmode from the connection string. I've copied the logic (roughly) from:

https://github.com/brianc/node-postgres/blob/8291b233b81312ce2fbfce12ccd98ceceb53f5b9/packages/pg/lib/connection-parameters.js#L21-L32

I didn't see a CONTRIBUTING file, so I've taken a guess at the steps to make a good PR:

- [x] Added fix
- [x] Noted the GitHub issue it closes
- [x] Added test cases / 100% code coverage
- [x] Matched surrounding code style / passed linting
- [x] Sponsor @brianc :sparkling_heart:

PR #2421, opened 2020-11-27
Document sslmode connection string parameter

sonicdoe:
https://github.com/brianc/node-postgres/pull/2345 added support for the `sslmode` connection string parameter. This pull request documents that change in pg-connection-string’s readme.


brianc (approved):
groovy! thanks!

PR #2436, opened 2020-12-29
Add more SASL validation and fix tests

sehrope:
Bunch of commits to clean up some of the SASL internals, add validation for function args, validation for data types of server responses, and some regex validation as well (e.g. base64 server responses should be valid base64). They're broken up into separate commits to make it easier to review and the final file is a bit easier to digest too.

Also removes the global `asssert.throws(...)` test helper as it was overwriting the real one in the built-in asserts module. The removed one did not actually check the error signature so it would report success for unrelated errors as long as the code block threw something (which was kind of confusing while testing this out...).

The only thing using that extra validation of the error signature was the SASL tests so it shouldn't break anything and all the tests pass with that change. The SASL tests have also been updated to add in some missing parameters. Previously they weren't supplying all the args needed to actually perform the tests so they'd have early or unrelated failures.

### DIFF
--- a/packages/pg/lib/sasl.js
+++ b/packages/pg/lib/sasl.js
@@ -58,53 +58,52 @@ function finalizeSession(session, serverData) {
     throw new Error('SASL: Last message was not SASLResponse')
   }
 
-  var serverSignature
-
-  String(serverData)
-    .split(',')
-    .forEach(function (part) {
-      switch (part[0]) {
-        case 'v':
-          serverSignature = part.substr(2)
-          break
-      }
-    })
+  const attrPairs = parseAttributePairs(serverData)
+  const serverSignature = attrPairs.get('v')
+  if (!serverSignature) {
+    throw new TypeError('SASL: server signature is missing')
+  }
 
   if (serverSignature !== session.serverSignature) {
     throw new Error('SASL: SCRAM-SERVER-FINAL-MESSAGE: server signature does not match')
   }
 }
 
-function extractVariablesFromFirstServerMessage(data) {
-  var nonce, salt, iteration
+function parseAttributePairs(text) {
+  if (typeof text !== 'string') {
+    throw new TypeError('SASL: attribute pairs text must be a string')
+  }
 
-  String(data)
-    .split(',')
-    .forEach(function (part) {
-      switch (part[0]) {
-        case 'r':
-          nonce = part.substr(2)
-          break
-        case 's':
-          salt = part.substr(2)
-          break
-        case 'i':
-          iteration = parseInt(part.substr(2), 10)
-          break
+  return new Map(
+    text.split(',').map((attrValue) => {
+      if (!/^.=/.test(attrValue)) {
+        throw new Error('SASL: Invalid attribute pair entry')
       }
+      const name = attrValue[0]
+      const value = attrValue.substring(2)
+      return [name, value]
     })
+  )
+}
 
+function extractVariablesFromFirstServerMessage(data) {
+  const attrPairs = parseAttributePairs(data)
+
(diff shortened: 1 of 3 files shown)


packages/pg/test/test-helper.js
```diff
@@ -111,16 +111,6 @@ assert.success = function (callback) {
...
-assert.throws = function (offender) {
-  try {
-    offender()
-  } catch (e) {
-    assert.ok(e instanceof Error, 'Expected ' + offender + ' to throw instances of Error')
-    return
-  }
-  assert.ok(false, 'Expected ' + offender + ' to throw exception')
-}
-
```
brianc, line 123: Yeah, good call.  Monkey-patching assert was such a bad idea I had such a long time ago.  The use of globals throughout some of the tests is also gross....need to fix up one day soonish. 🤦

packages/pg/lib/sasl.js
```diff
@@ -20,19 +20,27 @@ function continueSession(session, password, serverData) {
...
+    throw new Error('SASL: SCRAM-SERVER-FIRST-MESSAGE: serverData must be a string')
+  }
 
-  const sv = extractVariablesFromFirstServerMessage(serverData)
+  const sv = parseServerFirstMessage(serverData)
 
   if (!sv.nonce.startsWith(session.clientNonce)) {
     throw new Error('SASL: SCRAM-SERVER-FIRST-MESSAGE: server nonce does not start with client nonce')
+  } else if (sv.nonce.length === session.clientNonce.length) {
+    throw new Error('SASL: SCRAM-SERVER-FIRST-MESSAGE: server nonce is too short')
```
brianc, line 35: nit: could these two if conditions just be combined into `if (sv.nonce !== session.clientNonce)`? As I think if a string starts with another string and has exactly the same length then they're the same string?

sehrope: No it's two separate checks. The RFC, https://tools.ietf.org/html/rfc5802, says:

> In response, the server sends a "server-first-message" containing the
> user's iteration count i and the user's salt, and appends its own
> nonce to the client-specified one.

So the server's response must be the clients nonce plus some random server nonce. Need to ensure our original client nonce is there (which the previous code was already doing). The additional check is to ensure the server added something non-empty of it's own. Separate checks gives a bit prettier error message if something is off.

brianc: ohhh that makes sense yes you're right I actually read that backwards when I was reviewing - if the length is equal it __is__ a problem, not if the length is equal it is not a problem. Got it.

brianc (approved):
Thank you so much for this!  Most of the original SASL code was contributed by not me.  This is v helpful.  As far as merging goes, I typically do a squash merge to keep a somewhat "one commit per PR" flow on master. (I have accidentally or for some times on purpose reasons used merge commits in the past though...I'm not super consistent.)  Do you have a preference? I appreciate all the separate commits on this PR to make reviewing easier & don't want to rob you of that work if you want them all to land on master individually. It can make pin-pointing a regression in the future easier and can make the job of people who want to see what changed between versions a bit easier though if its a single squash-merge so they can kinda go "github commit page -> link to each PR which landed the change".

sehrope:
My pleasure and thanks as always for this awesome lib.

Yes I like the separate commits mainly to simplify reviewing too. Especially for the incremental ones like function renames as it let you focus on the real changes in other commits. Plus it lets me know along the way what (*I*) broke.

No preference on how it gets merged though. FYI, the additional validations break the (*broken*) SASL tests so it's really only a full unit as a whole anyway.

charmander (approved):
:tada:

PR #2442, opened 2021-01-15
Fix typo

sonicdoe:
https://github.com/brianc/node-postgres/commit/6be3b9022f83efc721596cc41165afaa07bfceb0 added support for the `sslmode` parameter, not `ssl-mode`.


charmander (approved):
Thanks!

PR #2445, opened 2021-01-17
pg: Re-export DatabaseError from 'pg-protocol'

### LINKED ISSUE #2378: Export DatabaseError in pg package
I wanted to catch errors I got from pg in order to improve my project's reporting, but found out that I needed to import pg-protocol package to get DatabaseError type.

This seems a little bit illogical: I think that if I interact with a package, it should export not only the type definitions, but also the errors it's going to throw up to me.

Another reason to do this is that now I have both node-postgres and pg-protocol in my package.json, and, for example, if I update node-postgres to a new version and forget to update pg-protocol, the actual runtime type of the DatabaseError I catch may change under me, and my code that's depending on the DatabaseError type imported from outdated pg-protocol package would silently cease to work.

I also wanted to point out that documentation seems to be a bit lacking on that regard, and doesn't mention any errors that I should be catching as a user of pg package.

If you approve such changes in principle, I would be happy to contribute PR implementing them, both re-export and docs (will create a separate issue to track docs progress).

cakoose:
Before, users would have to import DatabaseError from 'pg-protocol'.  If
there are multiple versions of 'pg-protocol', you might end up using the
wrong one.

Closes #2378

### DIFF
--- a/packages/pg/lib/index.js
+++ b/packages/pg/lib/index.js
@@ -4,6 +4,7 @@ var Client = require('./client')
 var defaults = require('./defaults')
 var Connection = require('./connection')
 var Pool = require('pg-pool')
+const { DatabaseError } = require('pg-protocol')
 
 const poolFactory = (Client) => {
   return class BoundPool extends Pool {
@@ -21,6 +22,7 @@ var PG = function (clientConstructor) {
   this._pools = []
   this.Connection = Connection
   this.types = require('pg-types')
+  this.DatabaseError = DatabaseError
 }
 
 if (typeof process.env.NODE_PG_FORCE_NATIVE !== 'undefined') {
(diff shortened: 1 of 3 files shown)


Minishlink:
@cakoose I'm not sure how you consume it? Since pg is not a Typescript project, there's no types associated to `DatabaseError` even if pg-protocol is a Typescript project. Or am I missing something here?

cakoose:
Thanks for the reminder! Just opened a PR for "@types/pg": https://github.com/DefinitelyTyped/DefinitelyTyped/pull/52804

(I had to wait until this landed before opening the PR on DefinitelyTyped.  The last attempt opened the PR too early: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/49188)

PR #2459, opened 2021-01-29
Fix README to separate sponsors onto separate lines

sehrope:
I noticed that the sponsor links show up side to side so it looks like it's one company named "*CrateDBeaze*". This splits them out onto separate lines. Also gets rid of an inline png that wasn't actually be rendered.


brianc:
Ah thanks a lot man!  Appreciate it.

PR #2471, opened 2021-02-13
Adding pg to peerDependencies

ed0wolf:
Yarn2 requires strict imports, I.E. all project dependencies need to exist in that project's package.json.

```
Required package: pg (via "pg/lib/result.js")
Required by: pg-cursor@npm:2.5.2 (via ....../.yarn/cache/pg-cursor-npm-2.5.2-b99730336d-1735cfa9a6.zip/node_modules/pg-cursor/)
```


packages/pg-cursor/package.json
```diff
@@ -18,5 +18,8 @@
   "devDependencies": {
     "mocha": "^7.1.2",
     "pg": "^8.5.1"
+  },
+  "peerDependencies": {
+    "pg": ">=8.0"
```
charmander, line 23: I’m not sure if this should assume it’ll be compatible with a new major version of pg. If it turns out to be, a new release can always expand the range. So, `"pg": "^8"`?
```suggestion
    "pg": "^8"
```

charmander (changes requested):

charmander (approved):

brianc:
thanks! Sorry for the delay - had a bunch of stuff happen in my life all at once so I had to take a computer break for a few weeks.

PR #2489, opened 2021-03-11
[pg-protocol] patch out the const enum in messages.d.ts

forivall:
Alternative fix to #2473

This is the diff it generates to the typedefs:

```patch
--- dist/messages.const-enum.d.ts	2021-03-11 15:16:01.000000000 -0800
+++ dist/messages.d.ts	2021-03-11 15:16:01.000000000 -0800
@@ -1,33 +1,32 @@
 /// <reference types="node" />
 export declare type Mode = 'text' | 'binary';
-export declare const enum MessageName {
-    parseComplete = "parseComplete",
-    bindComplete = "bindComplete",
-    closeComplete = "closeComplete",
-    noData = "noData",
-    portalSuspended = "portalSuspended",
-    replicationStart = "replicationStart",
-    emptyQuery = "emptyQuery",
-    copyDone = "copyDone",
-    copyData = "copyData",
-    rowDescription = "rowDescription",
-    parameterStatus = "parameterStatus",
-    backendKeyData = "backendKeyData",
-    notification = "notification",
-    readyForQuery = "readyForQuery",
-    commandComplete = "commandComplete",
-    dataRow = "dataRow",
-    copyInResponse = "copyInResponse",
-    copyOutResponse = "copyOutResponse",
-    authenticationOk = "authenticationOk",
-    authenticationMD5Password = "authenticationMD5Password",
-    authenticationCleartextPassword = "authenticationCleartextPassword",
-    authenticationSASL = "authenticationSASL",
-    authenticationSASLContinue = "authenticationSASLContinue",
-    authenticationSASLFinal = "authenticationSASLFinal",
-    error = "error",
-    notice = "notice"
-}
+export type MessageName =
+    | "parseComplete"
+    | "bindComplete"
+    | "closeComplete"
+    | "noData"
+    | "portalSuspended"
+    | "replicationStart"
+    | "emptyQuery"
+    | "copyDone"
+    | "copyData"
+    | "rowDescription"
+    | "parameterStatus"
+    | "backendKeyData"
+    | "notification"
+    | "readyForQuery"
+    | "commandComplete"
+    | "dataRow"
+    | "copyInResponse"
+    | "copyOutResponse"
+    | "authenticationOk"
+    | "authenticationMD5Password"
+    | "authenticationCleartextPassword"
+    | "authenticationSASL"
+    | "authenticationSASLContinue"
+    | "authenticationSASLFinal"
+    | "error"
+    | "notice";
 export interface BackendMessage {
     name: MessageName;
     length: number;
@@ -83,7 +82,7 @@
 export declare class CopyDataMessage {
     readonly length: number;
     readonly chunk: Buffer;
-    readonly name = MessageName.copyData;
+    readonly name = "copyData";
     constructor(length: number, chunk: Buffer);
 }
 export declare class CopyResponse {
@@ -161,7 +160,7 @@
     readonly length: number;
     readonly message: string | undefined;
     constructor(length: number, message: string | undefined);
-    readonly name = MessageName.notice;
+    readonly name = "notice";
     severity: string | undefined;
     code: string | undefined;
     detail: string | undefined;
```

### DIFF
--- a/packages/pg-protocol/package.json
+++ b/packages/pg-protocol/package.json
@@ -17,7 +17,7 @@
   },
   "scripts": {
     "test": "mocha dist/**/*.test.js",
-    "build": "tsc",
+    "build": "tsc && node script/remove-public-const-enum.js",
     "build:watch": "tsc --watch",
     "prepublish": "yarn build",
     "pretest": "yarn build"
--- a/packages/pg-protocol/script/remove-public-const-enum.js
+++ b/packages/pg-protocol/script/remove-public-const-enum.js
@@ -4,31 +4,39 @@
 const fs = require('fs')
 const path = require('path')
 
-const filepath = path.join(__dirname, '../dist/messages.d.ts')
-const backuppath = path.join(__dirname, '../dist/messages.const-enum.d.ts')
+const PATCH_SOURCE = process.argv.slice(2).indexOf('--src') >= 0
+const filepath = path.join(__dirname, PATCH_SOURCE ? '../src/messages.ts' : '../dist/messages.d.ts')
+const backuppath = path.join(
+  __dirname,
+  PATCH_SOURCE ? '../src/messages.const-enum.ts' : '../dist/messages.const-enum.d.ts'
+)
+const otherfiles = ['../src/parser.ts']
 /** @type {string} */
 let srcpath
-// use the filepath if it's newer
-try {
-  const backupStat = fs.statSync(backuppath)
-  const fileStat = fs.statSync(filepath)
-  srcpath = fileStat.mtimeMs > backupStat.mtimeMs ? filepath : backuppath
-} catch (err) {
-  if (err.code !== 'ENOENT') {
-    throw err
-  }
+if (PATCH_SOURCE) {
   srcpath = filepath
+} else {
+  // use the filepath if it's newer
+  try {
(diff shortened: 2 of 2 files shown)


brianc:
Hmm...I'm confused here is this an option if merging #2490 is not an option? Because I'm leaning towards #2490 right now (it looks good to me!)

forivall:
Either one would be fine. If you merge that one, close this one and we're good here!

brianc:
hmmm thinking about this a bit....there's [this PR](https://github.com/brianc/node-postgres/pull/2473) that fixes this by just not making it a `const` enum.  Do you think the linked PR preserves backwards compatibility more?  Changing from an enum to a string constant is a breaking change right? And then i'd have to do semver major for that which seems like a lot when we could just remove the `const`?  Thanks for your help on this btw!

forivall:
A const enum doesn't actually produce any code (except with the preserveConstEnums option, which you don't use); the generated code is equivalent to using string literal. So changing it to string literals may break people's typescript builds, but it won't cause runtime issues.

Changing it to a non-const enum is fine, though, it changes the usage from literals to property access on the enum object. as you were mentioning hot path code, you might want to evaluate if that perf would be an issue.

forivall:
Closing as the other solutions seem preferable.

PR #2490, opened 2021-03-11
[pg-protocol] use literals instead of const enum

forivall:
Alternative to #2473 that removes the const enum for literals (using the --src option of the script in #2489)


brianc:
Thanks for doing this! ❤️

forivall:
re: https://github.com/brianc/node-postgres/pull/2473#issuecomment-802144920 @brianc ping for bumping a new version

PR #2498, opened 2021-03-18
Implement a `PoolError` with dedicated codes

rmedaer:
Previously the Pool was throwing built-in JS `Error`. It's therefore difficult to identify each kind of error.
This patch allow the developer to identify each type of error raised by the Pool with a `PoolError`.

Each `PoolError` has its own code and a message (original message for backward compatibility).

Codes and errors are documented in README file.

### DIFF
--- a/packages/pg-pool/README.md
+++ b/packages/pg-pool/README.md
@@ -331,6 +331,20 @@ var bluebirdPool = new Pool({
 
 __please note:__ in node `<=0.12.x` the pool will throw if you do not provide a promise constructor in one of the two ways mentioned above.  In node `>=4.0.0` the pool will use the native promise implementation by default; however, the two methods above still allow you to "bring your own."
 
+## error management
+
+Errors throwed by the pool are instances of `PoolError`. Each `PoolError` has a message (short description of the exception) and a structured code composed of 5 characters string starting with letter `Z`.
+The following table list the existing `PoolError`s and their code:
+
+| Code  | Message                                                                  |
+| ----- | ------------------------------------------------------------------------ |
+| Z0001 | Cannot use a pool after calling end on the pool                          |
+| Z0002 | Passing a function as the first parameter to pool.query is not supported |
+| Z0003 | Release called on client which has already been released to the pool     |
+| Z0004 | Called end on pool more than once                                        |
+| Z0005 | Timeout exceeded when trying to connect                                  |
+| Z0009 | Unexpected condition                                                     |
+
 ## maxUses and read-replica autoscaling (e.g. AWS Aurora)
 
 The maxUses config option can help an application instance rebalance load against a replica set that has been auto-scaled after the connection pool is already full of healthy connections.
--- a/packages/pg-pool/index.js
+++ b/packages/pg-pool/index.js
@@ -24,7 +24,7 @@ class PendingItem {
 }
 
 function throwOnDoubleRelease() {
-  throw new Error('Release called on client which has already been released to the pool.')
+  throw new PoolError('Release called on client which has already been released to the pool.', 'Z0003')
 }
 
 function promisify(Promise, callback) {
@@ -56,6 +56,15 @@ function makeIdleListener(pool, client) {
     // the client has already been closed & purged and is unusable
     pool.emit('error', err, client)
   }
+}
+
+class PoolError extends Error {
+    // A PoolError is an error throwed during a Pool process. Each type of error contains a specific "code".
+    // Please document in README file the error codes.
+    constructor(message, code) {
+        super(message)
+        this.code = code
+    }
 }
 
 class Pool extends EventEmitter {
@@ -143,7 +152,7 @@ class Pool extends EventEmitter {
     if (!this._isFull()) {
       return this.newClient(pendingItem)
     }
-    throw new Error('unexpected condition')
+    throw new PoolError('unexpected condition', 'Z0009')
   }
 
   _remove(client) {
@@ -160,7 +169,7 @@ class Pool extends EventEmitter {
 
   connect(cb) {
     if (this.ending) {
-      const err = new Error('Cannot use a pool after calling end on the pool')
+      const err = new PoolError('Cannot use a pool after calling end on the pool', 'Z0001')
       return cb ? cb(err) : this.Promise.reject(err)
     }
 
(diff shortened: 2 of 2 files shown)


packages/pg-pool/README.md
```diff
@@ -331,6 +331,20 @@ var bluebirdPool = new Pool({
 
 __please note:__ in node `<=0.12.x` the pool will throw if you do not provide a promise constructor in one of the two ways mentioned above.  In node `>=4.0.0` the pool will use the native promise implementation by default; however, the two methods above still allow you to "bring your own."
 
+## error management
+
+Errors throwed by the pool are instances of `PoolError`. Each `PoolError` has a message (short description of the exception) and a structured code composed of 5 characters string starting with letter `Z`.
```
charmander, line 336: - How about meaningful text like the error codes in the Node standard library?
- I’m not sure if most of these deserve their own code. Nobody should be handling library misuse like multiple release, multiple end, use after end, or type errors programmatically. “Timeout exceeded when trying to connect” makes sense, though.
- throwed → thrown

packages/pg-pool/index.js
```diff
@@ -408,3 +419,5 @@ class Pool extends EventEmitter {
   }
 }
 module.exports = Pool
+module.exports.Pool = Pool
+module.exports.PoolError = PoolError
```
charmander, line 423: Putting `Pool` on itself doesn’t feel right pre-ESM, and it’s not easy to use via pg. Would this even need to be exported with non-colliding error codes like `ERR_PG_POOL_…`?

packages/pg-pool/index.js
```diff
@@ -58,6 +58,15 @@ function makeIdleListener(pool, client) {
   }
 }
 
+class PoolError extends Error {
+  // A PoolError is an error throwed during a Pool process. Each type of error contains a specific "code".
+  // Please document in README file the error codes.
```
charmander, line 63:
```suggestion
  // A PoolError is an error thrown during a Pool operation. Each type of error contains a specific "code", documented in pg-pool’s README.
```

charmander (changes requested):

sehrope:
If we're going to have custom Error types, it'd be nice to have a single hierarchy for all things pg and PoolError (or maybe "PgPoolError"?) would be a child of a more generic PgError.

It makes testing easier too as you can check for a specific type (ex: "PgScramHandshakeError") rather than patterns of message text and have generic error handling for all things DB related in app code.

rmedaer:
> If we're going to have custom Error types, it'd be nice to have a single hierarchy for all things pg and PoolError (or maybe "PgPoolError"?) would be a child of a more generic PgError.

Indeed, currently the `PoolError` I implemented inherits std `Error`. The errors coming from `pg-protocol` are also inheriting std `Error`. However they implement interface `NoticeOrError` which is really specific to postgresql protocol and its backend. I do not have so much information and btw the _Pool_ ~is~ should not be so specific to postgresql backend. Indeed, as far as I know, postgresql is not aware if you use a _Pool_ or not. I'm not sure we should "mix" errors from postgresql vs internal library errors.

Having said that, you maybe saw that I'm using the same `code` keyword in `PoolError` with the same pattern as [postgresql error codes](https://www.postgresql.org/docs/13/errcodes-appendix.html). As far as I know there is no reserved codes. I had to choose a "prefix": `Z`.

As a conclusion, I understand the goal of _"a single hierarchy for all things pg and PoolError"_ and I would enjoy to implement it even if I think it's not really required. Could you help us on this @brianc  ?

packages/pg-pool/index.js
rmedaer: Do you mean: replacing `PoolError` by constants (for instance `ERR_PG_POOL_CONNECT_TIMEOUT`, `ERR_PG_POOL_UNEXPECTED_CONDITION`, ...) ?

My goal was to implement the same kind of structure than `NoticeOrError` interface from `pg-protocol`. What would be the PROs/CONs ?

packages/pg-pool/README.md
rmedaer: > How about meaningful text like the error codes in the Node standard library?

Current implementation mimic the `NoticeOrError` interface from `pg-protocol`, cf [this message](https://github.com/brianc/node-postgres/pull/2498/files#r663694243 )

> Nobody should be handling library misuse like multiple release, multiple end, use after end, or type errors programmatically.

That's why it raises ~exceptions~ errors.

> “Timeout exceeded when trying to connect” makes sense, though.

That's actually the only one I want to catch for now. For instance in a HTTP API, my goal is to return (a more) meaningful message to the client (in this case a `503` which means that the service is unavailable for now).

charmander: I don’t think mimicking the format of PostgreSQL error codes (with values that don’t come from PostgreSQL) is useful. https://nodejs.org/api/errors.html#nodejs-error-codes

packages/pg-pool/index.js
matthieusieben: I believe what @charmander is trying to say, is that if, instead of error codes like `Z0001`, you used [NodeJS like](https://nodejs.org/api/errors.html#nodejs-error-codes) error codes, you would not need to do `instanceof PoolError` and you could simply perform a check like:

```js
catch (err) {
  if (err.code === 'ERR_PG_POOL_UNEXPECTED_STATE') throw err // something's wrong
  if (err.code === 'ERR_PG_POOL_CONNECTION_TIMEOUT') // handle overloaded database
}
```
In this case, you would not even need to export `PoolError` at all.

Note that because of the way `node_modules` are structured, checking for `instanceof` might yield false negative in case `pg-pool` ends up twice in you `node_modules` hierarchy, making this argument 100% relevant.

This also explain his [other comment](https://github.com/brianc/node-postgres/pull/2498/files#r597562408) explaining that you don't need to explicit every possible error case. A simple `ERR_PG_POOL_UNEXPECTED_STATE` (plz find a better name 😅) is probably enough.

I believe that his first comment is that this feels right in ESM:
```js
export default Pool
export { Pool, PoolError }
```
but the following is weird in common-js (pre-ESM)
```js
module.exports = Pool
module.exports.Pool = Pool
```
because it is the same as doing `Pool.Pool = Pool` which feels odd.

PR #2501, opened 2021-03-25
fix(pg-cursor): EventEmitter memory leak

Eronana:
You will got warning `MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 rowDescription listeners added to [Connection]. Use emitter.setMaxListeners() to increase limit` when use cursor for insert without returning.

This PR can fix the issue.

### DIFF
--- a/packages/pg-cursor/index.js
+++ b/packages/pg-cursor/index.js
@@ -28,6 +28,9 @@ util.inherits(Cursor, EventEmitter)
 Cursor.prototype._ifNoData = function () {
   this.state = 'idle'
   this._shiftQueue()
+  if (this.connection) {
+    this.connection.removeListener('rowDescription', this._rowDescription)
+  }
 }
 
 Cursor.prototype._rowDescription = function () {


packages/pg-cursor/index.js
```diff
@@ -28,6 +28,9 @@ util.inherits(Cursor, EventEmitter)
 Cursor.prototype._ifNoData = function () {
   this.state = 'idle'
   this._shiftQueue()
+  if (this.connection) {
```
charmander, line 31: Is it possible for `this.connection` to be falsy at this point, or was this just to match `_rowDescription`?

charmander (approved):

packages/pg-cursor/index.js
Eronana: I have no idea, because I am not very familiar with this project. I just follow what `_rowDescription` did to fix the problem and it works.

brianc:
Thanks for this!

PR #2507, opened 2021-04-01
util in connection not used

joksnet:
Removing `require('util')` that is not used.


sehrope:
LGTM.

Though I wonder why this wasn't caught by the linter. I'm gonna open a separate issue for that.

charmander (approved):

PR #2517, opened 2021-04-10
Require users to pass certs when PG environment variable PGSSLMODE=require/verify-ca/verify-full

RichardJCai:
Previously if `readSSLConfigFromEnvironment` was used to generate the SSL config, the ssl object would not be created to create certs.

I ran into this issue when trying to execute yarn tests to a secure cluster (SSLMODE=require), certs would never be read even when I specified `PGSSLCERT, PGSSLKEY, PGSSLROOTCERT`.

This fix allows the user to pass in certs through PG env variables.

I'm not great with JS so apologies in advance

## Why this is necessary.

To highlight this issue in a simple case

Running the script I attached below through `PGSSLMODE=require PGSSLCERT=/home/ubuntu/certs/client.testuser.crt PGSSLROOTCERT=/home/ubuntu/certs/ca.crt PGSSLKEY=/home/ubuntu/certs/client.testuser.key PGHOST=localhost PGPORT=26257 PGUSER=root node test.js` results in the following error:

```
const {Client, Pool} = require("./packages/pg")

const client = new Client()
client.connect(err => {
if (err) {
    console.error('error connecting', err.stack)
} else {
    console.log('connected')
    client.end()
}
})

const pool = new Pool()
pool
.connect()
.then(client => {
    console.log('connected')
    client.release()
})
.catch(err => console.error('error connecting', err.stack))
.then(() => pool.end())
```

```
error connecting Error: unable to verify the first certificate
    at TLSSocket.onConnectSecure (_tls_wrap.js:1497:34)
    at TLSSocket.emit (events.js:315:20)
    at TLSSocket._finishInit (_tls_wrap.js:932:8)
    at TLSWrap.ssl.onhandshakedone (_tls_wrap.js:706:12)
error connecting Error: unable to verify the first certificate
    at TLSSocket.onConnectSecure (_tls_wrap.js:1497:34)
    at TLSSocket.emit (events.js:315:20)
    at TLSSocket._finishInit (_tls_wrap.js:932:8)
    at TLSWrap.ssl.onhandshakedone (_tls_wrap.js:706:12)
```

This is because the ssl field in `ConnectionParameters` is simply set to true and the certs fields are not populated

```
ConnectionParameters {
  user: 'root',
  database: 'root',
  port: 26257,
  host: 'localhost',
  binary: false,
  options: undefined,
  ssl: true,
  client_encoding: '',
  replication: undefined,
  isDomainSocket: false,
  application_name: undefined,
  fallback_application_name: undefined,
  statement_timeout: false,
  idle_in_transaction_session_timeout: false,
  query_timeout: false,
  connect_timeout: 0
}
```


charmander (changes requested):
They shouldn’t be required, just supported.

RichardJCai:
> They shouldn’t be required, just supported.

Gotcha, thought erroring would be better since if certs aren't passed in with the SSLMODEs then they won't be able to connect and getting a somewhat vague 
```
error connecting Error: unable to verify the first certificate
    at TLSSocket.onConnectSecure (_tls_wrap.js:1497:34)
    at TLSSocket.emit (events.js:315:20)
    at TLSSocket._finishInit (_tls_wrap.js:932:8)
    at TLSWrap.ssl.onhandshakedone (_tls_wrap.js:706:12)
error connecting Error: unable to verify the first certificate
    at TLSSocket.onConnectSecure (_tls_wrap.js:1497:34)
    at TLSSocket.emit (events.js:315:20)
    at TLSSocket._finishInit (_tls_wrap.js:932:8)
    at TLSWrap.ssl.onhandshakedone (_tls_wrap.js:706:12)
```
is confusing since theres actually no certs to update. As far as I can tell, certs won't be populated anywhere when `readSSLConfigFromEnvironment ` is called. 

Do you think logging a warning is appropriate? It certainly would've helped me out.

RichardJCai:
Updated so that it'll return an object with the cert properties. 

Please let me know if this is reasonable, mostly the part about returning undefined if the environment variable is not present for the certs.

dyllandry:
I just tried passing a cert via env var `PGSSLROOTCERT` and it doesn't work. From reading the node-postgres docs it seems like it should.

> [node-postgres uses the same environment variables as libpq to connect to a PostgreSQL server.](https://node-postgres.com/features/connecting)

rafiss:
The CI here is failing because https://github.com/brianc/node-postgres/pull/3110 needs to be merged

But the PR is ready for review.

rafiss:
The CI here is failing because https://github.com/brianc/node-postgres/pull/3110 needs to be merged (cc @brianc)

But this PR is ready for review.

PR #2518, opened 2021-04-11
chore(pg-query-stream): remove through dependency

fpusch:
I was browsing through the tests for the `pg-query-stream` package to see how to wire a transform stream to the output and the dependency seemed unnecessary to me. Hence this PR is removing it and replacing it with a [stream.Transform](https://nodejs.org/docs/latest/api/stream.html#stream_new_stream_transform_options).

If you think changes like this are worthwhile I can have a look through the other packages for similar "easy" PRs to get started contributing to `pg`.


brianc:
hey this is cool thank you1

PR #2529, opened 2021-04-26
Remove broken test

charmander:
It’s missing `co.wrap`, so it doesn’t actually run (Mocha does nothing with the paused generator). The test group that follows it, “using an ended pool”, covers the same thing anyway.

### DIFF
--- a/packages/pg-pool/test/error-handling.js
+++ b/packages/pg-pool/test/error-handling.js
@@ -65,18 +65,6 @@ describe('pool error handling', function () {
     })
   })
 
-  describe('calling connect after end', () => {
-    it('should return an error', function* () {
-      const pool = new Pool()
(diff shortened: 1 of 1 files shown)


sehrope:
+1 for removing the non-running test.

FYI, the subsequent test never checks that the pool worked in the first place so the `connect()` and `query()` could fail for any reason, not just because the pool has been ended already. The second `.end()` should still legitimately fail though.

charmander:
> FYI, the subsequent test never checks that the pool worked in the first place so the `connect()` and `query()` could fail for any reason, not just because the pool has been ended already.

Yep! I hope to improve that in another PR soon, and check for specific errors instead of just any error to boot. (In the meantime, the tests above at least test that *some* pool instances work.)

brianc:
thanks y'all - appreciate it.

PR #2534, opened 2021-04-30
Making package compatible with ESM

mitar:
In recent version of node, node is using [cjs-module-lexer](https://github.com/guybedford/cjs-module-lexer) to try to detect names of exports in CJS modules. It is doing static analysis so it cannot detect everything. Previously this package was using a function which throw things off, but I do not there was much benefit of having that, or am I missing something? I simplified it into a simple object and now node can import it as a ESM module. Tested with node v16.0.0. E.g.:

```
import { Client } from 'pg';
```

### DIFF
--- a/packages/pg/lib/index.js
+++ b/packages/pg/lib/index.js
@@ -14,21 +14,23 @@ const poolFactory = (Client) => {
   }
 }
 
-var PG = function (clientConstructor) {
-  this.defaults = defaults
-  this.Client = clientConstructor
-  this.Query = this.Client.Query
-  this.Pool = poolFactory(this.Client)
-  this._pools = []
-  this.Connection = Connection
-  this.types = require('pg-types')
-  this.DatabaseError = DatabaseError
+module.exports = {
+  defaults: defaults,
+  _pools: [],
+  Connection: Connection,
+  types: require('pg-types'),
+  DatabaseError: DatabaseError,
 }
 
 if (typeof process.env.NODE_PG_FORCE_NATIVE !== 'undefined') {
-  module.exports = new PG(require('./native'))
+  var nativeClient = require('./native')
+  module.exports.Client = nativeClient
+  module.exports.Query = nativeClient.Query
+  module.exports.Pool = poolFactory(nativeClient)
 } else {
-  module.exports = new PG(Client)
+  module.exports.Client = Client
+  module.exports.Query = Client.Query
+  module.exports.Pool = poolFactory(Client)
 
   // lazy require native module...the native module may not have installed
   Object.defineProperty(module.exports, 'native', {


charmander (changes requested):
If the name `exports` is special, this seems like a hack. How about index.mjs and the appropriate package.json `exports` field?

mitar:
The whole `cjs-module-lexer` is a hack/workaround. But it has standard behaviors/structures which are documented and standardized and will stay the same. See [the one I am adding in this PR here](https://github.com/guybedford/cjs-module-lexer#named-exports-parsing).

The problem with `exports` in `package.json` is that it would require a major semver bump, namely after you add it, it is not possible anymore to import things from inside the package, just at the top level (unless we want to list all files manually there and keep them in sync all the time). See more details [here](https://redfin.engineering/node-modules-at-war-why-commonjs-and-es-modules-cant-get-along-9617135eeca1#a013):

> Beware: adding an exports map is always a “semver major” breaking change. By default, your users can reach into your package and require() any script they want, even files that you intended to be internal. The exports map ensures that users can only require/import the entry points that you deliberately expose.

I feel this is a rather drastic change. When the one I made works and officially supported by node, while stays backwards compatible.

Let's not make perfect the enemy of the good.

charmander:
> The problem with exports in package.json is that it would require a major semver bump

Seems appropriate enough?

mitar:
Just to be clear. I do not mind doing it as you are asking. I wanted to make it as simple as possible to make it easier to get it merged. If you are saying that I should do the full thing to get it merged, sure. I just do not want to spend super extra time and then this does not get merged (is one of those PRs which can get hard to keep merge-ready if it is not merged soon after it is made).

So to make it clear:
* You would prefer if I properly make it ESM compatible. My understanding is that we still want to keep it CJS compatible. I can do that.
* Major version bump is OK. Is this something I do in this PR or do you bump yourself after merging?
* Do you want this also on all other packages in this monorepo? I can update all of them in the same way.

benjamingwynn:
This would be a nice-to-have for us, is this still able to be merged with a major version bump to pg@8?

aleclarson (approved):

marziply:
What's preventing this from being merged in? Seems like a simple feature with ergonomic benefits given the shift from `require` to `import`.

brianc:
> What's preventing this from being merged in?

I can't tell if this contains breaking changes or not. Does this release require a semver major bump? Is there any way to write tests for both `require` and `imports` style of imports from JS?

aleclarson:
There's no breaking change here.

All it's doing is allowing `cjs-module-lexer` (used by Node for CJS compatibility in ESM environments) to statically detect the exports added in the `PG` constructor.

I just tested the PR locally and it works in both CJS and ESM environments. I think the chance of a regression in this behavior is incredibly small and thus the benefit of an automated test is negligible.

PR #2550, opened 2021-05-25
Add Cursor to index.js's module.exports

Bluenix2:
Even though the documentation refers to it as [pg.Cursor](https://node-postgres.com/api/cursor) it isn't actually defined in the library's index.js.

I don't completely get the argument why it wasn't added in the first place, but I am going to guess that it's because not everyone uses Cursors. I thought this would be an alright compromise between that.

One downside with this change, is the somewhat unexpected behaviour one might experience if they try to use `pg.Cursor` without `pg-cursor` installed. Best way to fix that downside is to simply add `pg-cursor` to the dependencies, and include it like Pool, but this might be unwanted?

### DIFF
--- a/packages/pg/lib/index.js
+++ b/packages/pg/lib/index.js
@@ -23,6 +23,14 @@ var PG = function (clientConstructor) {
   this.Connection = Connection
   this.types = require('pg-types')
   this.DatabaseError = DatabaseError
+
+  try {
+    this.Cursor = require('pg-cursor')
+  } catch (err) {
+    if (err.code !== 'MODULE_NOT_FOUND') {
+      throw err
+    }
+  }
 }
 
 if (typeof process.env.NODE_PG_FORCE_NATIVE !== 'undefined') {


charmander (changes requested):
I think it’d be better to fix the documentation instead. Even the way `pg.Pool` is included isn’t the cleanest, but it at least has a reason in that `pg.Pool` is a `Pool` type bound to `pg.Client` (i.e. it’s not exactly the same as the `Pool` you import from `pg-pool`).

brianc:
Thanks for doing this, but I agree with @charmander  I'd rather change the documentation than export the cursor in the main package.  It's an optional additional install & separate module.  

There's an argument to be made for including it and exporting it, but if that were the case I'd take a hard dependency on it so I could keep the versions locked together w/ the main `pg` package and `pg-cursor`.

Bluenix2:
> There's an argument to be made for including it and exporting it, but if that were the case I'd take a hard dependency on it so I could keep the versions locked together w/ the main `pg` package and `pg-cursor`.
@brianc

Is there a reason for this not to be done so far? I suppose it's not as widely used?

brianc:
yeah it's not as widely used and allows revving a major semver on the
cursor lib w/o having to do a major semver of pg as a whole.  I think the
thing pg-cursor is likely most widely used for is actually pg-query-stream
(which uses it internally).

PR #2553, opened 2021-05-28
Turn Cursor into an ES6 class

Bluenix2:
Unlike most other objects Cursor has remained unchanged to an ES6 class. As somewhat expected the git diff is kind of a mess when changes like these happen, I don't think there's anything that can be done considering everything had to be indented an additional time.

### DIFF
--- a/packages/pg-cursor/index.js
+++ b/packages/pg-cursor/index.js
@@ -190,13 +190,15 @@ class Cursor extends EventEmitter {
 
   // users really shouldn't be calling 'end' here and terminating a connection to postgres
   // via the low level connection.end api
-  end = util.deprecate(function (cb) {
-    if (this.state !== 'initialized') {
-      this.connection.sync()
-    }
-    this.connection.once('end', cb)
-    this.connection.end()
-  }, 'Cursor.end is deprecated. Call end on the client itself to end a connection to the database.')
+  end(cb) {
+    return util.deprecate(function () {
+      if (this.state !== 'initialized') {
+        this.connection.sync()
+      }
+      this.connection.once('end', cb)
+      this.connection.end()
+    }, 'Cursor.end is deprecated. Call end on the client itself to end a connection to the database.')()
+  }
 
   close(cb) {
     if (!this.connection || this.state === 'done') {
--- a/packages/pg-cursor/index.js
+++ b/packages/pg-cursor/index.js
@@ -7,7 +7,6 @@ const util = require('util')
 let nextUniqueID = 1 // concept borrowed from org.postgresql.core.v3.QueryExecutorImpl
 
 class Cursor extends EventEmitter {
-
   constructor(text, values, config) {
     super()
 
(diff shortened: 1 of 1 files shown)


packages/pg-cursor/index.js
```diff
@@ -6,230 +6,232 @@ const util = require('util')
...
+      rows: rows,
+    }
+    this.connection.execute(msg, true)
+    this.connection.flush()
+  }
+
+  // users really shouldn't be calling 'end' here and terminating a connection to postgres
+  // via the low level connection.end api
+  end(cb) {
+    return util.deprecate(function () {
```
charmander, line 193: This will print a deprecation warning every time it’s called. How about, after the class,

```js
Cursor.prototype.end = util.deprecate(Cursor.prototype.end, …)
```

charmander (changes requested):

packages/pg-cursor/index.js
Bluenix2: Yeah I thought about that, but it's pretty bad practice to mix. I was also thinking of putting it in the constructor, but I think that might be worse?

charmander: Yes, putting it in the constructor would be wrong. The line I suggested isn’t bad practice.

Bluenix2: Not in of itself, but mixing it. I suppose that does seem to be the best option all things considered. I'll change that tomorrow.

packages/pg-cursor/index.js
```diff
@@ -6,231 +6,232 @@ const util = require('util')
...
+    }
   }
-  this.connection.execute(msg, true)
-  this.connection.flush()
 }
 
-// users really shouldn't be calling 'end' here and terminating a connection to postgres
-// via the low level connection.end api
+// we can't correctly use util.deprecate in an ES6 method
 Cursor.prototype.end = util.deprecate(function (cb) {
```
charmander, line 227: You can keep defining it inside the class, it just needs this afterwards:

```js
Cursor.prototype.end = util.deprecate(Cursor.prototype.end, 'Cursor.end is deprecated. Call end on the client itself to end a connection to the database.')
```

charmander (changes requested):

packages/pg-cursor/index.js
Bluenix2: Oh perfect, that's much better!

charmander (approved):
Thanks!

brianc:
Sorry for the delay here - really appreciate this!  Merging now.  You marked this as a breaking change but none of the tests needed to change - is this a semver major thing?  I'm fine either way & can release it however you think is best.

Bluenix2:
> You marked this as a breaking change but none of the tests needed to change - is this a semver major thing? I'm fine either way & can release it however you think is best.

Did I? This shouldn't be breaking yes.

PR #2554, opened 2021-05-28
Add support for using promises in Cursor methods

### LINKED ISSUE #2545: Why does pg.Cursor still use callbacks?
Looking at the [the pg.Cursor API documentation](https://node-postgres.com/api/cursor) it appears to be using callbacks, is there a reason that there hasn't been a switch to the Promises API introduced in ES6?

Other parts of the package shows the usage of promises. As can be seen [here](https://node-postgres.com/features/queries).

Bluenix2:
Fixes #2545. This change was simply mirrored from [Client.query](https://github.com/brianc/node-postgres/blob/master/packages/pg/lib/client.js#L496-L578).

I did base this pull request on #2553, expecting that to also be merged (that way no merge conflicts arise). If this is unwanted I could change that.

### DIFF
--- a/packages/pg-cursor/test/promises.js
+++ b/packages/pg-cursor/test/promises.js
@@ -1,0 +1,60 @@
+const assert = require('assert')
+const Cursor = require('../')
+const pg = require('pg')
+
+const text = 'SELECT generate_series as num FROM generate_series(0, 5)'
+
+describe('cursor using promises', function () {
+  beforeEach(function (done) {
+    const client = (this.client = new pg.Client())
+    client.connect(done)
+
+    this.pgCursor = function (text, values) {
+      return client.query(new Cursor(text, values || []))
+    }
+  })
+
+  afterEach(function () {
+    this.client.end()
+  })
+
+  it('resolve with result', function (done) {
+    const cursor = this.pgCursor(text)
+    cursor
+      .read(6)
+      .then((res) => assert.strictEqual(res.length, 6))
+      .error((err) => assert.ifError(err))
+      .finally(() => done())
+  })
+
+  it('reject with error', function (done) {
+    const cursor = this.pgCursor('select asdfasdf')
+    cursor.read(1).error((err) => {
+      assert(err)
+      done()
+    })
+  })
+
+  it('read multiple times', async function (done) {
+    const cursor = this.pgCursor(text)
+    let res
+
+    try {
+      res = await cursor.read(2)
+      assert.strictEqual(res.length, 2)
+
+      res = await cursor.read(3)
+      assert.strictEqual(res.length, 3)
+
+      res = await cursor.read(1)
+      assert.strictEqual(res.length, 1)
+
+      res = await cursor.read(1)
+      assert.strictEqual(res.length, 0)
+    } catch (err) {
+      assert.ifError(err)
+    } finally {
+      done()
+    }
+  })
+})
--- a/packages/pg-cursor/test/promises.js
+++ b/packages/pg-cursor/test/promises.js
@@ -18,13 +18,10 @@ describe('cursor using promises', function () {
     this.client.end()
   })
 
-  it('resolve with result', function (done) {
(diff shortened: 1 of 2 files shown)


Bluenix2:
> I did base this pull request on #2553, expecting that to also be merged (that way no merge conflicts arise). If this is unwanted I could change that.

I reverted this in the case I want to add more commits to that pull request. This means that currently this pull request will fail, but I'll rebase this onto master when that (possibly?) gets merged.

Bluenix2:
Thought I'd correctly mark this as draft until the pull request this currently depends on gets reviewed.

brianc:
Adding promises in here is totally cool...need some tests for the functionality though!

Bluenix2:
Yup I am going to get to that now!

brianc:
rad

Bluenix2:
When writing tests, prettier tries to enforce the following style:
```js
  it('fetch 6 when asking for 10', function (done) {
    const cursor = this.pgCursor(text)
    cursor
      .read(10)
      .then((res) => assert.strictEqual(res.length, 6))
      .error((err) => assert.ifError(err))
  })
```

When I'd expect the following instead:
```js
  it('fetch 6 when asking for 10', function (done) {
    const cursor = this.pgCursor(text)
    cursor.read(10)
      .then((res) => assert.strictEqual(res.length, 6))
      .error((err) => assert.ifError(err))
  })
```
Thoughts on this?

Bluenix2:
There should be tests for this now; I thought testing resolving, rejecting and multiple read() calls was sufficient.

packages/pg-cursor/test/promises.js
```diff
@@ -0,0 +1,60 @@
...
+      res = await cursor.read(1)
+      assert.strictEqual(res.length, 1)
+
+      res = await cursor.read(1)
+      assert.strictEqual(res.length, 0)
+    } catch (err) {
+      assert.ifError(err)
+    } finally {
+      done()
+    }
```
charmander, line 58: Can be simplified since no callbacks are involved:

```suggestion
  it('read multiple times', async () => {
    const cursor = this.pgCursor(text)
    let res

    res = await cursor.read(2)
    assert.strictEqual(res.length, 2)

    res = await cursor.read(3)
    assert.strictEqual(res.length, 3)

    res = await cursor.read(1)
    assert.strictEqual(res.length, 1)

    res = await cursor.read(1)
    assert.strictEqual(res.length, 0)
```

packages/pg-cursor/test/promises.js
```diff
@@ -0,0 +1,60 @@
...
+    this.client.end()
+  })
+
+  it('resolve with result', function (done) {
+    const cursor = this.pgCursor(text)
+    cursor
+      .read(6)
+      .then((res) => assert.strictEqual(res.length, 6))
+      .error((err) => assert.ifError(err))
+      .finally(() => done())
```
charmander, line 27: Similar to the other comment:

```suggestion
  it('resolve with result', async function () {
    const cursor = this.pgCursor(text)
    const res = await cursor.read(6)
    assert.strictEqual(res.length, 6)
```

brianc:
> When writing tests, prettier tries to enforce the following style:



> Thoughts on this?

I always just let prettier do the formatting.  Easier to not even have to care at all & everything looks consistent.

brianc (approved):
Thanks for this! looks good to me.

nit: you could use `let` instead of `var`.

Bluenix2:
> Thanks for this! looks good to me.
> 
> nit: you could use `let` instead of `var`.

I wasn't sure if it would cause issues with scoping since there's a bit of a difference, but that should be done now @brianc

Bluenix2:
@brianc Is there anything hindering this from getting merged?

brianc:
This is great really sorry for the delay - life is crazy 💻 🐱  I'll get a semver minor out today w/ the changes. Thank you so much! ❤️

PR #2567, opened 2021-06-30
Async password function

kyawlinmyat:
Support password async function.


charmander (changes requested):
Already supported by the nature of `Promise#then`.

Issue #2078, opened 2020-01-18
unref setTimeout use in pg-pool


gajus:
In at least two cases, `pg-pool` creates a referenced timeout:

1. https://github.com/brianc/node-postgres/blob/3f6760c62ee2a901d374b5e50c2f025b7d550315/packages/pg-pool/index.js#L292-L295
2. https://github.com/brianc/node-postgres/blob/3f6760c62ee2a901d374b5e50c2f025b7d550315/packages/pg-pool/index.js#L171-L177

These timeouts are currently keeping the process from dying naturally.

charmander:
I *think* (1) is cleared by `pool.end()` (which is required for a process to end naturally). (2) is cleared by a successful connection, also required for a process to end naturally.

Have you seen a bug (in a minimal script) where pool timeouts are keeping the process running after ending the pool?

fluggo:
It occurred to me that this could be useful. The http.Agent class in Node.js unrefs sockets that are not currently handling an HTTP request, so they stick around in memory, but they don't stop the process from exiting if they're not actively being used.

So the question is: can pg-pool unref sockets or timeouts if they're not currently in use? Can I create a pool that will let the process end without calling pool.end() first?

charmander:
@fluggo Yep, the pool could do that.

PR #2568, opened 2021-07-02
Allow Node to exit if the pool is idle

fluggo:
Based on the suggestion from #2078.

Unfortunately I don't yet know how to test this under Mocha, as the whole point is to let Node exit, but I did test it with the following script:

```javascript
const Pool = require('./index')

const pool = new Pool({ idleTimeoutMillis: 1000 })
pool.query('SELECT NOW()', (err, res) => console.log(res))
pool.on('remove', () => {
  console.log('removed')
  done()
})

setTimeout(() => {
  pool.query('SELECT * from generate_series(0, 1000)', (err, res) => console.log(res))
}, 500)
```

The program waits for the idle timeout during the time between the first query and the second, but exits immediately after the second query has finished running. After the second query, Node's loop is empty, and since the idle timeout and the one connected socket are unref'd, Node can exit right away.

If anyone has any suggestions on how to test this under Mocha, I'm all ears.

### DIFF
--- a/packages/pg-pool/index.js
+++ b/packages/pg-pool/index.js
@@ -83,6 +83,7 @@ class Pool extends EventEmitter {
 
     this.options.max = this.options.max || this.options.poolSize || 10
     this.options.maxUses = this.options.maxUses || Infinity
+    this.options.allowExitOnIdle = this.options.allowExitOnIdle || false
     this.log = this.options.log || function () {}
     this.Client = this.options.Client || Client || require('pg').Client
     this.Promise = this.options.Promise || global.Promise
@@ -136,6 +137,7 @@ class Pool extends EventEmitter {
       const idleItem = this._idle.pop()
       clearTimeout(idleItem.timeoutId)
       const client = idleItem.client
+      client.ref()
       const idleListener = idleItem.idleListener
 
       return this._acquireClient(client, pendingItem, idleListener, false)
@@ -323,6 +325,15 @@ class Pool extends EventEmitter {
         this.log('remove idle client')
         this._remove(client)
       }, this.options.idleTimeoutMillis)
+
+      if (this.options.allowExitOnIdle) {
+        // allow Node to exit if this is all that's left
+        tid.unref()
+      }
+    }
+
+    if (this.options.allowExitOnIdle) {
+      client.unref()
     }
 
     this._idle.push(new IdleItem(client, idleListener, tid))
--- a/packages/pg-pool/test/idle-timeout-exit.js
+++ b/packages/pg-pool/test/idle-timeout-exit.js
@@ -1,0 +1,16 @@
+// This test is meant to be spawned from idle-timeout.js
+if (module === require.main) {
+  const allowExitOnIdle = process.env.ALLOW_EXIT_ON_IDLE === '1'
+  const Pool = require('../index')
+
+  const pool = new Pool({ idleTimeoutMillis: 200, ...(allowExitOnIdle ? { allowExitOnIdle: true } : {}) })
+  pool.query('SELECT NOW()', (err, res) => console.log('completed first'))
+  pool.on('remove', () => {
+    console.log('removed')
+    done()
+  })
+
+  setTimeout(() => {
+    pool.query('SELECT * from generate_series(0, 1000)', (err, res) => console.log('completed second'))
+  }, 50)
+}
--- a/packages/pg-pool/test/idle-timeout.js
+++ b/packages/pg-pool/test/idle-timeout.js
@@ -4,6 +4,8 @@ const expect = require('expect.js')
 
 const describe = require('mocha').describe
 const it = require('mocha').it
+const { fork } = require('child_process')
+const path = require('path')
 
 const Pool = require('../')
 
@@ -84,4 +86,33 @@ describe('idle timeout', () => {
       return pool.end()
     })
   )
+
+  it('unrefs the connections and timeouts so the program can exit when idle when the allowExitOnIdle option is set', function (done) {
+    const child = fork(path.join(__dirname, 'idle-timeout-exit.js'), [], {
+      silent: true,
+      env: { ...process.env, ALLOW_EXIT_ON_IDLE: '1' },
+    })
+    let result = ''
+    child.stdout.setEncoding('utf8')
+    child.stdout.on('data', (chunk) => (result += chunk))
+    child.on('error', (err) => done(err))
+    child.on('close', () => {
+      expect(result).to.equal('completed first\ncompleted second\n')
+      done()
+    })
+  })
+
+  it('keeps old behavior when allowExitOnIdle option is not set', function (done) {
+    const child = fork(path.join(__dirname, 'idle-timeout-exit.js'), [], {
+      silent: true,
+    })
+    let result = ''
+    child.stdout.setEncoding('utf8')
+    child.stdout.on('data', (chunk) => (result += chunk))
+    child.on('error', (err) => done(err))
+    child.on('close', () => {
+      expect(result).to.equal('completed first\ncompleted second\nremoved\n')
+      done()
+    })
+  })
 })
--- a/packages/pg/lib/client.js
+++ b/packages/pg/lib/client.js
@@ -577,6 +577,14 @@ class Client extends EventEmitter {
     return result
   }
 
+  ref() {
+    this.connection.ref()
+  }
+
+  unref() {
+    this.connection.unref()
+  }
+
   end(cb) {
     this._ending = true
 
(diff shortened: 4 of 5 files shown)


charmander:
I wonder if it’d be better to introduce `client.ref()` and `client.unref()` and have the pool use those?

> If anyone has any suggestions on how to test this under Mocha, I'm all ears.

Maybe by spawning new processes?

fluggo:
> I wonder if it’d be better to introduce `client.ref()` and `client.unref()` and have the pool use those?

I thought about this, but I didn't know how to test across the packages.

> > If anyone has any suggestions on how to test this under Mocha, I'm all ears.
> 
> Maybe by spawning new processes?

Good idea. I'll give it a shot.

fluggo:
All done, ref/unref now appear on Client and Connection, and pg-pool uses them to maintain the pool. There is now also a Mocha test.

**HOWEVER 📢🧨️🚨** this now causes a new relationship between the next version of pg-pool and the next version of pg. pg-pool will need to depend on the version of pg that has the ref/unref in it.

fluggo:
Is there anything further that needs to be done here?

brianc:
> Is there anything further that needs to be done here?

I just need to take a look at this & get it merged & released! Should be able to look here today or tomorrow!

brianc:
okay I checked this out...looks pretty good!  I have a concern though - because this is a breaking change I'll need to rev a new semver major to release this in its current state.  I like to batch up breaking changes and limit the number of times I do that.  Would it be possible to make this an optional config on the pool like `const pool = new Pool({ allowExitOnIdle: true })` or something, and default it to false (existing behavior) - that way we can release this in a semver minor as soon as it's ready.

fluggo:
Is this a breaking change? It would be a very odd program that would fall prey to this—it would have to be relying on the thread pool's idle timer to keep the program alive while another unref'd timer or socket was doing work.

brianc:
Yeah...I know what you're saying 100%.  There's this synthesizer I make music with sometimes called [the deluge](https://synthstrom.com/product/deluge/).  It's a pretty incredible little device developed by just two folks from New Zealand.  One of it's features is when the CPU gets overloaded on the device instrument voices drop from playback (istead of distorting or stopping playback). The lead developer says he has to semver the releases when he makes the code more CPU efficient because people have compositions which accidentally rely on the CPU overload protection voice dropout, when the code gets better their music changes and they aren't happy.  pg has been installed 1.75 million times _this week_, and most likely there are at least a handful of programs which accidentally rely on the pool's idle timeout keeping their program alive to allow some other unreffed thing to complete in the background.  I'm okay adding the flag myself if you want to step away from this work for now - I'm already grateful for your contribution...but yeah I'd consider this a silent, sneaky, incredibly hard to debug "breaking" change to people who are relying on unspecified but up until now steady behavior of the library.  I'm also fine putting this behind a flag w/ the original behavior "off" and the flipping the default to "on" in the next semver major release.  With the size of the install base of this library for better & for worse I try to above most everything (outside security and performance) maintain backwards compatibility of everything outside of semver major bumps.

fluggo:
I had a feeling I wouldn't win that argument, but I thought I'd try!  😆 

Patch updated.

fluggo:
Are we good to go on this one?

brianc:
Yes! I was visiting parents this week out of state. Flying home today, will
merge + rlease tomrrow!

brianc:
will be releasing this this morning - sorry for the delay & thanks so much for the PR!

packages/pg-pool/index.js
```diff
@@ -136,6 +137,7 @@ class Pool extends EventEmitter {
       const idleItem = this._idle.pop()
       clearTimeout(idleItem.timeoutId)
       const client = idleItem.client
+      client.ref()
```
brianc, line 140:
```suggestion
      const client = idleItem.client
      if (this.options.allowExitOnIdle) {
        client.ref()
      }
```

don't we want the symmetry here to match line 331? Otherwise the stream on the connection will be refed an additional time and not unreffed?

brianc:
> will be releasing this this morning - sorry for the delay & thanks so much for the PR!

err one question I put above before I get this out

packages/pg-pool/index.js
fluggo: Ref/unref is a flag, not a counter. The stream comes ref'd by default, so ref'ing it more than once is harmless.

brianc: ohhh well then! I'll get this released here in a few min

fluggo:
Thanks! Excited to incorporate it into my projects!

brianc:
released! sorry for the delay!

vitaly-t:
Wonderful update, finally, kudos to @fluggo! I have updated `pg-promise` to support it.

e50:
> Yeah...I know what you're saying 100%. There's this synthesizer I make music with sometimes called [the deluge](https://synthstrom.com/product/deluge/). It's a pretty incredible little device developed by just two folks from New Zealand. One of it's features is when the CPU gets overloaded on the device instrument voices drop from playback (istead of distorting or stopping playback). The lead developer says he has to semver the releases when he makes the code more CPU efficient because people have compositions which accidentally rely on the CPU overload protection voice dropout, when the code gets better their music changes and they aren't happy. pg has been installed 1.75 million times _this week_, and most likely there are at least a handful of programs which accidentally rely on the pool's idle timeout keeping their program alive to allow some other unreffed thing to complete in the background. I'm okay adding the flag myself if you want to step away from this work for now - I'm already grateful for your contribution...but yeah I'd consider this a silent, sneaky, incredibly hard to debug "breaking" change to people who are relying on unspecified but up until now steady behavior of the library. I'm also fine putting this behind a flag w/ the original behavior "off" and the flipping the default to "on" in the next semver major release. With the size of the install base of this library for better & for worse I try to above most everything (outside security and performance) maintain backwards compatibility of everything outside of semver major bumps.


I see this is merged, but isn't calling `pool.end()` faster and better?

---

Also that story deprecates xkcd 1172


---

fluggo:
> I see this is merged, but isn't calling `pool.end()` faster and better?

Faster? Unlikely, as `pool.end()` has to drain the pool first (which admittedly shouldn't take long). When Node exits the main loop, [it does no object teardown at all](https://github.com/nodejs/node/blob/f52dc17519cb7ce0f36ce2e2667ec4b5a1cb7a93/src/api/embed_helpers.cc#L56), so it's up to the kernel to close any remaining connections, and that might happen after the process exits. That's not really the point of this commit, though.

Better? Depending on your programming style, you might think of `pool.end()` as cleaner, as you're forcing yourself to have a known path toward your program's exit so that you can call this function _once_ and _only after all tasks have completed_. I'm getting older and deciding not to be so masochistic. I have several programs where it's not clear when the right time _is_ to call `pool.end()`, such as servers that keep handling existing connections after the listener is closed. I also have scheduled tasks and utility programs that call into the pool, not to mention the regression tests, and rather than having to figure out the correct point in each program to call `pool.end()` (and _debugging_ and _debugging_ when I get it wrong), I'd much rather just have a pool that lets me exit naturally: when everything else I care about is done.

This has a precedent in several Node.js APIs. Sockets and timeouts can be unref'd to let the process exit naturally in spite of their existence, and the http.Agent object unrefs all of its idle sockets without asking—to do otherwise would require you to manually clean up even the default Agent after any HTTP calls, which I think breaks the feel of Node.js as a scripting runtime.

e50:
> I have servers that keep handling existing connections after I close the server's listener...

OK. The feature is to gracefully stop servers without a `cluster`

> not to mention the regression tests

FWIW, in Mocha, it's possible to:
```js
function serveUntilDone() {
  const server = http.createServer(router);
  before(done => {
    initDatabasePool();
    server.listen(serverAddr, done);
  })
  after(done => {
    db.pool.end();
    server.close(done);
  })
```

fluggo:
Yes, I have that same code in my mocha tests. I still have trouble getting it to exit gracefully. I'm looking to keep the hair I have.

If you have a point in favor of making this harder to use than http.Agent, say so.

e50:
> If you have a point in favor of making this harder to use than http.Agent, say so.

I just wanted to know if there was something I needed to worry about this feature.

karlbecker:
Just wanted to add a note here saying thank you for avoiding making this a breaking feature @fluggo  - and for the nice story about the synthesizer @brianc 😊  

> With the size of the install base of this library for better & for worse I try to above most everything (outside security and performance) maintain backwards compatibility of everything outside of semver major bumps.

Thank you x 1000

fluggo:
Oh hey! And it's in the docs now! Kudos!

Issue #2439, opened 2021-01-10
pool.on('error) doesnt populate client errors as well


ynahmany:
Hi, 
I came across a strange behavior recently.
Env setup:

Linux- Ubuntu 18.
Docker - 19.06
Node and Postgres running inside docker using docker-compose.

When I manually restarted(or shut down) Postgres docker -> the Node app crashes with Error: 
```
Error: server closed the connection unexpectedly
 This probably means the server terminated abnormally
 before or while processing the request.
    at module.exports.Client._readError (/workspaces/node_modules/pg-native/index.js:154:13)
    at module.exports.Client._read (/workspaces/node_modules/pg-native/index.js:203:17)
    at PQ.emit (events.js:315:20)
Emitted 'error' event on  instance at:
    at module.exports.<anonymous> (/workspaces/node_modules/pg/lib/native/client.js:99:14)
at module.exports.emit (events.js:315:20)
at module.exports.Client._readError (/workspaces/node_modules/pg-native/index.js:155:8)
     at module.exports.Client._read (/workspaces/node_modules/pg-native/index.js:203:17)
at PQ.emit (events.js:315:20)

```
I configured the `on.error(callback)` listener yet I keep getting this error which causes my node application to **crash**

The solution was to add `on.error` to the client that is being returned from the callback above. 

```
export const pool = new pg.native.Pool(config);

const handleConnectionFailure = (e: Error, _client: pg.PoolClient) => {
     _client.on('error', (err) => { // did the trick );
});
pool.connect(handleConnectionFailure);
pool.on('error', (err) => {
    // didnt help
});
```

charmander:
That’s right – you need to add error handlers when checking clients out of the pool, in addition to the pool-level error handler, *and* remove them before returning clients to the pool. Correct error handling isn’t very easy with pg at the moment.

ynahmany:
@charmander Thanks for the quick reply.

For future use, this one also work perfectly:
```
export let pool = new pg.native.Pool(config);

pool.on('connect', (_client: pg.PoolClient) => {
  // On each new client initiated, need to register for error(this is a serious bug on pg, the client throw errors although it should not)
  _client.on('error', (err: Error) => {
    console.log(err);
  });
});
```

jszopi:
I notice that @ynahmany 's comment with the solution was linked from other issues and got quite a few :+1:s I worry that the comment there:

> (this is a serious bug on pg, the client throw errors although it should not)

[...]

PR #2569, opened 2021-07-03
Catch errors client throws in pool

zlotnika:
Old behavior: if you do something like

```js
try {
  await (new Pool()).query(null);
} catch (error) {
  console.log(error);
}
```

you get an uncaught exception. If you have a server that generates text to pass in to query the pool, it will crash your server.

New behavior: This change catches the error thrown at https://github.com/zlotnika/node-postgres/blob/master/packages/pg/lib/client.js#L505 and treats it the same as it does with any other. This way, a query of `null` works similarly to a query of `'over 9000'` or `new Date()` or whatever other not-so-SQL-y thing you come up with.

An alternate solution would be to call the callback with an error in the client instead of throwing an error. I personally prefer throw-catching.

### DIFF
--- a/packages/pg-pool/index.js
+++ b/packages/pg-pool/index.js
@@ -375,20 +375,24 @@ class Pool extends EventEmitter {
 
       client.once('error', onError)
       this.log('dispatching query')
-      client.query(text, values, (err, res) => {
-        this.log('query dispatched')
-        client.removeListener('error', onError)
-        if (clientReleased) {
-          return
-        }
-        clientReleased = true
-        client.release(err)
-        if (err) {
-          return cb(err)
-        } else {
-          return cb(undefined, res)
-        }
-      })
+      try {
+        client.query(text, values, (err, res) => {
+          this.log('query dispatched')
+          client.removeListener('error', onError)
+          if (clientReleased) {
+            return
+          }
+          clientReleased = true
+          client.release(err)
+          if (err) {
+            return cb(err)
+          } else {
+            return cb(undefined, res)
+          }
+        })
+      } catch (err) {
+        return cb(err)
+      }
     })
     return response.result
   }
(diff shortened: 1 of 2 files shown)


benk79:
IMHO this is a crucial point for an application to run in production. Error management is really a must, not just a nice to have. Tryed to see if i can write a test, but had so many already existing tests failing that gave up.

benk79:
My exact problem being almost the same as @bryanph but also for DB server being shut down or having failures, i finally ended with this simple solution

```js

pool.on('connect', client => {
    client.on('error', (err)=> {
        pool.emit('error', err, client);
    });
});

```

zlotnika:
I've tried writing a test, but I can't seem to get testing working on my machine. It looks like there's a lot of stuff set up for this, but no clear documentation. Could anyone either 1. write a test for me or 2. add something on running tests to the Readme or 3. tell me where else to find info on it?

zlotnika:
Alright. I've added a test that should work right. My environment is a bit different, so hopefully it'll run correctly on travis etc. Let me know if it works or whatever. Feel free to suggest fixes for it!

josh-cain:
@brianc Is this PR still being considered?  Have a problem similar to #2439 and was hoping to get something here in the upstream rather than fix elsewhere.

bbeesley:
Hey @brianc is there any chance of some feedback on this PR? We've been suffering this issue using node-postgres with big aws aurora clusters where I work. We keep getting paged about it since we're getting containers dying in production every 15 mins or so. Is there any reason not to merge this PR? If you need better test coverage then I'm happy to help @zlotnika out there. Just hoping to be able to avoid having to fork this repo and publish a patched version of it since it's overhead I'd much rather do without.

brianc:
yes i am sorry - you're absolutely right this should be merged, and I will do that now & get a new release out tomorrow.  There's some epic storm in Austin right now so...my power is going in and out but i'll get ya patched up tomorrow along w/ a salvo of other fixes.  Sorry for the delay!

PR #2573, opened 2021-07-15
pg-query-stream: make rows available as soon as possible

petrosagg:
pg-cursor buffers the rows until the specified number of them are accumulated before firing the callback. However, in the streaming setting is preferable to emit the rows as soon as possible.

This patch hooks into the event emitter of pg cursor instead of waiting for the callback.


matthieusieben:
Since you are not going to really use the rows buffered in `pg-cursor`, you should not keep them buffered to avoid un-necessary memory pressure.

PR #2583, opened 2021-07-27
Only call client.ref if it exists. Fixes #2582

### LINKED ISSUE #2582: pg-pool Exception
TypeError: client.ref is not a function
at BoundPool._pulseQueue (../node_modules/pg-pool/index.js:140:14)

I'm using pg@8.6.0 which, as of today, now installs pg-pool@3.4.0

brianc:


brianc:
fixes 2584

brianc:
fixes #2584

Issue #2588, opened 2021-07-29
Cursor - close behavior


CaptainHansen:
Is it safe to never call close() on a cursor unless I explicitly want to close the cursor prematurely?  As of now, we're calling close everytime the cursor runs to completion and when an error occurs.  However, I've noticed the following scenarios:

1. If close() is called on a cursor that has completed (state === 'done'), a promise is not returned.
2. if close() is called on a cursor that is in the error state, the promise/callback never runs (e.g. following a "Connection terminated unexpectedly" error)

Tested using most recent versions of pg and pg-cursor (8.7.1, 2.7.1 respectively), postgres v. 12.7, node v. 10.23.3

Description of 2:

One of our services uses cursors to gather potentially huge datasets for processing.  The database cluster sometimes terminates connections due to, for example, a conflict with recovery.  Our code attempts to close the cursor whether we reached the end or an error occurs.  If we've reached the end of the cursor, close() always resolves/completes the callback successfully.  However, when an error occurs, the callback/promise never completes causing our program to hang indefinitely.

Since the error that causes this scenario is always "Connection terminated unexpectedly", I simulated the error by looking up the PID of the connection to postgres and killing it on the server which yields the same effect.

Looking back at the documentation, it looks like close() should only be called when I want to `to close the cursor early`.  What about when an error occurs on the cursor?  Is it expected that the cursor is already closed in which case close() shouldn't be called at all?

Here is some JS that reproduces the issues (both 1 and 2):

```
const { Pool } = require('pg')
    , Cursor = require('pg-cursor')
    ;

const pool = new Pool({
...
});

console.log('wait until cursor completes to reproduce 1.');
const cursor = new Cursor('SELECT 1 FROM some_big_table');

return pool.connect()
.then(client => {
  client.query(cursor);

  client.on('error', e => {
    console.error(e);
  });

  console.log('Kill pid ' + client.processID + ' to cause a connection termination error and reproduce 2.');

  const loopFn = () => {
    return new Promise((resolve, reject) => {
      cursor.read(100, (err, rows) => {
        if (err) {
          reject(err);
        } else {
          resolve(rows);
        }
      });
    })
    .then(rows => {
      if (rows.length == 0) {
        return false;
      }

      console.log('Got ' + rows.length + ' rows.');
      return new Promise(resolve => {
        setTimeout(() => {
          resolve(true);
        }, 5000);
      });
    })
    .then(should_continue => {
      if (should_continue) {
        return loopFn();
      }
[...]

PR #2589, opened 2021-07-29
Return promise on cursor end

brianc:
Addresses one of the issues brought up in #2588


charmander (approved):
There’s also a redundant `if (cb)` a little farther down.

brianc:
> There’s also a redundant `if (cb)` a little farther down.

ty ❤️ ❤️

PR #2609, opened 2021-09-07
fix: Prevent closing the portal twice

matthieusieben:
Followup of brianc/node-postgres#2120
Fixes brianc/node-postgres#2119

### DIFF
--- a/packages/pg-cursor/index.js
+++ b/packages/pg-cursor/index.js
@@ -86,6 +86,8 @@ class Cursor extends EventEmitter {
   }
 
   _closePortal() {
+    if (this.state === 'done') return
+
     // because we opened a named portal to stream results
     // we need to close the same named portal.  Leaving a named portal
     // open can lock tables for modification if inside a transaction.
@@ -97,6 +99,8 @@ class Cursor extends EventEmitter {
     if (this.state !== 'error') {
       this.connection.sync()
     }
+
+    this.state = 'done'
   }
 
   handleRowDescription(msg) {
@@ -213,7 +217,6 @@ class Cursor extends EventEmitter {
     }
 
     this._closePortal()
-    this.state = 'done'
     this.connection.once('readyForQuery', function () {
       cb()
     })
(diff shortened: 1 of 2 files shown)


brianc:
Thanks @matthieusieben!  Is there any way you could write a test for this?

matthieusieben:
Hey @brianc I added a test that indeed fails on my setup (without the fix of course). Not sure this test is a 100% useful as I don't know if it will fail in every setup, ad I don't know how to trigger it in other setups...

brianc:
Hey @matthieusieben  - I was gonna merge this today.  CI was broken. I fixed it...would you be able to rebase your branch on to current `HEAD` of my `master` branch? I did so locally & ran the tests & a pg-query-stream error recovery test broke.

matthieusieben:
Hey @brianc , my wife just gave birth which is why I didn't answer earlier.

I rebased & fixed the issue. The problem was due to the ordering of the operation in the `_closePortal` method that was preventing `this.connection.sync()` from being executed.

Should be good now.

matthieusieben:
[CI still seem broken](https://github.com/matthieusieben/node-postgres/runs/4332627623?check_suite_focus=true) but I don't think its related to these changes:

```
connection-timeout-tests.js
  successful connection ✔
  expired connection timeout Message: listen EADDRINUSE: address already in use 127.0.0.1:54322
Error: listen EADDRINUSE: address already in use 127.0.0.1:54322
    at Server.setupListenHandle [as _listen2] (net.js:1280:14)
    at listenInCluster (net.js:1328:12)
    at GetAddrInfoReqWrap.doListen [as callback] (net.js:1461:7)
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:61:10)
```

matthieusieben:
RE-running CI worked:
https://github.com/matthieusieben/node-postgres/actions/runs/1507064878

There might be a race condition making the test suite fail ;-)

matthieusieben:
@brianc any change this gets merged any time soon ?

PR #2611, opened 2021-09-08
 feat: add conditions to errors, export POSTGRES_ERRORS_BY_CODE and P…

KristjanTammekivi:
Previously I've used a hack to override the error constructor but I feel like it's time to implement it directly in pg-protocol.

Features added:
* condition property on errors
* exported POSTGRES_ERROR_CODES and POSTGRES_ERRORS_BY_CODE (so it's easier to do stuff like `if (error.code === POSTGRES_ERROR_CODES.UNIQUE_VIOLATION)`)
* added a script to generate the aforementioned dictionaries
* added a tighter typing on postgres error code


matthieusieben:
You should make this optional, for example by using another package:
```ts
const { getPgErrorCondition } = require('pg-error-details')

const condition = getPgErrorCondition(pgErr)
```

KristjanTammekivi:
> You should make this optional, for example by using another package:
> 
> ```ts
> const { getPgErrorCondition } = require('pg-error-details')
> 
> const condition = getPgErrorCondition(pgErr)
> ```

Well, it's just a new property on the error so I'd say it's optional

matthieusieben:
The 20k addition to the package size are not really optional :/

PR #2614, opened 2021-09-13
Only call queryCallback if it's valid

adam-nielsen:
If `query()` is called with a single object parameter (such as the example code in the [node-pg-copy-streams README](https://github.com/brianc/node-pg-copy-streams)) then there is no callback function available.  However if the query times out, the timeout function will try to call the handler anyway.

This small fix ensures the user callback will only be called if one was supplied.  It fixes the error:

        at Timeout._onTimeout (node_modules/pg/lib/client.js:532:9)

### DIFF
--- a/packages/pg/lib/client.js
+++ b/packages/pg/lib/client.js
@@ -546,7 +546,7 @@ class Client extends EventEmitter {
 
       query.callback = (err, res) => {
         clearTimeout(readTimeoutTimer)
-        queryCallback(err, res)
+        if (queryCallback) queryCallback(err, res)
       }
     }
 


charmander:
This looks like a change in the success case; does it also apply to the one in `readTimeoutTimer`’s function?

charmander:
Never mind, see https://github.com/brianc/node-postgres/pull/2000#issuecomment-559608868.

adam-nielsen:
Sorry, this change is no longer required with brianc/node-pg-copy-streams#125, closing.

avivshafir:
why is this not merged and released?

adam-nielsen:
My fix appears to hide situations where the library is used incorrectly, so rather than apply this fix (hiding the problem) the calling code was changed to use the library the correct way.

You can still apply my fix as a temporary workaround, however as I understand it the ultimate fix would be to work out which part of the code is using the library incorrectly and fix that instead.  As you saw, I added a test in #2664 that reproduces the error so the maintainers can hopefully track down the bug and fix it at its originating point.

Issue #2280, opened 2020-07-16
pg-connection-string ConnectionOptions interface is not compatible with PoolConfig


hjr3:
There are a few issues:

1. pg-connection-string `ConnectionOptions` has `database: string | null | undefined` vs `PoolConfig` has  `database?: string;` (which effectively means `database: string | undefined`).
1. pg-connection-string `ConnectionOptions` has `ssl?: boolean | string` vs `PoolConfig` has  `ssl?: boolean | ConnectionOptions`
   - `string` is not present in `PoolConfig`
   - `PoolConfig` specifies  `tls.ConnectionOptions` and pg-connection-string `ConnectionOptions` does not. (Note: Do not get confused here. There is [pg-connection-string ConnectionOptions](https://github.com/brianc/node-postgres/blob/master/packages/pg-connection-string/index.d.ts#L3) and [node's tls ConnectionOptions](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/tls.d.ts#L442)

hjr3:
I originally thought about making a short term fix to include [tls.ConnectionOption](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/tls.d.ts#L442) as part of pg-connection-string. This may break some TypeScript code though if people were doing something like:

```ts
if (config.ssl === '1') {
   // some logic
} else {
   // implicitly thinking it is now a bool
}
```

My recommendation is:

- Replace `ConnectionOptions` with `PoolConfig`
- Update pg-connection-string to `3.0.0`

ThisIsMissEm:
@brianc I just encountered this in the Mastodon codebase, and the only solution I had for working around it was this:

```javascript
    const parsedUrl = dbUrlToConfig(env.DATABASE_URL);

    // The result of dbUrlToConfig from pg-connection-string is not type
    // compatible with pg.PoolConfig, since parts of the connection URL may be
    // `null` when pg.PoolConfig expects `undefined`, as such we have to
    // manually create the baseConfig object from the properties of the
    // parsedUrl.
    //
    // For more information see: https://github.com/brianc/node-postgres/issues/2280
    if (typeof parsedUrl.password === 'string') baseConfig.password = parsedUrl.password;
    if (typeof parsedUrl.host === 'string') baseConfig.host = parsedUrl.host;
    if (typeof parsedUrl.user === 'string') baseConfig.user = parsedUrl.user;
[...]

PR #2637, opened 2021-10-17
ConnectionOptions interface is incompatible with @types/pg definitions

jameshfisher:
Using pg-connection-string with @types/pg currently results in a type error like this:

    Argument of type 'ConnectionOptions' is not assignable to parameter of type 'PoolConfig'.
      Types of property 'database' are incompatible.
        Type 'string | null | undefined' is not assignable to type 'string | undefined'.
          Type 'null' is not assignable to type 'string | undefined'.ts(2345)

See clashing definition: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/08f0feedca9329fe95d23ccc64886e3db03195bf/types/pg/index.d.ts#L18

### DIFF
--- a/packages/pg-connection-string/index.d.ts
+++ b/packages/pg-connection-string/index.d.ts
@@ -5,7 +5,7 @@ export interface ConnectionOptions {
   password?: string
   user?: string
   port?: string | null
-  database: string | null | undefined
+  database: string | undefined
   client_encoding?: string
   ssl?: boolean | string
 


hjr3:
There are more problems with having `ConnectionOptions` match `ClientConfig`. See https://github.com/brianc/node-postgres/issues/2280

We landed https://github.com/brianc/node-postgres/pull/3128 to remedy this issue. Please let me know if you are still having problems.

PR #2654, opened 2021-11-16
Use github actions for CI

brianc:
Travis is giving me no end of headaches.  I need to get tests running again so we can safely merge & push things forward again.  Working on that here.


brianc:
@charmander I've got a bunch of PR backlog to churn through, and a security patch to get out as well.  I have the ssl tests working locally & I'd like to merge this & then we can work on getting SSL tests integrated into the test suite in a future PR since it's been way too long and travis CI just continually stops working for us.  Thoughts?

Issue #2652, opened 2021-11-10
Set timeout for 1 query when using pg-pool


Paxa:
Is it possible to set timeout for 1 query when using pg-pool?

As I see here it only check `this.connectionParameters.query_timeout`
https://github.com/brianc/node-postgres/blob/master/packages/pg/lib/client.js#L513

And if I use `pool.query(...)` I can't modify `connectionParameters` object only for 1 query

btw what is a meaning of `if (typeof config.submit === 'function')` (line 506)

PR #2655, opened 2021-11-17
Add config.query_timeout for query()

Paxa:
```js
client.query('select pg_sleep(10)', {query_timeout: 5})
  .then(...)
  .catch(error => { e.message == 'Query read timeout' })
```

Solve #2652

### DIFF
--- a/packages/pg/lib/client.js
+++ b/packages/pg/lib/client.js
@@ -510,7 +510,7 @@ class Client extends EventEmitter {
         query.callback = query.callback || values
       }
     } else {
-      readTimeout = this.connectionParameters.query_timeout
+      readTimeout = config.query_timeout || this.connectionParameters.query_timeout
       query = new Query(config, values, callback)
       if (!query.callback) {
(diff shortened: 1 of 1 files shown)


brianc:
thanks for the PR! Could you rebase on current `master` as I've re-enabled CI (travis ci kept breaking).  And then if you could please write a quick integration test for this so we can be sure it will continue to work in the future!

Paxa:
Hi, thanks for reply, I'm afraid I don't have time to bring this PR to the right shape

Feel free to continue where I left

Paxa:
Applied to lastest master https://github.com/Paxa/node-postgres/commit/03630e1a212518bfe0c751472cdd80f3fd7aa741

eladchen:
@brianc What's needed for this to be released?

brianc:
Apologies. Been crazy spinning down my current job...I'll try to get this released today or tomorrow. Going forward, after a few weeks of vacation, you can expect much more prompt responses here.

PR #2656, opened 2021-11-21
fix ECONNREFUSED on Node v17+

### LINKED ISSUE #2643: Node v17 ECONNREFUSED
## Description
On node v16.13.0, connecting to the database works as expected. I decided to try out the changes in v17.0.1 and I can't connect to the db anymore.

## Test script:
```js
import pg from 'pg';

export const pool = new pg.Pool({
    user: process.env.POSTGRES_USER,
    password: process.env.POSTGRES_PASS,
    database: 'kb'
});

console.log(process.version);
pool.query(`SELECT NOW()`).then(({ rows }) => console.log(rows[0]));
```

## Output (both versions)
```
pg>node test
v16.13.0
{ now: 2021-10-29T18:14:27.011Z }
pg>node test
v17.0.1
node:internal/process/promises:246
          triggerUncaughtException(err, true /* fromPromise */);
          ^

Error: connect ECONNREFUSED ::1:5432
    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1161:16) {
  errno: -4078,
  code: 'ECONNREFUSED',
  syscall: 'connect',
  address: '::1',
  port: 5432
}

Node.js v17.0.1
```

KhafraDev:
Fixes #2643 

I have no clue if this will break anything else, should `family` get added to config options? Anyways I tested on v14, v16, v17 (all latest) and it works fine.

### DIFF
--- a/packages/pg/lib/connection.js
+++ b/packages/pg/lib/connection.js
@@ -35,7 +35,11 @@ class Connection extends EventEmitter {
 
     this._connecting = true
     this.stream.setNoDelay(true)
-    this.stream.connect(port, host)
+    this.stream.connect({
+      port,
+      host,
+      family: 4
+    });
 
     this.stream.once('connect', function () {
       if (self._keepAlive) {
--- a/packages/pg/lib/connection.js
+++ b/packages/pg/lib/connection.js
@@ -4,10 +4,14 @@ var net = require('net')
 var EventEmitter = require('events').EventEmitter
 
 const { parse, serialize } = require('pg-protocol')
+const { setDefaultLocalhost } = require('./defaults')
 
 const flushBuffer = serialize.flush()
 const syncBuffer = serialize.sync()
 const endBuffer = serialize.end()
+
+/** @type {import('dns').LookupAddress} */
+let lookup;
 
 // TODO(bmc) support binary mode at some point
 class Connection extends EventEmitter {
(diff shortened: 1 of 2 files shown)


charmander (changes requested):
This looks like it’ll prevent an IPv6 connection, won’t it?

darkgl0w (changes requested):
> This looks like it’ll prevent an IPv6 connection, won’t it?

Yes this will force IPv4 connections.

For reference, this issue is related to dns changes introduced with `Node.js 17.0.0` (https://github.com/nodejs/node/pull/39987)

Maybe adding something like this to the `packages/pg/lib/defaults.js` file should fix the problem (works on Node.js 8+):

```javascript
'use strict'

const dns = require('dns')
const util = require('util')

async function setDefaultLocalhost () {
  const dnsLookup = util.promisify(dns.lookup)

  return await dnsLookup('localhost')
}

const host = setDefaultLocalhost()

module.exports = {
  // database host. defaults to localhost IPv4 ('127.0.0.1') or IPv6 ('::1') detected value
  host: host.address,
```

KhafraDev:
> This looks like it’ll prevent an IPv6 connection, won’t it?
Yes, just trying a solution in hopes of someone else having a better one now that the problem is known. :)

> Maybe adding something like this to the packages/pg/lib/defaults.js file should fix the problem (should work on Node.js 8+):
I've tried implementing this in a different way because the function is async and can't be resolved as it is in your example.

charmander:
Just set `host: '127.0.0.1'`?

treysis:
Rather, PostgreSQL should be set to `listen_addresses = '127.0.0.1,::1'` until it is fixed to listen on dual stack by default when no address is specified.

PR #2663, opened 2021-11-30
Add a Chinese cryptography algorithm, its name is SM3, currently used…

w345731923:
Add a Chinese cryptography algorithm, Its name is SM3, similar to MD5,currently used by many Chinese database companies. Later I will submit the use in /pg.

### DIFF
--- a/packages/pg/package.json
+++ b/packages/pg/package.json
@@ -24,9 +24,10 @@
     "packet-reader": "1.0.0",
     "pg-connection-string": "^2.5.0",
     "pg-pool": "^3.4.1",
-    "pg-protocol": "^1.5.0",
     "pg-types": "^2.1.0",
-    "pgpass": "1.x"
+    "pgpass": "1.x",
+    "postgres-date": "^2.0.0",
+    "sm-crypto": "^0.3.6"
   },
   "devDependencies": {
     "async": "0.9.0",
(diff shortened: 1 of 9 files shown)


charmander:
I don’t think support for this should be added.

darkgl0w (changes requested):
As @charmander I think that support for this should not be added.

adam-nielsen:
Apparently SM3 is an open standard and based on the NSA's SHA-2 algorithm, and offers similar security to SHA-256.  I've done some cursory investigation and it appears that there aren't any security issues around it, however it also doesn't appear to have been widely investigated.  It would be interesting to see if someone like the NSA has reviewed it and what their opinion of it is.  At any rate, this PR only proposes using the SM3 algorithm for authentication, which does not appear to be a cause for concern as it would only be used if you configured your PG server to also use SM3.

However it does concern me that the algorithm implementation is pulled in from an external library with quite a small user base, as this increases the risk of unauthorised code injection.  If SM3 support is to be added (which I have no problem with), I think it would be more secure to include the algorithm implementation as part of the PR itself, to avoid pulling in another external library that is not under the control of the existing node-postgres maintainers.

packages/pg/package.json
```diff
@@ -24,9 +24,10 @@
     "packet-reader": "1.0.0",
     "pg-connection-string": "^2.5.0",
     "pg-pool": "^3.4.1",
-    "pg-protocol": "^1.5.0",
```
adam-nielsen, line 27: How come `pg-protocol` is being removed?  Don't you need this for the new algorithm that has been added?

w345731923:
I agree with you,SM3 Code 13 May not be universal, at least for now.

brianc:
> However it does concern me that the algorithm implementation is pulled in from an external library with quite a small user base, as this increases the risk of unauthorised code injection. If SM3 support is to be added (which I have no problem with), I think it would be more secure to include the algorithm implementation as part of the PR itself, to avoid pulling in another external library that is not under the control of the existing node-postgres maintainers.

100% agree with this. Thank you!

PR #2560, opened 2021-06-04
Added default value to prevent error queryCallback is not a function

drzippie:
fix issue #1860 

Set default value to prevent Fatal exception with timeouts in a queryStream

PR #2664, opened 2021-12-01
[pg-cursor] Add query_timeout parameter to test code to trigger queryCallback bug

adam-nielsen:
There is a bug in the query callback handling (e.g. #1860 #2560) but the test code doesn't pick it up.  This change ensures that code path runs in the tests, so it will probably make the tests fail until the bug is fixed.  (I say 'probably' because the exception doesn't get thrown until after all the tests have completed successfully.)

```
/srv/its-pf1pr0gd/code/node-postgres/packages/pg-cursor/node_modules/mocha/lib/runner.js:911
    throw err;                                                               
    ^                                                                        
                                                                             
    at Timeout._onTimeout (/srv/its-pf1pr0gd/code/node-postgres/packages/pg-cursor/node_modules/pg/lib/client.js:532:9)
    at listOnTimeout (node:internal/timers:557:17)                           
    at processTimers (node:internal/timers:500:7)
```

### DIFF
--- a/packages/pg-cursor/test/index.js
+++ b/packages/pg-cursor/test/index.js
@@ -6,7 +6,9 @@ const text = 'SELECT generate_series as num FROM generate_series(0, 5)'
 
 describe('cursor', function () {
   beforeEach(function (done) {
-    const client = (this.client = new pg.Client())
+    const client = (this.client = new pg.Client({
+      query_timeout: 15000,
+    }))
     client.connect(done)
 
     this.pgCursor = function (text, values) {


brianc:
The weird thing is github actions isn't running on this PR. :( I might not have added the right thing to it to run on a pull or something...I'll have to look into that.

brianc:
Thanks for this btw! And sorry for the delay: was out of town until last night.

darkgl0w:
@brianc you must add `on: [push, pull_request]` to the `ci.yml` config file ^^

brianc:
> @brianc you must add `on: [push, pull_request]` to the `ci.yml` config file to run on a PR ^^

roger, wilco!

avivshafir:
@brianc is there any solution to this bug?

yaverin:
@brianc, Please provide ETA for this fix - to prevent Fatal exception using "query_timeout" with queryStream.

Issue #2027, opened 2019-12-13
Add optional maximum connection lifetime parameter


hempels:
One of the issues we've run into with using Aurora Postgresql with node-postgres is that while the database is capable of auto-scaling read cluster instances, the existing connections do not react to those changes.

For example, if I have a pool of 60 connections against a single read cluster DB instance and then add a second read-cluster DB instance, the 60 pooled connections remain attached to the original instance. The second instance sits idle while the first is pegged.

My thinking is the simplest solution may be to apply an automatic connection reset strategy whereby no pool connection would live longer than X minutes. In a busy environment where the pool is maxed, you wouldn't want to close and re-open multiple connections at once, they would need to be handled one at a time. But it seems to me the performance impact of infrequently resetting connections would be low, and the advantage of benefiting from an auto-scaling DB would easily offset it.

julienvey:
To add more insight to this issue, this configuration options exists in most of PG libraries for other langages

See https://golang.org/pkg/database/sql/#DB.SetConnMaxLifetime or https://www.npgsql.org/doc/connection-string-parameters.html#pooling (ConnectionLifetime)

We have the same issue as @hempels. We use auto-scaled read-only DBs on a kubernetes cluster and the connections to the different DBs are not evenly distributed

jodem:
Hi all, I think this has been solved when implementing the maxUses options. It's not based on time but on how many requests you did with the connection : https://github.com/brianc/node-postgres/pull/2157 .

hempels:
maxUses is one lever that can be dialed, but is inadequate by itself. Set too low and system performance will be severely degraded under load. Set too high and it will not trigger often enough under lighter loads to be of any use.

jolynch:
This would also be  very useful for any case where the client is connecting through proxies for failover/scaling (haproxy, envoy, service-meshes etc ...) where you need the maximum connection lifetime to be under some deadline so you can reload connections gracefully. We've been working around this by catching and handling the close errors but that's not particularly great.

hempels:
> To add more insight to this issue, this configuration options exists in most of PG libraries for other langages

The .net library's docs even say this in the description for its ConnectionLifetime parameter:
> This is useful in clustered configurations to force load balancing between a running server and a server just brought online.

ChrisWritable:
The rebased update is here: https://github.com/brianc/node-postgres/pull/2698

jolynch:
[...]

PR #2668, opened 2021-12-14
Add connection lifetime limit option and tests

ChrisWritable:
This addresses issue #2027 "Add optional maximum connection lifetime parameter"
https://github.com/brianc/node-postgres/issues/2027

Adds the optional `maxLifetimeSeconds` parameter to the `Pool` constructor. If the parameter is left undefined or zero there is no change in behavior. When the parameter is set, it specifies a number of seconds to wait before recycling a connection. It waits until the connection is idle before recycling.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -4793,9 +4793,9 @@ path-key@^3.1.0:
   integrity sha512-ojmeN0qd+y0jszEtoY48r0Peq5dwMEkIlCOu6Q5f41lfkswXuKtYrhgoTpLnyIcHm24Uhqx+5Tqm2InSwLhE6Q==
 
 path-parse@^1.0.6:
-  version "1.0.6"
-  resolved "https://registry.yarnpkg.com/path-parse/-/path-parse-1.0.6.tgz#d62dbb5679405d72c4737ec58600e9ddcf06d24c"
-  integrity sha512-GSmOT2EbHrINBf9SR7CDELwlJ8AENk3Qn7OikK4nFYAu3Ote2+JYNVvkpAEQm3/TLNEJFD/xZJjzyxg3KBWOzw==
+  version "1.0.7"
+  resolved "https://registry.yarnpkg.com/path-parse/-/path-parse-1.0.7.tgz#fbc114b60ca42b30d9daf5858e4bd68bbedb6735"
+  integrity sha512-LDJzPVEEEPR+y48z93A0Ed0yXb8pAByGWo/k5YYdYgpY2/2EsOsksJrq7lOHxryrVOn1ejG6oAp8ahvOIQD8sw==
 
 path-type@^1.0.0:
   version "1.1.0"


charmander (changes requested):
Juggling indices (which change when you `splice`) feels uncomfortable. How about adding clients that are checked out when they expire to a `WeakSet`, then checking for previous expiration that way in `release`?

Also, this PR changes the mode of packages/pg-pool/index.js; could you revert that? (Or I can.)

ChrisWritable:
I believe I've made the changes requested. Let me know if this newer pull request 3ea8ff26aa isn't what you were asking for.

packages/pg-pool/index.js
```diff
@@ -414,6 +439,10 @@ class Pool extends EventEmitter {
     return this._idle.length
   }
 
+  get expiredCount() {
```
charmander, line 442: Is this mainly for the test, or do you see it being useful as a public API?

packages/pg-pool/test/lifetime-timeout.js
```diff
@@ -0,0 +1,48 @@
...
+const co = require('co')
+const expect = require('expect.js')
+
+const describe = require('mocha').describe
+const it = require('mocha').it
+const path = require('path')
+
+const Pool = require('../')
+
+const wait = (time) => new Promise((resolve) => setTimeout(resolve, time))
```
charmander, line 11: Unused?

packages/pg-pool/index.js
```diff
@@ -248,6 +251,19 @@ class Pool extends EventEmitter {
       } else {
         this.log('new client connected')
 
+        if (0 !== this.options.maxLifetimeSeconds) {
+          setTimeout(() => {
+            this.log('ending client due to expired lifetime')
+            this._expired.add(client)
+            if (this._idle.length) {
```
charmander, line 258: Seems like the `-1` check already covers this.

packages/pg-pool/index.js
```diff
@@ -248,6 +251,19 @@ class Pool extends EventEmitter {
       } else {
         this.log('new client connected')
 
+        if (0 !== this.options.maxLifetimeSeconds) {
```
charmander, line 254: minor style consistency

```suggestion
        if (this.options.maxLifetimeSeconds !== 0) {
```

packages/pg-pool/index.js
```diff
@@ -248,6 +251,19 @@ class Pool extends EventEmitter {
...
+          setTimeout(() => {
+            this.log('ending client due to expired lifetime')
+            this._expired.add(client)
+            if (this._idle.length) {
+              const idleIndex = this._idle.findIndex(idleItem => idleItem.client === client)
+              if (-1 !== idleIndex) {
```
charmander, line 260:
```suggestion
              if (idleIndex !== -1) {
```

charmander (changes requested):
Yep, exactly!

packages/pg-pool/index.js
ChrisWritable: It's useful as a public API to poll the count of busy connections pending expiration. For example, if you have some long running queries that are preventing expiration you could see that occurring. It is also useful for testing.

packages/pg-pool/test/lifetime-timeout.js
ChrisWritable: Yes. I can remove that.

packages/pg-pool/index.js
ChrisWritable: Yes. I was just thinking it was a quicker initial check. I'll remove it.

ChrisWritable:
I've updated the patch. Let me know what you think.

hempels:
@brianc Any thoughts on this PR? Would love to see it incorporated into node-postgres sooner than later if it passes muster.

brianc:
Yeah it looks good, sorry for missing this.  I'll try to get this released today!

brianc (approved):

brianc:
@ChrisWritable - would it be possible to rebase this PR on top of current master of this repo?  Because the tests haven't been run in CI on this PR AFAICT - I ported them to run on github actions so if you rebase on master & push back to this PR it should run them. Then I will merge. Thanks for your patience!

brianc:
Thanks for the PR @ChrisWritable - just FYI in the future if you don't want to open a new PR I'm pretty sure force pushing to your existing branch will just update the existing PR w/ the new commits or history.  Either way works though!

ChrisWritable:
Ok. Thanks.

PR #2682, opened 2022-01-07
add support for user lookup function

thijs:
Similar to the dynamic lookup of a password, this PR makes it possible to support a lookup of the user as well. This is needed for AWS RDS rotation where the username is (can, in some rotation schemes) rotated as well.

### DIFF
--- a/packages/pg/lib/client.js
+++ b/packages/pg/lib/client.js
@@ -114,16 +114,28 @@ class Client extends EventEmitter {
     }
 
     // once connection is established send startup message
-    con.on('connect', function () {
+    con.on('connect', async function () {
       if (self.ssl) {
         con.requestSsl()
       } else {
-        con.startup(self.getStartupConf())
+        try {
+          const conf = await self.getStartupConf()
+          con.startup(conf)
+        }
+        catch(err) {
+          con.stream.destroy(err)
+        }
       }
     })
 
-    con.on('sslconnect', function () {
-      con.startup(self.getStartupConf())
+    con.on('sslconnect', async function () {
+      try {
+        const conf = await self.getStartupConf()
+        con.startup(conf)
+      }
+      catch(err) {
+        con.stream.destroy(err)
+      }
     })
 
     this._attachListeners(con)
@@ -385,11 +397,26 @@ class Client extends EventEmitter {
     this.emit('notice', msg)
   }
 
-  getStartupConf() {
+  async getStartupConf() {
     var params = this.connectionParameters
 
+    var user = params.user
+    if (typeof this.user === 'function') {
+      user = await this.user()
+
+      if (user !== undefined) {
+        if (typeof user !== 'string') {
+          throw new TypeError('User must be a string')
+          return
+        }
+        this.connectionParameters.user = this.user = user
+      } else {
+        this.connectionParameters.user = this.user = null
+      }
+    }
+
     var data = {
-      user: params.user,
+      user: user,
       database: params.database,
     }
 
--- a/packages/pg/test/integration/connection/dynamic-password-tests.js
+++ b/packages/pg/test/integration/connection/dynamic-password-tests.js
@@ -6,9 +6,10 @@ const pg = require('../../../lib/index')
 const Client = pg.Client
 
 const password = process.env.PGPASSWORD || null
+const user = process.env.PGUSER || null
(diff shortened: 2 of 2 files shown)


ssteuteville:
I'd use this.

thijs:
Is there something else I need to do to get this merged?

brianc:
Gosh this PR is old - sorry for letting it go by w/o any comments.  It looks alright to me, and appreciate the PR!  I think in the future it'd probably be good to just make a single generic thing like `new Client({ getConfig: async () => {} })` or something so everything can be dynamically controlled by the user...then we could get rid of explicit `pg-pass` support and so on. But for now this does the job!

brianc:
hmmm....looks like a test is failing.

thijs:
No worries, thanks for taking a look now. It's been so long, I'm not entirely surprised tests fail. I'll take a look at fixing them...

thijs:
If I run the tests locally on `master` I get an error (so without any of my changes) in `dynamic-password-tests.js`:
```
dynamic-password-tests.js
  Get password from a sync function FAILED!

AssertionError [ERR_ASSERTION]: Our password function should have been called
    at /home/thijs/code/node-postgres/packages/pg/test/integration/connection/dynamic-password-tests.js:26:12
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

So the existing functionality to get a dynamic password fails the test on master. For me. Am I doing something wrong?

brianc:
> So the existing functionality to get a dynamic password fails the test on master. For me. Am I doing something wrong?

hmmm lemme take a look

brianc:
I just ran all tests locally on commit `adbe86d4a057b942298cab1d19b341c67a94d922` under `packages/pg` by running `make test` with no issue....hmmm possible something w/ the way your local postgres is configured? Do you have `PGHOST` and `PGUSER` etc env vars set?  The pg tests are a bit brittle as some of them are __very__ old.

thijs:
Yes, `make test` runs okay for me, but that one does not run the integration tests, which is what is failing. Also on that commit (`adbe86d4a057b942298cab1d19b341c67a94d922`). If you run `make test-all` you will get the error I posted above.

And seeing as I added my test basically copying what was already in that file, I'd need for that test to pass before tackling my own stuff...

thijs:
Interestingly, in the CI logs I actually see this:
```
***Testing connection***
Created table person
Filling it with people
Inserted 26 people
***Testing Pure Javascript***
xargs: warning: options --max-args and --replace/-I/-i are mutually exclusive, ignoring previous --max-args value
bound-command-tests.js..
copy-tests.js..
notification-tests.js.
query-tests.js.
dynamic-password-tests.js
  Get password from a sync function ✔
  Throw error from a sync function ✔
  Get password from a function asynchronously ✔
  Throw error from an async function ✔
  Password function must return a string ✔
  Get user from a sync function ✔
  Throw error from user sync function ✔
  Get user from a function asynchronously ✔
  Throw error from async user function ✔
  User function must return a string ✔

[...]
```

And this last part in that partial paste:
```
  Get user from a sync function ✔
  Throw error from user sync function ✔
  Get user from a function asynchronously ✔
  Throw error from async user function ✔
  User function must return a string ✔
```
are the tests that I added. So they are not failing in the CI pipeline. The tests are failing elsewhere, namely here:
```
Unhandled promise rejection
quick-disconnect-tests.jsMessage: write after end
Error [ERR_STREAM_WRITE_AFTER_END]: write after end
    at new NodeError (internal/errors.js:322:7)
    at Socket.Writable.write (internal/streams/writable.js:292:11)
    at Connection.startup (/home/runner/work/node-postgres/node-postgres/packages/pg/lib/connection.js:128:17)
    at Connection.<anonymous> (/home/runner/work/node-postgres/node-postgres/packages/pg/lib/client.js:122:15)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
```
And I have no clue why, or if that is related to the changes I made. On first glance I would say no, but I find it all pretty weird that my tests are failing for me locally and not in CI...

thijs:
Sorry about the multiple posts, but if I run `make test-all` on my branch, then it errors out on the same error as in CI, only the order of the tests is different, so locally the test run never gets to my tests, so I don't see the results of that locally. Maybe they are ok here too.

charmander (changes requested):
needs a correct rebase

PR #2687, opened 2022-01-13
Support types/getTypeParser in query method for pg-native

### LINKED ISSUE #2686: types/getTypeParser in query method doesn't work with pg-native?
Hello, I have simple code like this:

```javascript
var pg = require("pg");

const client = new pg.Client();
// ...
client.query({
        text: "SELECT num FROM data WHERE id = $1",
        values: [1],
        rowMode: "array",
        types: {
            getTypeParser: () => val => val,
        }
    },
    (err, result) {
        console.log(result.rows);
    });
```

This works great and I get the expected output: `[ "567" ]`

But when I change the first "require" line to use "pg-native", like this:

```javascript
var pg = require("pg").native;
```

then my custom parser seems to be ignored, and I get the following output: `[ 567 ]`

I am using:
  - pg 8.7.1
  - pg-native 3.0.0

Thank you

benny-medflyt:
This fixes #2686

This is a simple fix that gets the job done. I've done the same thing that is done with `_arrayMode`: update the value on the `client.native` and then change it back when we are done.

### DIFF
--- a/packages/pg/lib/native/query.js
+++ b/packages/pg/lib/native/query.js
@@ -13,6 +13,7 @@ var NativeQuery = (module.exports = function (config, values, callback) {
   this.callback = config.callback
   this.state = 'new'
   this._arrayMode = config.rowMode === 'array'
+  this.types = config.types;
 
   // if the 'row' event is listened for
   // then emit them as they come in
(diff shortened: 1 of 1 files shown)


brianc:
Hey @benny-medflyt - thanks for this! Would it be possible to write a test to cover this?  Also, if you rebase on master the existing tests should run on this PR which makes it easier for me to merge!

Issue #2646, opened 2021-10-31
implement pipelining mode allowed by libpq 14


abenhamdine:
libpq shipped with Postgres 14 now allows using pipelining mode 
https://www.postgresql.org/docs/14/libpq-pipeline-mode.html

as stated by these docs
> While the pipeline API was introduced in PostgreSQL 14, it is a client-side feature which doesn't require special server support and works on any server that supports the v3 extended query protocol.

So I don't know exactly when v3 extended query protocol has appeared (at least 2010 I will guess), but  it's there since a lot of time, so probably all supported version of pg server (9.6 and above) are able to support pipelining mode.

For the record, previous discussion on pipelining mode (before libpq supports it) is here :  https://github.com/brianc/node-postgres/issues/660

jadbox:
Any update on this?

abenhamdine:
> Any update on this?

I did some tests locally with a fork, built upon https://github.com/brianc/node-postgres/pull/662 and noticed some interesting perf gains in my uses cases (between 5% and 20%).
I will send a PR as a POC.
Hower it adress only the js driver, not the native one.

marcbachmann:
Explicit pipelining would be interesting. E.g. `ioredis` supports `client.pipeline([['get', 'foo'], ['get', 'bar']]).exec()`.
`pg` could do the same to ensure they are sent on the same connection without any other slow query interfering.
e.g. `const [[errA, resA], [errB, resB]] = await pg.pipeline([{text, values}, {text, values}])`.

abenhamdine:
> Explicit pipelining would be interesting.

I don't get your point. If pipelining is possible, why would you want to get it explicitely and would not want not always use it ?   
   
I can understand to put it behind a flag (eg `--pipelining` at connection level) until we are sure  it's safe to use it for every use case, but I fail to see why it's should be explicit in the api.

marcbachmann:
With transactions it will get executed to a single client, so that will work as desired.
 
But when using a pool, pipeline would get distributed among all the clients. In some cases you might want to execute it on the same client instance so other queries won't be scheduled in between. But maybe that's not worth it to implement.

abenhamdine:
> But when using a pool, pipeline would get distributed among all the clients

Indeed, but it will be the case with or without pipelining, thus why would it be an issue ? Or perhaps I miss something in your explanation ?

elchinoo:
[...]

PR #2706, opened 2022-02-16
[WIP][POC] Use pipelining mode

abenhamdine:
Work in progress to adress #2646  
Base on the previous work of @jusou here https://github.com/brianc/node-postgres/pull/662, thx to him !

### Principle

In current behaviour of node-postgres, a query is not sent to the database until the previous query has completed and the associated result received.
With this PR, pipelining mode can be enabled, it allows the client to send all of the queries to the server up front, minimizing time spent by one side waiting for the other to finish sending data.

Important : pipelining has nothing to do with query execution concurrency/parallelism. With or without pipelining mode, the PostgreSQL server is executing the queries sequentially (while using parallelism capabilities if enabled), pipelining just allows both sides of the connection to work concurrently when possible, and to minimize round-trip time.  

The higher the latency is between the driver and the posgres serveur and the number of requests sent, the higher the performance gain will be.

### API changes 
- new property `client.pipelining` can be set to true to enable pipelining mode (default false)
```js
const client = await pg.connect()
client.pipelining = true
```

### Implementation  
 
- [ ] add tests
   - [ ] add unit tests 
       - [x] queries ready to emit
       - [ ] other ?
   - [ ]  add integration tests
       - [ ] TBD     
- [ ] currently it adress only js driver, not native one (I know nothing about the native driver, so I doubt I will be able to adress that)
- [ ] prepared statements are broken (probably because of race conditions in parsing), need to be fixed  
- [x] the default behaviour has to be defined (probably optout until we are sure it's safe) **false by default in this PR**
- [x] allow to optin (or optout, depending on the default) at the pool/client level **client.pipelining = true** 

While it's not finished, we are currently testing it with our product, with massive usage of queries (50 tasks in parallel sending each thousands of queries) and it seems to work (except of prepared statements).

I'm of course open to suggestions/review but I'm very busy nowadays thus don't be frustrated if I don't answer to remarks in a decent delay.

#### More details concerning the issue with the prepared statements broken in the PR : 

If we send immediatly 2 queries with the same named Foo,  we got the error `Prepared statement Foo already exists` 

Looks like the driver try to prepare the statement twice, probably because the event `parseComplete` has not been received yet for the 1st query, when the 2nd query is sent.

I'm a bit confused about how to find the proper place to fix it.
First I thought that it would be simple to debounce the parsing in https://github.com/brianc/node-postgres/blob/21ccd4f1b6e66774bbf24aecfccdbfe7c9b49238/packages/pg/lib/query.js#L36
by checking if connection.preparedStatements[this.name] exists  
but then I realized this check is already performed here :  
 https://github.com/brianc/node-postgres/blob/21ccd4f1b6e66774bbf24aecfccdbfe7c9b49238/packages/pg/lib/query.js#L192
so I don't truly understand why this error occurs 🤔   

If someone can help/give some hints, it would be useful !

#### Perf gain in our use case (⚠️  YMMV)

In our real application (payroll processing) :  
165 tasks, concurrency  between 1 and 40 :   
distributed on 4 clients  
each task is sending thousands of DML queries, mostly SELECT  

gain : 1 second (no gain expected because of no concurrency)
concurrency 1 w/o pipeline : 2m28s
concurrency 1 with pipeline : 2m27s

gain : 18 seconds
concurrency 5 w/o pipeline : 2m20s
concurrency 5 with pipeline : 2m02s

 gain : 25 seconds (test n°2 : 26 seconds)
concurrency 20 w/o pipeline : 2m22s (test 2 : 2m24s)
concurrency 20 with pipeline : 1m57s (test 2 : 1m58s)

gain : 27 seconds (test n°2 : 28 seconds)
concurrency 40 w/o pipeline : 2m20s (test 2 2m27s)
concurrency 40 with pipeline : 1m53s (test 2 1m59s

#### Conclusion : 
Very interesting gain of 20% with concurrency 40 (our default in production) 
It's also important to note that we have disabled prepared statements with pipelining mode due to the bug, so 
the gain is perhaps even greater (by <5% more I would say).

### DIFF
--- a/packages/pg/lib/client.js
+++ b/packages/pg/lib/client.js
@@ -498,6 +498,10 @@ class Client extends EventEmitter {
       }
       this.blocked = query.blocking
       this.sentQueryQueue.push(query)
+      if (query.name) {
+        console.log(`we store that ${query.name} has been submitted`)
+        this.connection.submittedNamedStatements[query.name] = query.text
+      }
     }
 
     if (this.readyForQuery === true) {
--- a/packages/pg/lib/client.js
+++ b/packages/pg/lib/client.js
@@ -56,8 +56,8 @@ class Client extends EventEmitter {
 
     // Client.sentQueryQueue is the queue of queries that have been sent on the wire
     this.sentQueryQueue = []
-    // Client.sendImmediately can be set to false to restore the previous behavior
-    this.sendImmediately = true
+    // Client.pipelining can be set to true to enable experimental pipelining mode
+    this.pipelining = false
 
     this.binary = c.binary || defaults.binary
     this.processID = null
@@ -484,7 +484,7 @@ class Client extends EventEmitter {
       return
     }
 
-    while ((this.sendImmediately && !this.blocked) || (this.activeQuery === null && this.sentQueryQueue.length === 0)) {
+    while ((this.pipelining && !this.blocked) || (this.activeQuery === null && this.sentQueryQueue.length === 0)) {
       var query = this.queryQueue.shift()
       if (!query) break
 
--- a/packages/pg/lib/client.js
+++ b/packages/pg/lib/client.js
@@ -53,6 +53,12 @@ class Client extends EventEmitter {
         encoding: this.connectionParameters.client_encoding || 'utf8',
       })
     this.queryQueue = []
+
+    // Client.sentQueryQueue is the queue of queries that have been sent on the wire
+    this.sentQueryQueue = []
+    // Client.sendImmediately can be set to false to restore the previous behavior
+    this.sendImmediately = true
+
     this.binary = c.binary || defaults.binary
     this.processID = null
     this.secretKey = null
@@ -286,6 +292,7 @@ class Client extends EventEmitter {
     const { activeQuery } = this
     this.activeQuery = null
     this.readyForQuery = true
+    this.handshakeDone = true
     if (activeQuery) {
       activeQuery.handleReadyForQuery(this.connection)
     }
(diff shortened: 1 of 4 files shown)


abenhamdine:
Concerning prepared statements : 
with a async.map on the same named query Foo,  we got the error `Prepared statement Foo already exists` 

Looks like the driver try to create the prepared statement twice, probably because the 2nd query has been parsed while the 1st query has not completed yet.

I'm a bit confused about how to find the proper place to fix it.
First I thought that in https://github.com/brianc/node-postgres/blob/21ccd4f1b6e66774bbf24aecfccdbfe7c9b49238/packages/pg/lib/query.js#L36
I need to check if connection[this.name] exists  
but then I realized this check is already performed there https://github.com/brianc/node-postgres/blob/21ccd4f1b6e66774bbf24aecfccdbfe7c9b49238/packages/pg/lib/query.js#L192
so I don't truly understand why this error occurs.

abenhamdine:
> This is really rad! If you can write a failing test for the race condition and put it on this branch I might be able to help with looking into it.

Strangely I didn't succeed in writing an integration test to reproduce the bug.

Though I managed to avoid the bug with a crude hack by tracking the prepared statements submitted to server when queries are submitted here : 

https://github.com/brianc/node-postgres/blob/a33084f10028a9b4725bd796afd9b06adfe7cc9a/packages/pg/lib/client.js#L370

instead to track it in [handleParseComplete_handleParseComplete](https://github.com/brianc/node-postgres/blob/a33084f10028a9b4725bd796afd9b06adfe7cc9a/packages/pg/lib/client.js#L370)

and I check if prepared statement has been already processed here : 

https://github.com/brianc/node-postgres/blob/a33084f10028a9b4725bd796afd9b06adfe7cc9a/packages/pg/lib/client.js#L370

I could have used this.connection.preparedStatements to do it, but I prefered to use a distinct property to avoid potential side effects.

Issue #2281, opened 2020-07-16
no way to set rejectUnauthorized from pg-connection-string


hjr3:
Now that `rejectUnauthorized` is being defaulted to true, there is no way to modify that from the connection string. Some libraries ([example](https://github.com/gajus/slonik/blob/ee43f02378dc8272d7d2702ccd105f7a3abd04f2/src/factories/createPoolConfiguration.js#L31) are setting opposite defaults to try and ease the migration.

hjr3:
My recommendation:

In general, we should adhere to the [libpq connection URI documentation](https://www.postgresql.org/docs/12/libpq-connect.html#LIBPQ-CONNSTRING). Specifically, we should support the [Parameter Key Words](https://www.postgresql.org/docs/12/libpq-connect.html#LIBPQ-PARAMKEYWORDS).

- introduce `sslmode`. For non-native mode, I do not think we can easily support `allow` or `prefer`. I also think `verify-ca` will not work as expected as `rejectUnauthorized` seems to check the host. We can document the mapping for non-native mode like this:

sslmode | Will connection via SSL/TLS? | rejectUnauthorized
------------ | ------------- | -------------
disable | no | n/a
allow | no | n/a
prefer | no | n/a
require | yes | false
verify-ca | yes | false
verify-full | yes | true

- deprecate `ssl`. If both `ssl` and `sslmode` are specified, `sslmode` will win
- if `ssl` is set, also set `rejectUnauthorized: true`. This maintains the node default but makes it more obvious as to what is going on

hjr3:
If https://github.com/brianc/node-postgres/issues/2263 lands, then maybe `verify-ca` will _just work_ :tm:

lroal:
I have the same problem. Any solution on this ?

radcapitalist:
+1 for supporting **sslmode** keywords.  I expected:
`ssl: {
   require: false,
}`
to be equivalent to the "allow" sslmode keyword, so that my connection would use SSL if the server allowed or required it but not if it didn't.  I don't see a way to get the equivalent of "allow" with node-postgres.

the21st:
We are struggling with this as well. Ideally we want to be able to set `sslmode: prefer` for the ssl option.

bdwyertech:
You can pass `sslmode=no-verify` in the connection string or you can set env var `PGSSLMODE=no-verify`

ollytheninja:
[...]

Issue #2375, opened 2020-10-11
sslmode=require in connection string breaks with self signed certificate


zakjan:
After #2345 was merged, `sslmode=require` in connection string breaks with self signed certificate. SSL config is overriden to `{}`, leading to error "self signed certificate in certificate chain".

According to the table in #2281, correct SSL config is `{rejectUnauthorized: false}` in this case.

katlimruiz:
Hi, just wanted to make sure this was a suggestion how to fix it, but I guess is still not fixed?

ceefour:
this buggy `sslmode=require` behavior still exists in v8.8.0. If the user wants to verify certificate, it can use `verify-ca` or `verify-full`.

Currently the buggy behavior treats `require` similar as `verify-ca`.

PR #2709, opened 2022-02-21
fix(pg-connection-string): get closer to libpq semantics for `sslmode`

pmalouin:
## Summary

We have found that the handling of the `sslmode` connection string parameter is inconsistent with other PG libraries and with the reference [libpq documentation](https://www.postgresql.org/docs/12/libpq-connect.html#LIBPQ-CONNSTRING). This PR proposes some changes to `sslmode` behavior that are more aligned with libpq.

## Detailed `sslmode` behavior

Here is the list of all `sslmode` values and their expected behavior with this PR:

#### `sslmode=verify-full`
Require an SSL connection and verify the CA and server identity.
No changes in this PR.

#### `sslmode=verify-ca` *(changed)*
Require an SSL connection and verify the CA, but not the server identity. This is achieved by setting `ssl.checkServerIdentity` to a no-op function (see [docs](https://nodejs.org/api/tls.html#tlscheckserveridentityhostname-cert)). Previously, this mode behaved like `verify-full` but that was not consistent with the libpq implementation. SECURITY WARNING: Using `sslmode=verify-ca` requires specifying a CA with `sslrootcert`. If a public CA is used, `verify-ca` allows connections to a server that somebody else may have registered with the CA, making you vulnerable to Man-in-the-Middle attacks.

#### `sslmode=require` *(changed)*
If a root CA is provided via the `sslrootcert` parameter of the connection string, it behaves like `verify-ca`. Otherwise, require an SSL connection but do not verify CA and server identity (`ssl.rejectUnauthorized` is set to `false`).
Previously, this mode behaved like `verify-full` but that was not consistent with the libpq implementation.

#### `sslmode=no-verify`
Require an SSL connection but do not verify CA and server identity (`ssl.rejectUnauthorized` is set to `false`).
No changes in this PR. Note: this mode is not documented in libpq and does not appear to be broadly supported in other libraries, but has been kept as-is for the sake of backwards-compatibility. One option could be to mark it as deprecated since `sslmode=require` could be an alternative, but doing so might have little value for this project.

#### `sslmode=prefer` *(changed)*
Require an SSL connection but do not verify CA and server identity (`ssl.rejectUnauthorized` is set to `false`). Previously, this mode behaved like `verify-full` but that was not consistent with the libpq implementation.
In reality, this mode should be even less strict and support a fallback logic from SSL to non-SSL connection if SSL is not accepted by the server. Implementing a fallback logic seems to be more complex to solve and I did not dare touch this, but this could eventually be addressed if users of this library deem this mode valuable.

#### `sslmode=allow`
Not supported by this library.
No changes in this PR. For this mode also, there could be an opportunity to implement a fallback logic from non-SSL to SSL, but I did not dare touch this and I don't have data that suggests that this might be valuable for this project.

#### `sslmode=disable`
Only try a non-SSL connection.
No changes in this PR

An important note is that this PR potentially introduces **semver breaking changes**, in particular because it relaxes the security constraints of some `sslmode` values:

- `sslmode=prefer` is less strict, users should switch to `sslmode=verify-full` to keep parity.
- `sslmode=require` is less strict, users should switch to `sslmode=verify-full` to keep parity.
- `sslmode=verify-ca` is less strict, users should switch to `sslmode=verify-full` to keep parity.

## Prior discussions about `sslmode`

I believe this PR addresses concerns raised in these two GH issues in the past:
https://github.com/brianc/node-postgres/issues/2281
https://github.com/brianc/node-postgres/issues/2009

In particular, there has been [one case](https://github.com/brianc/node-postgres/issues/2281#issuecomment-1015824739) where the `sslmode=verify-ca` is currently too strict when connecting through AWS RDS Proxy, but the work-around of using `sslmode=no-verify` would disable CA verification completely.

## Other languages/libraries and their support for `sslmode`

Just as a reference, these two libraries are also trying to be more or less consistent with libpq:
- [Golang - pq](https://pkg.go.dev/github.com/lib/pq#hdr-Connection_String_Parameters)
  - Supports `disable`, `require`, `verify-ca` and`verify-full` only.
- [PostgreSQL JDBC Driver](https://jdbc.postgresql.org/documentation/head/ssl-client.html)
  - Supports all 6 modes.

---

Thanks for considering this change and please let me know how I can polish this further for acceptance 🙏

### DIFF
--- a/packages/pg-connection-string/README.md
+++ b/packages/pg-connection-string/README.md
@@ -87,11 +87,11 @@ Query parameters follow a `?` character, including the following special query p
  * `host=<host>` - sets `host` property, overriding the URL's host
  * `encoding=<encoding>` - sets the `client_encoding` property
  * `ssl=1`, `ssl=true`, `ssl=0`, `ssl=false` - sets `ssl` to true or false, accordingly
- * `sslmode=<sslmode>`
+* `sslmode=<sslmode>`
    * `sslmode=disable` - sets `ssl` to false
-   * `sslmode=no-verify`, `sslmode=prefer`, - sets `ssl` to `{ rejectUnauthorized: false }`
-   * `sslmode=require`, - sets `ssl` to `{ rejectUnauthorized: false }` unless `sslrootcert` is specified, in which case it behaves like `verify-ca`
-   * `sslmode=verify-ca` - sets `ssl` to `{ checkServerIdentity: no-op}` (verify CA, but not server identity)
+   * `sslmode=no-verify`, `sslmode=prefer` - sets `ssl` to `{ rejectUnauthorized: false }`
+   * `sslmode=require` - sets `ssl` to `{ rejectUnauthorized: false }` unless `sslrootcert` is specified, in which case it behaves like `verify-ca`
+   * `sslmode=verify-ca` - sets `ssl` to `{ checkServerIdentity: no-op }` (verify CA, but not server identity)
    * `sslmode=verify-full` - sets `ssl` to `{}` (verify CA and server identity)
  * `sslcert=<filename>` - reads data from the given file and includes the result as `ssl.cert`
  * `sslkey=<filename>` - reads data from the given file and includes the result as `ssl.key`
--- a/packages/pg-connection-string/README.md
+++ b/packages/pg-connection-string/README.md
@@ -90,7 +90,7 @@ Query parameters follow a `?` character, including the following special query p
  * `sslcompat=libpq` - use libpq semantics for `sslmode`
  * `sslmode=<sslmode>` when `sslcompat=libpq`
    * `sslmode=disable` - sets `ssl` to false
-   * `sslmode=no-verify`, `sslmode=prefer` - sets `ssl` to `{ rejectUnauthorized: false }`
+   * `sslmode=prefer` - sets `ssl` to `{ rejectUnauthorized: false }`
    * `sslmode=require` - sets `ssl` to `{ rejectUnauthorized: false }` unless `sslrootcert` is specified, in which case it behaves like `verify-ca`
    * `sslmode=verify-ca` - sets `ssl` to `{ checkServerIdentity: no-op }` (verify CA, but not server identity). This verifies the presented certificate against the effective CA, i.e. the one specified in sslrootcert or the system CA if sslrootcert was not specified.
    * `sslmode=verify-full` - sets `ssl` to `{}` (verify CA and server identity)
--- a/packages/pg-connection-string/README.md
+++ b/packages/pg-connection-string/README.md
@@ -89,8 +89,10 @@ Query parameters follow a `?` character, including the following special query p
  * `ssl=1`, `ssl=true`, `ssl=0`, `ssl=false` - sets `ssl` to true or false, accordingly
  * `sslmode=<sslmode>`
    * `sslmode=disable` - sets `ssl` to false
-   * `sslmode=no-verify` - sets `ssl` to `{ rejectUnauthorized: false }`
-   * `sslmode=prefer`, `sslmode=require`, `sslmode=verify-ca`, `sslmode=verify-full` - sets `ssl` to true
+   * `sslmode=no-verify`, `sslmode=prefer`, - sets `ssl` to `{ rejectUnauthorized: false }`
+   * `sslmode=require`, - sets `ssl` to `{ rejectUnauthorized: false }` unless `sslrootcert` is specified, in which case it behaves like `verify-ca`
+   * `sslmode=verify-ca` - sets `ssl` to `{ checkServerIdentity: no-op}` (verify CA, but not server identity)
+   * `sslmode=verify-full` - sets `ssl` to `{}` (verify CA and server identity)
  * `sslcert=<filename>` - reads data from the given file and includes the result as `ssl.cert`
  * `sslkey=<filename>` - reads data from the given file and includes the result as `ssl.key`
  * `sslrootcert=<filename>` - reads data from the given file and includes the result as `ssl.ca`
--- a/packages/pg-connection-string/README.md
+++ b/packages/pg-connection-string/README.md
@@ -87,12 +87,17 @@ Query parameters follow a `?` character, including the following special query p
  * `host=<host>` - sets `host` property, overriding the URL's host
  * `encoding=<encoding>` - sets the `client_encoding` property
  * `ssl=1`, `ssl=true`, `ssl=0`, `ssl=false` - sets `ssl` to true or false, accordingly
-* `sslmode=<sslmode>`
+ * `sslcompat=libpq` - use libpq semantics for `sslmode`
+ * `sslmode=<sslmode>` when `sslcompat=libpq`
    * `sslmode=disable` - sets `ssl` to false
    * `sslmode=no-verify`, `sslmode=prefer` - sets `ssl` to `{ rejectUnauthorized: false }`
    * `sslmode=require` - sets `ssl` to `{ rejectUnauthorized: false }` unless `sslrootcert` is specified, in which case it behaves like `verify-ca`
    * `sslmode=verify-ca` - sets `ssl` to `{ checkServerIdentity: no-op }` (verify CA, but not server identity)
    * `sslmode=verify-full` - sets `ssl` to `{}` (verify CA and server identity)
+ * `sslmode=<sslmode>` when `sslcompat` is not set
+   * `sslmode=disable` - sets `ssl` to false
+   * `sslmode=no-verify` - sets `ssl` to `{ rejectUnauthorized: false }`
+   * `sslmode=prefer`, `sslmode=require`, `sslmode=verify-ca`, `sslmode=verify-full` - sets `ssl` to true
  * `sslcert=<filename>` - reads data from the given file and includes the result as `ssl.cert`
  * `sslkey=<filename>` - reads data from the given file and includes the result as `ssl.key`
  * `sslrootcert=<filename>` - reads data from the given file and includes the result as `ssl.ca`
--- a/packages/pg-connection-string/README.md
+++ b/packages/pg-connection-string/README.md
@@ -87,7 +87,7 @@ Query parameters follow a `?` character, including the following special query p
  * `host=<host>` - sets `host` property, overriding the URL's host
  * `encoding=<encoding>` - sets the `client_encoding` property
  * `ssl=1`, `ssl=true`, `ssl=0`, `ssl=false` - sets `ssl` to true or false, accordingly
- * `sslcompat=libpq` - use libpq semantics for `sslmode`
+ * `uselibpqcompat=true` - use libpq semantics
  * `sslmode=<sslmode>` when `sslcompat` is not set
    * `sslmode=disable` - sets `ssl` to false
    * `sslmode=no-verify` - sets `ssl` to `{ rejectUnauthorized: false }`
--- a/packages/pg-connection-string/README.md
+++ b/packages/pg-connection-string/README.md
@@ -96,7 +96,7 @@ Query parameters follow a `?` character, including the following special query p
    * `sslmode=disable` - sets `ssl` to false
    * `sslmode=prefer` - sets `ssl` to `{ rejectUnauthorized: false }`
    * `sslmode=require` - sets `ssl` to `{ rejectUnauthorized: false }` unless `sslrootcert` is specified, in which case it behaves like `verify-ca`
-   * `sslmode=verify-ca` - sets `ssl` to `{ checkServerIdentity: no-op }` (verify CA, but not server identity). This verifies the presented certificate against the effective CA, i.e. the one specified in sslrootcert or the system CA if sslrootcert was not specified.
+   * `sslmode=verify-ca` - sets `ssl` to `{ checkServerIdentity: no-op }` (verify CA, but not server identity). This verifies the presented certificate against the effective CA specified in sslrootcert.
    * `sslmode=verify-full` - sets `ssl` to `{}` (verify CA and server identity)
  * `sslcert=<filename>` - reads data from the given file and includes the result as `ssl.cert`
  * `sslkey=<filename>` - reads data from the given file and includes the result as `ssl.key`
--- a/packages/pg-connection-string/README.md
+++ b/packages/pg-connection-string/README.md
@@ -92,7 +92,7 @@ Query parameters follow a `?` character, including the following special query p
    * `sslmode=disable` - sets `ssl` to false
    * `sslmode=no-verify`, `sslmode=prefer` - sets `ssl` to `{ rejectUnauthorized: false }`
    * `sslmode=require` - sets `ssl` to `{ rejectUnauthorized: false }` unless `sslrootcert` is specified, in which case it behaves like `verify-ca`
-   * `sslmode=verify-ca` - sets `ssl` to `{ checkServerIdentity: no-op }` (verify CA, but not server identity)
+   * `sslmode=verify-ca` - sets `ssl` to `{ checkServerIdentity: no-op }` (verify CA, but not server identity). This verifies the presented certificate against the effective CA, i.e. the one specified in sslrootcert or the system CA if sslrootcert was not specified.
    * `sslmode=verify-full` - sets `ssl` to `{}` (verify CA and server identity)
  * `sslmode=<sslmode>` when `sslcompat` is not set
    * `sslmode=disable` - sets `ssl` to false
--- a/packages/pg-connection-string/README.md
+++ b/packages/pg-connection-string/README.md
@@ -88,18 +88,21 @@ Query parameters follow a `?` character, including the following special query p
  * `encoding=<encoding>` - sets the `client_encoding` property
  * `ssl=1`, `ssl=true`, `ssl=0`, `ssl=false` - sets `ssl` to true or false, accordingly
  * `sslcompat=libpq` - use libpq semantics for `sslmode`
+ * `sslmode=<sslmode>` when `sslcompat` is not set
+   * `sslmode=disable` - sets `ssl` to false
+   * `sslmode=no-verify` - sets `ssl` to `{ rejectUnauthorized: false }`
+   * `sslmode=prefer`, `sslmode=require`, `sslmode=verify-ca`, `sslmode=verify-full` - sets `ssl` to true
  * `sslmode=<sslmode>` when `sslcompat=libpq`
    * `sslmode=disable` - sets `ssl` to false
    * `sslmode=prefer` - sets `ssl` to `{ rejectUnauthorized: false }`
    * `sslmode=require` - sets `ssl` to `{ rejectUnauthorized: false }` unless `sslrootcert` is specified, in which case it behaves like `verify-ca`
    * `sslmode=verify-ca` - sets `ssl` to `{ checkServerIdentity: no-op }` (verify CA, but not server identity). This verifies the presented certificate against the effective CA, i.e. the one specified in sslrootcert or the system CA if sslrootcert was not specified.
    * `sslmode=verify-full` - sets `ssl` to `{}` (verify CA and server identity)
- * `sslmode=<sslmode>` when `sslcompat` is not set
-   * `sslmode=disable` - sets `ssl` to false
-   * `sslmode=no-verify` - sets `ssl` to `{ rejectUnauthorized: false }`
-   * `sslmode=prefer`, `sslmode=require`, `sslmode=verify-ca`, `sslmode=verify-full` - sets `ssl` to true
  * `sslcert=<filename>` - reads data from the given file and includes the result as `ssl.cert`
  * `sslkey=<filename>` - reads data from the given file and includes the result as `ssl.key`
  * `sslrootcert=<filename>` - reads data from the given file and includes the result as `ssl.ca`
 
 A bare relative URL, such as `salesdata`, will indicate a database name while leaving other properties empty.
+
+> [!CAUTION]
+> Choosing an sslmode other than verify-full has serious security implications. Please read https://www.postgresql.org/docs/current/libpq-ssl.html#LIBPQ-SSL-SSLMODE-STATEMENTS to understand the trade-offs.
--- a/packages/pg-connection-string/index.js
+++ b/packages/pg-connection-string/index.js
@@ -107,6 +107,11 @@ function parse(str) {
         break
       }
       case 'verify-ca': {
+        if (!config.ssl.ca) {
+          throw new Error(
+            'SECURITY WARNING: Using sslmode=verify-ca requires specifying a CA with sslrootcert. If a public CA is used, verify-ca allows connections to a server that somebody else may have registered with the CA, making you vulnerable to Man-in-the-Middle attacks. Either specify a custom CA certificate with sslrootcert parameter or use sslmode=verify-full for proper security.'
+          )
+        }
         config.ssl.checkServerIdentity = function () {}
         break
       }
--- a/packages/pg-connection-string/index.js
+++ b/packages/pg-connection-string/index.js
@@ -87,11 +87,11 @@ function parse(str, options = {}) {
     config.ssl.ca = fs.readFileSync(config.sslrootcert).toString()
   }
 
-  if (options.useLibpqCompat && config.sslcompat) {
-    throw new Error('Both useLibpqCompat and sslcompat are set. Please use only one of them.')
+  if (options.useLibpqCompat && config.uselibpqcompat) {
+    throw new Error('Both useLibpqCompat and uselibpqcompat are set. Please use only one of them.')
   }
 
-  if (config.sslcompat === 'libpq' || options.useLibpqCompat) {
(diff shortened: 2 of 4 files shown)


packages/pg-connection-string/README.md
```diff
@@ -68,8 +68,10 @@ Query parameters follow a `?` character, including the following special query p
  * `ssl=1`, `ssl=true`, `ssl=0`, `ssl=false` - sets `ssl` to true or false, accordingly
  * `sslmode=<sslmode>`
    * `sslmode=disable` - sets `ssl` to false
-   * `sslmode=no-verify` - sets `ssl` to `{ rejectUnauthorized: false }`
-   * `sslmode=prefer`, `sslmode=require`, `sslmode=verify-ca`, `sslmode=verify-full` - sets `ssl` to true
+   * `sslmode=no-verify`, `sslmode=prefer`, - sets `ssl` to `{ rejectUnauthorized: false }`
+   * `sslmode=require`, - sets `ssl` to `{ rejectUnauthorized: false }` unless `sslrootcert` is specified, in which case it behaves like `verify-ca`
+   * `sslmode=verify-ca` - sets `ssl` to `{ checkServerIdentity: no-op}` (verify CA, but not server identity)
```
charmander, line 73:
```suggestion
   * `sslmode=no-verify`, `sslmode=prefer` - sets `ssl` to `{ rejectUnauthorized: false }`
   * `sslmode=require` - sets `ssl` to `{ rejectUnauthorized: false }` unless `sslrootcert` is specified, in which case it behaves like `verify-ca`
   * `sslmode=verify-ca` - sets `ssl` to `{ checkServerIdentity: no-op }` (verify CA, but not server identity)
```

packages/pg-connection-string/test/parse.js
```diff
@@ -258,19 +258,24 @@ describe('parse', function () {
...
+      rejectUnauthorized: false,
+    })
   })
 
   it('configuration parameter sslmode=verify-ca', function () {
     var connectionString = 'pg:///?sslmode=verify-ca'
     var subject = parse(connectionString)
-    subject.ssl.should.eql({})
+    subject.ssl.should.have.property('checkServerIdentity').that.is.a('function')
+    expect(subject.ssl.checkServerIdentity()).be.undefined
```
charmander, line 278:
```suggestion
    expect(subject.ssl.checkServerIdentity()).to.be.undefined
```

packages/pg-connection-string/test/parse.js
```diff
@@ -282,9 +287,9 @@ describe('parse', function () {
   it('configuration parameter ssl=true and sslmode=require still work with sslrootcert=/path/to/ca', function () {
     var connectionString = 'pg:///?ssl=true&sslrootcert=' + __dirname + '/example.ca&sslmode=require'
     var subject = parse(connectionString)
-    subject.ssl.should.eql({
-      ca: 'example ca\n',
-    })
+    subject.ssl.should.have.property('ca', 'example ca\n')
+    subject.ssl.should.have.property('checkServerIdentity').that.is.a('function')
+    expect(subject.ssl.checkServerIdentity()).be.undefined
```
charmander, line 292:
```suggestion
    expect(subject.ssl.checkServerIdentity()).to.be.undefined
```

charmander (approved):

cyx:
@charmander do we have an idea when this will land / get released as part of pg9?

ceefour:
Please merge this as it fixes #2281 and #2375

ceefour:
Help @brianc to merge this

ThisIsMissEm:
@brianc / @hjr3 what's the status on this? We've a PR on Mastodon blocked by this change: https://github.com/mastodon/mastodon/pull/25483

charmander:
@ThisIsMissEm Is that PR blocked on this? It looks more like it would be made unnecessary by this.

ThisIsMissEm:
> @ThisIsMissEm Is that PR blocked on this? It looks more like it would be made unnecessary by this.

I'm saying "blocked" in the sense of if upstream is going to fix, that'd be preferable to a fix in our code for it

charmander:
@ThisIsMissEm On second read, it looks like the Mastodon PR is necessary either way. Unless you were planning on deleting all of that code if this were merged and released soon enough?

hjr3:
This PR introduces breaking changes. We should probably:

- merge https://github.com/brianc/node-postgres/pull/3128
- release version 2.7.0
- merge this PR
- see if there are any other breaking changes that need to be made
- release version 3.0

ThisIsMissEm:
> This PR introduces breaking changes. We should probably:
> 
> * merge [feat(pg-connection-string): ClientConfig helper functions #3128](https://github.com/brianc/node-postgres/pull/3128)
> * release version 2.7.0
> * merge this PR
> * see if there are any other breaking changes that need to be made
> * release version 3.0

@hjr3 that sounds good to me!

ThisIsMissEm:
Just wanted to follow up and see if there's any plans for forwards progress on this?

Trott:
[Over on Mastodon](https://indieweb.social/@thisismissem@hachyderm.io/113472286880311310), @ThisIsMissEm asked if I might know some Node.js TLS/SSL people who might be able to review this. I hope I'm not pinging too widely, but I'm certainly interested in trying to unblock things for the Mastodon project so....hey, can anyone review? @indutny @jasnell @joyeecheung @tniessen @pimterry @panva @mertcanaltin

charmander:
I’m not really up to speed on the Mastodon context here, but some information, if it helps:

- This implementation is (AFAIK) as correct as it can be in terms of translating libpq `sslmode`s to pg configuration. But [although it’s documented, there is one `sslmode` hazard in libpq](https://www.postgresql.org/docs/current/libpq-ssl.html#LIBPQ-SSL-PROTECTION) we could probably improve on.

    > The difference between `verify-ca` and `verify-full` depends on the policy of the root CA. If a public CA is used, `verify-ca` allows connections to a server that somebody else may have registered with the CA. In this case, `verify-full` should always be used. If a local CA is used, or even a self-signed certificate, using `verify-ca` often provides enough protection.

- I’ve already approved this PR, but since it’s a breaking change it isn’t going to be a default until pg 9. There are a lot of changes that would ideally go in pg 9, and I don’t expect it to come out soon. Maybe a new major version of pg-connection-string could be released sooner, if you’d be interested in adding a dependency on a newer version of pg-connection-string and using it manually to pass pg a configuration object instead of a connection string.

- As @ThisIsMissEm mentioned in the linked pull request, pg doesn’t support `sslmode=prefer` (and it doesn’t support `allow` either). I think `prefer` is a bad idea (because it adds complexity and provides a false sense of security), assume/hope it’s only the libpq default for historical reasons, and don’t expect pg to support it in the future.

ThisIsMissEm:
@charmander `sslmode=prefer` is explicitly what our users have asked for because we support it on the ruby side of the project which uses the native pg implementation

packages/pg-connection-string/index.js
```diff
@@ -87,15 +87,26 @@ function parse(str) {
...
+      if (config.sslrootcert) {
+        // If a root CA is specified, behavior of `sslmode=require` will be the same as that of `verify-ca`
+        config.ssl.checkServerIdentity = function () {}
+      } else {
+        config.ssl.rejectUnauthorized = false
+      }
+      break
+    }
+    case 'verify-ca': {
+      config.ssl.checkServerIdentity = function () {}
```
charmander, line 104: On second thought, even though libpq has and has documented this same problem, maybe we should enforce that `config.ssl.ca` is truthy here.

```suggestion
      if (!config.ssl.ca) {
        throw new Error('sslmode=verify-ca requires specifying a CA with sslrootcert')
      }
      config.ssl.checkServerIdentity = function () {}
```

The main issue with this would be for anyone who wants to add a `ca` key to the returned `config.ssl` *after* parsing. I don’t think a warning is the way to go (too easy for those who need it to miss or ignore, and annoying for those who don’t need it to suppress). Would adding a whole options parameter to support this (`parse(str, {deferSslCa: true})` – probably a better name option) be too much? (Related issue, merging configuration with a connection string is hard in general: #3327)

charmander (dismissed):

charmander:
> @charmander `sslmode=prefer` is explicitly what our users have asked for because we support it on the ruby side of the project which uses the native pg implementation

@ThisIsMissEm In the end it’s up to @brianc, but I’m against supporting `sslmode=prefer` in pg – it’s convenient and unsafe, which isn’t a good combo.

Looking back at more of the context and questions, e.g.
> I'm wanting an expert opinion on the changes in the upstream PR before I make any further changes here, since I'm by no means an expert on TLS/SSL, and I'm cautious about introducing accidental security vulnerabilities here.

`sslmode=require` is implemented in this PR without adding any vulnerabilities[^1] *relative to libpq*, but libpq’s `sslmode=require` is inherently vulnerable to MITM even though it might sound like “require normal SSL” by the name – it doesn’t check certificates at all, so it only provides opportunistic encryption. Most people should be using either `sslmode=verify-ca` (with `sslrootcert`) or `sslmode=verify-full`. `verify-full` is the one that gives the basic guarantees of TLS that are defaults in places like Node’s `node:tls` module (+ most other platforms) and web browsers.

(Re: managed database providers: It looks like DigitalOcean, the managed database provider in https://github.com/mastodon/mastodon/issues/11445, is giving bad advice by encouraging people to connect with `sslmode=require` over the internet, and [only mentioning specifying `sslrootcert` as an afterthought](https://docs.digitalocean.com/products/databases/postgresql/how-to/connect/#download-the-ssl-certificate), plus a quick link to libpq documentation. Heroku is even worse: [`require` over the internet](https://devcenter.heroku.com/articles/connecting-to-heroku-postgres-databases-from-outside-of-heroku#connect-via-a-desktop-sql-client), [`require` from generic code](https://devcenter.heroku.com/articles/connecting-heroku-postgres#connecting-in-node-js), no mention of the lack of guarantees anywhere. [Google Cloud SQL is fully safe](https://cloud.google.com/static/sql/docs/postgres/connect-connectors#node.js), but it does its own thing and passes pre-connected streams to pg. [AWS RDS documents a secure way to do it](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/PostgreSQL.Concepts.General.SSL.html#PostgreSQL.Concepts.General.SSL.Connecting), as well as what look like some less secure ways.)

[^1]: but I’m not the Node TLS expert opinion requested in pings, just a pg maintainer implementing a few related features in the context of this library/libpq connection strings/the PostgreSQL wire protocol, so looking forward to that review too.

hjr3:
@charmander 

> I’ve already approved this PR, but since it’s a breaking change it isn’t going to be a default until pg 9.

I suggested a path forward here: https://github.com/brianc/node-postgres/pull/2709#issuecomment-1905837754

It is not clear to me why we would need to wait until pg 9 to release pg-connection-string v3.0

charmander:
@hjr3 As I mentioned in the same comment:

> Maybe a new major version of pg-connection-string could be released sooner, if you’d be interested in adding a dependency on a newer version of pg-connection-string and using it manually to pass pg a configuration object instead of a connection string.

i.e. we don’t need to wait, but the new major version of pg-connection-string won’t be the default until pg 9, so people will still need to make application-side changes.

hjr3:
> @hjr3 As I mentioned in the same comment:
> 
> > Maybe a new major version of pg-connection-string could be released sooner, if you’d be interested in adding a dependency on a newer version of pg-connection-string and using it manually to pass pg a configuration object instead of a connection string.
> 
> i.e. we don’t need to wait, but the new major version of pg-connection-string won’t be the default dependency in pg (that it uses to parse `connectionString`) until pg 9, so people will still need to make application-side changes.

I somehow missed that. My apologies. I understand and agree with what you are saying.

I do think it is fairly common for teams to use newer versions of pg-connection-string with pg.

ThisIsMissEm:
In mastodon, we already directly pull in `pg-connection-string`, specifically to try to better handle connection options from server operators. Would be fully in favor of a release that includes this and #3128. Would sponsorship or something help move this along? (I might be able to get some support for that)

hjr3:
I want to continue working on `pg-connection-string` and I do not need sponsorship. The goal of https://github.com/iceddev/pg-connection-string/issues/37 was to make it easier to keep packages in sync with each other. Unfortunately, I lost my merge permissions when the transition was made.

If @brianc can give me merge permissions, I can move the package forward. If there are concerns with my ability to merge other packages, I started https://github.com/brianc/node-postgres/pull/3129 which should allow us to enforce the proper permissions.

brianc:
hey sorry I had the flu last week...literally never been that sick before in my life. still only about 80% back to health.  Yeah @hjr3 I've "known" you for a long time on github, and trust you. happy to give you merge permissions!  You might need to ping me if you need a critical release done to npm at any time, but even though sometimes I go awol for a week or two I am always watching the repo...just hard to do computer stuff when you think you might actually die from the flu!  Instead of merging this myself I'm going to give you permissions @hjr3 so we can make sure they work. :)

brianc:
@hjr3 sent you a collaborator invite.

hjr3:
I originally wanted to merge this in as-is and publish a 3.0.0. I now wonder if we should add a flag (.e.g. `sslcompat=libpq`) to enable this behavior in 2.x.x so users have an upgrade path. We can then make this the default behavior in 3.0.0.

ThisIsMissEm:
> I originally wanted to merge this in as-is and publish a 3.0.0. I now wonder if we should add a flag (.e.g. `sslcompat=libpq`) to enable this behavior in 2.x.x so users have an upgrade path. We can then make this the default behavior in 3.0.0.

I'd probably say just release 3.0.0 with this as default and thoroughly document the breaking changes

packages/pg-connection-string/index.js
hjr3: added the check in https://github.com/brianc/node-postgres/pull/2709/commits/74a314ef959bcddd323c138eaacc1b8212e991ed

pmalouin: Is it really needed to enforce `config.ssl.ca` for this sslmode?

If `config.ssl.ca` is not specified, the TLS client will use the CA certificates already available via `tls.getCACertificates('default')`, which can include certificates included in the bundled certificate store, the system's trusted store and any extra certificate configured via the `NODE_EXTRA_CA_CERTS` env variable. [Docs](https://nodejs.org/api/tls.html#tlsgetcacertificatestype)

In our implementation, we use the latter option (`NODE_EXTRA_CA_CERTS`) to configure the CA certificates from our Postgres endpoint and it works fine.

hjr3: TIL

I will unwind this change. It will match the behavior in https://github.com/lib/pq/commit/585ef5d6d502fd43589d48e448b91b449745e369#diff-4f427d2b022907c552328e63f137561f6de92396d7a6e8f6c2ea1bcf0db52654R836-R838

charmander: @pmalouin This is exactly why I thought a check might be useful :) Anyone can get a valid certificate from a system-trusted CA, simply by using a domain that they actually own. Without `config.ssl.ca` enforced, such a certificate can then be used to MITM any `sslmode=verify-ca` connection, providing essentially the same security level as `sslmode=require`.

charmander (dismissed):
same as my previous review

packages/pg-connection-string/index.js
hjr3: This behavior seems consistent, well documented and understood by those who want libpq compatibility. If someone has a MITM concern, then they should use `sslmode=verify-full` or specify a more narrow `sslrootcert`.

Also, I am not sure the proposed check will have the intended results. If a user encounters this error, they will almost certainly add `&sslrootcert=/etc/ssl/certs/ca-certificates.crt` to the connection string and move on. In fact, I originally thought specifying the system CA (or a custom CA) was the intention of the check because I did not realize that the TLS client will use the CA certificates already available.

All that being said, I did see the following chart in the JDBC docs which outlines the risks in a way that is easy to understand. I think we should incorporate a similar chart into the README so people are aware of the MITM risks.

image

charmander: I mentioned that it was documented by libpq in my original comment, but I think @pmalouin’s comment demonstrates that the implications are still not obvious to people. A more descriptive error could be a good idea. I don’t think any amount of documentation is sufficient to replace the check.

hjr3: I appreciate the value of guiding people in a safe direction. The current behavior is is unchanged: `prefer`, `require` and `verify-ca` are all aliases `verify-full` by default.

The proposed changes are behind a flag that explicitly asks for libpq compatibility. If someone is explicitly asking for libpq behavior by adding `sslcompat=libpq` then we should be consistent with libpq's behavior.

pmalouin: Makes sense to me too. The check would act as a safe default, steering users towards the safer way to use `verify-ca`: explicitly setting the CA that they expect to be used by their server, instead of opening it up to any system trusted CA by default 👍

pmalouin: Also, in the spirit of following libpq semantics more closely, it appears that libpq does not use the system trusted certificates by default (ex. when `sslrootcert` is omitted). So this would suggest to do the same here.

libpq also supports the `sslrootcert=system` [option](https://www.postgresql.org/docs/16/libpq-connect.html#LIBPQ-CONNECT-SSLROOTCERT) that specifies to use the system trusted CAs, but libpq would return an error if you use that in combination with `sslmode=verify-ca`.

brianc:
> I'd probably say just release 3.0.0 with this as default and thoroughly document the breaking changes

since this is part of the monorepo it isn't independently versioned exactly. I need to maintain backwards compatibility with node-postgres as it stands now unless we want to do a major version bump of PG if this is a breaking change....which I'd _rather_ avoid doing. An opt-in flag or some way to turn this on _via_ the connection string would be a nice default behavior for now, and then when I flip to `pg@9.x` and have the proper documentation & upgrade guide and stuff we can switch the defaults.  Does that help / make sense @hjr3 ?  Thanks for your help on this btw!

brianc:
note: the tests are failing due to some istambul test coverage metric I never set up myself...feel free to lower the threshold slightly if that's the nudge it needs to get over the finish line. :)

hjr3:
The more I think about this, the more I am convinced that this should be behind a flag and then we can figure out in the far-term future if we ever want to make this a default. I am very concerned about changing something like `sslmode=prefer` from being an alias to `sslmode=verify-full` to a much weaker sslmode.

All that is to say: no major version upgrade is planned.

brianc:
> All that is to say: no major version upgrade is planned.

awesome - lmk if you need help gettin' the PR over the finish line! I'm all in favor of not breaking changes! =D 

> behind a flag

that would be like another, optional, non-"standard" query param in the connection string, yea?

packages/pg-connection-string/index.js
hjr3: I think where we are at is:

1. put the check back in place to match libpq semantics
2. improve the error message to warn folks about mitm attacks
3. make the sslmode documentation tradeoffs more clear, especially in libpq mode

I will send an updated PR soon :tm:

brianc:
radical!

packages/pg-connection-string/index.js
```diff
@@ -87,20 +87,49 @@ function parse(str) {
...
-    case 'disable': {
-      config.ssl = false
-      break
-    }
-    case 'prefer':
-    case 'require':
-    case 'verify-ca':
-    case 'verify-full': {
-      break
+  if (config.sslcompat === 'libpq') {
```
ThisIsMissEm, line 90: rather than using a value from the connection string, I think it'd be better to have `parse` take an option: `parse(str: string, useLibpqCompat: boolean)` — that way connection URLs don't have non-standard parameters?

Xiretza: Then every application using this library has to implement this out-of-band, right? That's simply not going to happen, so from a user perspective this effectively wouldn't fix the issue of not supporting standard libpq connection strings.

ThisIsMissEm: I think this would be an application concern, not a user choice?

hjr3: Many libraries that wrap `pg` accept a DSN. Those libraries would have to be updated to have whatever function that accepts the DSN also accept this extra parameter.

packages/pg-connection-string/index.js
hjr3: I believe https://github.com/brianc/node-postgres/pull/2709/commits/165b5c6e05dadbbea9175a312ccdb4cbaf28c500 resolves the raised concerns. Please let me know if there is still more to discuss.

charmander:
I don’t know what the best way forward is. It seems like there should be at least a code-side way to set the compatibility flag, as @ThisIsMissEm brought up, if we’re going the route of a compatibility flag? This doesn’t preclude supporting `sslcompat=libpq` in the query string, if people think that’s also important enough. Or the equivalent of a code-side compatibility flag could be a separate major version of pg-connection-string that implements this as a breaking change…?

Anyway, my suggestion would be:

- new major version of pg-connection-string using the more libpq-compatible interpretation of `sslmode`
- pg continues to depend on the same older version range of pg-connection-string, with any future changes to pg-connection-string informed by that (i.e. maybe more backporting than usual)
- introduce a deprecation warning on the old version of pg-connection-string when `sslmode`s of `require` or `verify-ca` are used, recommending better replacements:
    - `verify-full` if the connection string is already being used successfully with no further changes, or
    - installing the new version of pg-connection-string and calling into it directly instead of going through pg’s `connectionString` option, if the library user is currently patching the `ssl` property after parsing to achieve the effect of one of the modes
- another deprecation warning with just the latter recommendation (plus mention of the standard form `sslmode=require`) when `sslmode=no-verify` is used

Side note: It’s important to fix [the pg surprise where the entire `ssl` configuration is replaced if SSL-related options exist in `connectionString`](https://node-postgres.com/features/ssl#usage-with-connectionstring) before `sslmode=require` becomes weaker in the interpretation of `connectionString` – i.e. in pg@9, under this plan – but it’s less of an issue when mixing pg@8 with the newer pg-connection-string, since the option merging is then exposed as top-level JavaScript operations with clear precedence.

packages/pg-connection-string/README.md
```diff
@@ -87,12 +87,22 @@ Query parameters follow a `?` character, including the following special query p
...
+ * `sslcompat=libpq` - use libpq semantics for `sslmode`
+ * `sslmode=<sslmode>` when `sslcompat` is not set
    * `sslmode=disable` - sets `ssl` to false
    * `sslmode=no-verify` - sets `ssl` to `{ rejectUnauthorized: false }`
    * `sslmode=prefer`, `sslmode=require`, `sslmode=verify-ca`, `sslmode=verify-full` - sets `ssl` to true
+ * `sslmode=<sslmode>` when `sslcompat=libpq`
+   * `sslmode=disable` - sets `ssl` to false
+   * `sslmode=prefer` - sets `ssl` to `{ rejectUnauthorized: false }`
+   * `sslmode=require` - sets `ssl` to `{ rejectUnauthorized: false }` unless `sslrootcert` is specified, in which case it behaves like `verify-ca`
+   * `sslmode=verify-ca` - sets `ssl` to `{ checkServerIdentity: no-op }` (verify CA, but not server identity). This verifies the presented certificate against the effective CA, i.e. the one specified in sslrootcert or the system CA if sslrootcert was not specified.
```
pmalouin, line 99:
```suggestion
   * `sslmode=verify-ca` - sets `ssl` to `{ checkServerIdentity: no-op }` (verify CA, but not server identity). This verifies the presented certificate against the effective CA specified in sslrootcert.
```

hjr3:
I plan on merging this in a few days unless someone objects.

jawj:
Guess I'm too late to the party, but I'm not a big fan of this change. I think node-postgres has historically done well to provide _the connection security many people assumed they were getting from libpq even when they weren't_ for `sslmode=require`. So IMHO it would be better to keep things as they have been, for now, in the reasonable hope that libpq will at some point rename/improve its own connection security options.

Reposting my comments from #3473:

> Usually I would be all for adopting the standard or upstream semantics — but I actually think this is an example where libpq gets it badly wrong and current node-postgres has historically done much better. libpq's `sslmode=require` really sounds secure but [it absolutely is not](https://neon.com/blog/postgres-needs-better-connection-security-defaults): this is a serious footgun that people are shooting themselves in the foot with all the time.
> 
> The change here means that, if and when developers upgrade node-postgres to 9.0, connections that used to be secure will suddenly become insecure. I mean, _hopefully_ they'll see the warning and do something about it ... but surely many will not.
> 
> IMHO it would be much better to have users actively opt in to this downgrade in TLS protection. That would mean keeping the `uselibpqcompat` flag, and keeping it `false` by default.
> 
> As I understand it, node-postgres historically treats every `sslmode` from `require` upwards as not just `verify-full`, but as `sslrootcert=system`. IMHO this is also behaviour that libpq got wrong and node-postgres got right. If the developer asks for `verify-full` and doesn't provide a root CA cert, libpq throws. In the same situation, node-postgres validates against the system root CA certs. That matches the behaviour of https and is the logic I think most current developers would expect.
> 
> Is it proposed to match libpq here too, and throw an error if the user specifies `sslmode=verify-full` without also providing a certificate? I don't think it should be — but I do think that's where the logic of matching libpq's behaviour takes you (i.e. demanding `sslrootcert=system` for the behaviour that used to be provided by `sslmode=require` and above).

Issue #2716, opened 2022-03-06
client.end() hangs if called twice


cody-greene:
```
const client = new pg.Client()
... 
await client.end() // ok
await client.end() // uh-oh! This never resolves!
```

Ideally we should just not try to end the client twice in a row. But in using other packages, that wrap node-postgres, I've encountered a situation where I need to access the pg Client instance and `end()` it, before returning it to some wrapper class which may also try to `end()` it at some point. Arguably this wrapper class should be smarter and avoid calling `end()` if it sees the "end" *event*. But just the same `client.end()` shouldn't hang either.

After a [quick look at the source](https://github.com/brianc/node-postgres/blob/21ccd4f1b6e66774bbf24aecfccdbfe7c9b49238/packages/pg/lib/client.js#L611):
```
return new this._Promise((resolve) => {
  this.connection.once('end', resolve)
})
```
When the client has already ended, we just wait for a "end" *event* which never arrives. I imagine the fix here is to check for a private field like `this._ended` (as a compliment to `this._ending`) and return/resolve early.

PR #2717, opened 2022-03-07
fix: double client.end() hang

cody-greene:
fixes https://github.com/brianc/node-postgres/issues/2716

`client.end()` will resolve early if the connection is already dead, rather than waiting for an "end" event that will never arrive. This also makes `client.end()` resolve only when the underlying socket is fully closed, both the readable part ("end" event) & writable part ("close" event).

https://nodejs.org/docs/latest-v16.x/api/net.html#event-end
> Emitted when the other end of the socket signals the end of
> transmission, thus ending the readable side of the socket.

https://nodejs.org/docs/latest-v16.x/api/net.html#event-close_1
> Emitted once the socket is fully closed.

test before change:
```
2716-tests.js
  client.end() should resolve if already ended
Error: test: client.end() should resolve if already ended did not complete withint 5000ms
    at Timeout._onTimeout (/home/cody/repos/node-postgres/packages/pg/test/suite.js:52:19)
    at listOnTimeout (node:internal/timers:557:17)
    at processTimers (node:internal/timers:500:7)
```

test after change:
```
2716-tests.js
  client.end() should resolve if already ended ✔
```

### DIFF
--- a/packages/pg/test/integration/gh-issues/2716-tests.js
+++ b/packages/pg/test/integration/gh-issues/2716-tests.js
@@ -7,14 +7,32 @@ const suite = new helper.Suite()
 suite.testAsync('client.end() should resolve if already ended', async () => {
   const client = new helper.pg.Client()
   await client.connect()
-  await client.end()
-  // connection "end" event is emitted twice; once on stream "close" and once
-  // on stream "end" so we need to make sure our second client.end() is not
-  // resolved by these.
-  await sleep(1)
-  await client.end() // this should resolve early, rather than waiting forever
-})
 
-function sleep(ms) {
-  return new Promise((resolve) => setTimeout(resolve, ms))
-}
+  // this should resolve only when the underlying socket is fully closed, both
+  // the readable part ("end" event) & writable part ("close" event).
+
+  // https://nodejs.org/docs/latest-v16.x/api/net.html#event-end
+  // > Emitted when the other end of the socket signals the end of
+  // > transmission, thus ending the readable side of the socket.
+
+  // https://nodejs.org/docs/latest-v16.x/api/net.html#event-close_1
+  // > Emitted once the socket is fully closed.
+
+  // here: stream = socket
+
+  await client.end()
+  // connection.end()
+  //   stream.end()
+  // ...
+  // stream emits "end"
+  //   not listening to this event anymore so the promise doesn't resolve yet
+  // stream emits "close"; no more events will be emitted from the stream
+  //   connection emits "end"
+  //     promise resolved
+
+  // This should now resolve immediately, rather than wait for connection.on('end')
+  await client.end()
+
+  // this should resolve immediately, rather than waiting forever
+  await client.end()
+})
--- a/packages/pg/test/integration/gh-issues/2716-tests.js
+++ b/packages/pg/test/integration/gh-issues/2716-tests.js
@@ -1,0 +1,20 @@
+'use strict'
+const helper = require('../test-helper')
+
+const suite = new helper.Suite()
+
+// https://github.com/brianc/node-postgres/issues/2716
+suite.testAsync('client.end() should resolve if already ended', async () => {
+  const client = new helper.pg.Client()
+  await client.connect()
+  await client.end()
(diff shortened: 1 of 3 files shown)


packages/pg/test/integration/gh-issues/2716-tests.js
```diff
@@ -0,0 +1,20 @@
...
+
+// https://github.com/brianc/node-postgres/issues/2716
+suite.testAsync('client.end() should resolve if already ended', async () => {
+  const client = new helper.pg.Client()
+  await client.connect()
+  await client.end()
+  // connection "end" event is emitted twice; once on stream "close" and once
+  // on stream "end" so we need to make sure our second client.end() is not
+  // resolved by these.
+  await sleep(1)
```
charmander, line 14: The `await` on the line above should take care of this, right?

```suggestion
```

cody-greene: Unfortunately it does not. There seems so be some process.nextTick() shenanigans which means that with just:
```
await client.end()
await client.end()
```
The 2nd end() is actually resolved as a result of events emitted from the first end() and the test will succeed without code changes. Adding the sleep() call makes the test actually fail before the code change, and then pass with the code change. If that still doesn't make sense then I can try to point to the source with more  detail.

charmander: Please do. It looks to me like `end` would either have to be emitted twice on the connection for that to happen (bad), or the connection’s `_connecting` would have to transition back to `false`.

cody-greene: To further clarify: the "end" event which resolves the `client.end()` promise/callback is emitted when `client.connection.stream` [emits "end" itself](https://github.com/brianc/node-postgres/blob/21ccd4f1b6e66774bbf24aecfccdbfe7c9b49238/packages/pg/lib/connection.js#L107) i.e. the readable side of the socket is closed, and again when the [stream emits "close"](https://github.com/brianc/node-postgres/blob/21ccd4f1b6e66774bbf24aecfccdbfe7c9b49238/packages/pg/lib/connection.js#L57). The close event may not be emitted on the same tick, and could be several ticks later, basically whenever the socket is fully and finally closed.

So I could actually write this test with 3 `end()`s rather than a `sleep()`
```javascript
await client.end()
// connection.end()
//   stream.end()
// ...
// stream emits "end"
//    connection emits "end"
//       promise resolved
await client.end()
// connection.end()
//   stream.end() but this is already closing and does nothing
// ...
// stream emits "close"; no more events will be emitted from the stream
//   connection emits "end"
//     promise resolved
await client.end() // <-- this will wait forever
```

charmander: Yeah, connection emitting `end` twice doesn’t seem good. Not sure if it’s intended.

cody-greene: Right, I was initially expecting the first `client.end()` to resolve only when the underlying socket is fully closed e.g. `stream.on('close')`. At this point I'm wondering if that's something I should try to change as well. Reason being: resolving before the client is really dead is technically incorrect.

brianc:

cody-greene:
Is anything else needed here?

MasterOdin:
I tested this PR and it will also fix the issue I just reported in #2777. :+1:

cody-greene:
Can I do anything else to help get this merged, @brianc ?

charmander (approved):

mastermatt (approved):

mastermatt:
This will also fix https://github.com/brianc/node-postgres/issues/2923.
@brianc as today is the one year anniversary for this PR, could we get you to take a fresh look at the change?

brianc:
> as today is the one year anniversary for this PR, could we get you to take a fresh look at the change?

haha yes I'll get this released today 😄

cody-greene:
dreams-do-come-true-sm

brianc:
haha yeah sorry it took so long...sometimes things get buried for me in my inbox and go bye bye - appreciate it deeply though! ❤️

PR #2739, opened 2022-05-03
ci: bump github actions

Fdawgs:
Bumps GitHub Actions to their latest major versions. Main change is they're now using Node 16 instead of 12, which was EOL at end of April.

- Changelog for `actions/checkout` can be [found here](https://github.com/actions/checkout/blob/main/CHANGELOG.md)
- Releases for `actions/setup-node` can be [found here](https://github.com/actions/setup-node/releases)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,9 +21,9 @@ jobs:
         os: [ubuntu-latest, windows-latest, macos-latest]
     name: Node.js ${{ matrix.node }} (${{ matrix.os }})
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Setup node
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node }}
           cache: yarn


brianc:
Thanks for doing this! Definitely want to upgrade but....hmmmm any idea why all the CI checks failed?

charmander:
@brianc One of stream-tester’s transitive dependencies. It’s been failing the same way on current master for a while.

Fdawgs:
> @brianc One of stream-tester’s transitive dependencies. It’s been failing the same way on current master for a while.

Thanks charmander, had looked at the errors thrown and couldn't tell why it was failing.

brianc:
> One of stream-tester’s transitive dependencies. It’s been failing the same way on current master for a while.

barf. going to just kill this dep & remove stream tester. hasn't really provided much value anyway.  I'll get that patched up, and then unfortunatley ask PR contributors to rebase on master once I remove this broken dep.  Wish github made that busy-work easier to do behind a button like "bring PR up to date with merge target" or something.

Fdawgs:
Great, thanks @brianc!

Issue #1748, opened 2018-10-12
Exception For database connection Timeout. Store IP of Database in cache


RaviDavda:
When We have created database connection through connection string, it might store ip in the cache in your library and for next every connection it uses that library. Now my database configuration changed but host, username, password is same and private ip of database changed. Now pg library try to connect on old IP and we are having new IP. so My every API gets failed.So I want to know the possible solution for this.

sehrope:
This is not caused by the driver. It doesn't perform any DNS caching. Your operating system / network stack is probably doing it though.

Check the DNS records for your database to see what the TTL (Time To Live) value. That's how many seconds the value can be cached. The default for your DNS provider is probably something large like 3600 (1 hour) or 7200 (2 hours). A smaller value like 60 (1-minute) would lower the failover time.

zaggino:
I ran into similar issue. Our DB is hosted by Amazon's RDS and we connect to it by hostname, when RDS goes through upgrade our apps immediately go to `ECONNREFUSED` mode. I though it to be an IP change behind the hostname so I started making notes of IP before an update, IP after an update and then I realized that the IP doesn't change. Anything else is cached by the library which might be causing this? Restarting the node app makes it connect immediately without problems until next Postgres update.

sysrun:
I think the "problem" ist this: https://github.com/brianc/node-postgres/blob/0993e4b61ad1bafabc84b0335bc9b4caa8b72847/lib/connection-parameters.js#L130

It's only called once upon initialisation

PR #2740, opened 2022-05-03
Using hostaddr will cache the DNS lookup

mjgp2:
"If host is specified without hostaddr, a host name lookup occurs."

https://www.postgresql.org/docs/8.1/libpq.html#LIBPQ-CONNECT

You do not need to do the DNS lookup in javascript.

Fixes: https://github.com/brianc/node-postgres/issues/1748


charmander:
> You do not need to do the DNS lookup in javascript.

The reason it’s done in JavaScript is that it’s blocking in libpq, IIRC.

mjgp2:
It is blocking at within libpq (uses `getaddrinfo`), but it's seemingly done on a separate thread from the comment in the code, so should not block the node event loop?

https://github.com/brianc/node-libpq/blob/master/index.js#L34

FWIW we are experiencing a stale `hostaddr` in the connection string causing downtime on services.

brianc:
Yah looking at this....I think the dns lookup in JS was done before I rewrote libpq bindings and made a proper background async connection worker thread.  So this isn't needed now that we're running libpq's connect in node's threadpool.

brianc:
I'll try rerunning it - very weird error there I didn't on my branches or on [master](https://github.com/brianc/node-postgres/actions/runs/2302982047)

brianc:
It looks like the errors in CI might be related to this PR...the other PR's I've rebased have all passed CI okay.  😢

mjgp2:
@brianc - I rebased, can you try rerunning?

mjgp2:
Also, can you point me towards the settings for running unit tests against the local pg instance

PR #2741, opened 2022-05-03
Add support for password function to native

mjgp2:
This PR allows using a function for the password in the native client (as in the JS client).


brianc:

brianc:
@mjgp2 If you rebase on latest this should be good to go.

mjgp2:
Done @brianc , thanks!

packages/pg/lib/connection-parameters.js
```diff
@@ -125,9 +125,25 @@ class ConnectionParameters {
...
   getLibpqConnectionString(cb) {
-    var params = []
-    add(params, this, 'user')
-    add(params, this, 'password')
+    if (typeof this.password === 'function') {
+      const pw = this.password();
+      if (typeof pw === 'string') {
+        return this._getLibpqConnectionString(cb, pw);
+      }
+      pw.then(pwString => this._getLibpqConnectionString(cb, pwString));
```
charmander, line 133: This probably needs to handle rejections.

charmander (changes requested):

packages/pg/lib/connection-parameters.js
mjgp2: Yep, fixed!

mjgp2:
This should be good to go.

PR #2751, opened 2022-05-29
feat: add connection parameter nativeConnectionString

furiozo-ga:
for when you want to pass connection string AS IS to pg-native
which allows the use of features present in libpq but still not
implemented in the js drivers
like multihost URI


brianc (changes requested):
Sorry for the delay here on my part with the review.  Can you supply a test case that covers this? I typically don't merge code unless its covered by tests.

alexkvak:
There is another one PR with the same changes but with tests https://github.com/brianc/node-postgres/pull/2941

PR #2779, opened 2022-08-05
Support lock_timeout

mkubliniak:
Allow passing `lock_timeout` in connection params.

https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-LOCK-TIMEOUT


brianc:
nice! thank you for this!

PR #2804, opened 2022-09-01
Allow setting `ssl.servername`

### LINKED ISSUE #2803: ssl option `servername` is ignored
The `servername` option of `tls.connect` is necessary if the host for SNI is different from the host for the connection. But node-postgres ignores whatever you set for `servername` and instead uses `host`:

https://github.com/brianc/node-postgres/blob/8250af4aed9b8977932560733fe8665831aeef4d/packages/pg/lib/connection.js#L90-L97

If `options.servername` is specified, node-postgres should *not* override that.

Relevant [Node docs](https://nodejs.org/api/tls.html#tlsconnectoptions-callback):
> Server name for the SNI (Server Name Indication) TLS extension. It is the name of the host being connected to, and must be a host name, and not an IP address. It can be used by a multi-homed server to choose the correct certificate to present to the client

gabegorelick:
Fixes #2803

### DIFF
--- a/packages/pg/lib/connection.js
+++ b/packages/pg/lib/connection.js
@@ -87,7 +87,7 @@ class Connection extends EventEmitter {
         }
       }
 
-      if (net.isIP(host) === 0) {
+      if (!options.servername && net.isIP(host) === 0) {
         options.servername = host
       }
       try {


brianc:
Thanks for this! Is it possible to write a test for this?  Without a test it becomes pretty hard to maintain the code for me over long periods of time.

gabegorelick:
>  Is it possible to write a test for this? 

I'm happy to if someone points me to a good existing test to copy.

PR #2810, opened 2022-09-07
fix(pg-query-stream): invoke `this.callback` on cursor end/error

### LINKED ISSUE #2013: pool.query(new QueryStream(...)) unexpected behaviour
Hi,
if I use the `query` method on pg.Pool with a query object that implements `sumbit` (i.e. pg-query-stream or pg-cursor) like this:
```
import { Pool } from "pg";
import QueryStream = require("pg-query-stream");
const pool = new Pool();
// ...
const query = new QueryStream("SELECT * FROM ...");
const q = pool.query(query);
console.log(q); // Promise { <pending>, ...
await q; //!
```
the returned promise never resolves and the pool never releases the used client

I could ignore the return value and continue with the result, e.g.
```
for await (const row in query) { ...process row...}
```
but the client leak is a problem

furthermore, the typescript definition is plain wrong:
```
export class Pool extends events.EventEmitter {
// ...
    query<T extends Submittable>(queryStream: T): T;
// ...
}
```

aleclarson:
Closes #2013

This issue was caused by two things:
- the `Client` wasn't using the `callback` argument that `Pool#query` passes with the `QueryStream` submittable
- the `QueryStream` wasn't attaching end/error listeners to call its `callback` inside of


brianc:
Thanks for doing this!  Can you include a test to reproduce the issue / prove its fixed? Without tests I am unlikely to merge PRs as they become a future bug waiting to re-occur! 😄

peterp:
I think `pg-copy-streams` may suffer from the same issue, and am happy to provide a failure case/ test.

aleclarson:
@peterp I believe you can send a PR to my fork and it'll show up here when I merge

peterp:
@aleclarson I'm sorry; I'm not going to find enough time to write this.

brianc:
> I'm sorry; I'm not going to find enough time to write this.

While I greatly appreciate the work in fixing this, the PR will remain unmerged until a test case is submitted.  I can't merge untested code as it becomes a forever maintenance burden for me in the future with no way to check for expected behavior.

aleclarson:
@ashmit-coder 👍

brianc:
@aleclarson I let CI run on this PR so it should letcha know if tests pass. 😄

brianc:
BETTER LATE THAN NEVER! 🤦

PR #2813, opened 2022-09-13
Update package.json

matthieusieben:
Installing `pg-query-stream` using `yarn@v3` causes the following error:

```
/workspace/.pnp.cjs:42329
      Error.captureStackTrace(firstError);
            ^

Error: pg-cursor tried to access pg (a peer dependency) but it isn't provided by its ancestors; this makes the require call ambiguous and unsound.

Required package: pg (via "pg/lib/result.js")
Required by: pg-cursor@virtual:b15c8e58b2b8594a0187b88c869a498d86e8516db99b3d1eb03c094d7c7a1382793a82b929ca663d905d880350406f40122bf7227483202673f134928631ae7c#npm:2.7.4 (via /workspace/.yarn/__virtual__/pg-cursor-virtual-9bc4494085/0/cache/pg-cursor-npm-2.7.4-599c8a9a52-f150287f88.zip/node_modules/pg-cursor/)

Ancestor breaking the chain: pg-query-stream@npm:4.2.4


Require stack:
- /workspace/.yarn/__virtual__/pg-cursor-virtual-9bc4494085/0/cache/pg-cursor-npm-2.7.4-599c8a9a52-f150287f88.zip/node_modules/pg-cursor/index.js
- /workspace/.yarn/cache/pg-query-stream-npm-4.2.4-b15c8e58b2-635fdf986a.zip/node_modules/pg-query-stream/dist/index.js
- /workspace/index.js
    at Function.require$$0.Module._resolveFilename (/workspace/.pnp.cjs:42329:13)
    at Function.require$$0.Module._load (/workspace/.pnp.cjs:42183:42)
    at Module.require (node:internal/modules/cjs/loader:1028:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (/workspace/.yarn/__virtual__/pg-cursor-virtual-9bc4494085/0/cache/pg-cursor-npm-2.7.4-599c8a9a52-f150287f88.zip/node_modules/pg-cursor/index.js:2:16)
    at Module._compile (node:internal/modules/cjs/loader:1126:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1180:10)
    at Object.require$$0.Module._extensions..js (/workspace/.pnp.cjs:42373:33)
    at Module.load (node:internal/modules/cjs/loader:1004:32)
    at Function.require$$0.Module._load (/workspace/.pnp.cjs:42213:14)

```

This change makes it explicit that `pg-query-stream` has a (peer) dependency on `pg`.

### DIFF
--- a/packages/pg-query-stream/package.json
+++ b/packages/pg-query-stream/package.json
@@ -42,6 +42,9 @@
     "ts-node": "^8.5.4",
     "typescript": "^4.0.3"
   },
+  "peerDependencies": {
+    "pg": "^8"
+  },
   "dependencies": {
     "pg-cursor": "^2.7.4"
   }


charmander:
pg is only used by pg-query-stream for types, so it should probably be an `import type`, too.

matthieusieben:
The problem here is that `pg-cursor` [requires](https://github.com/brianc/node-postgres/blob/master/packages/pg-cursor/index.js#L2-L3) `pg` but does not list it as a `dependency` (it is listed as a `peerDependency`).

For some reason, it seems that yarn's PnP mode requires that `peerDependencies` of a module's dependencies be marked as `peerDependencies` in that module as well.

matthieusieben:
Note that the PnP error can be worked around using the following in `.yarnrc.yml`:

```yml
packageExtensions:
  # https://github.com/brianc/node-postgres/pull/2813
  pg-query-stream@*:
    peerDependencies:
      pg: '*'
```

brianc:
I agree with @charmander doing a `require type { Submittable }` should produce a bundle that's consumable without any dependency on the core library....as `require type` statements are effectively "erased" at compile time....would that be an effective fix?

matthieusieben:
Again, the problem is *not* the import in `pg-query-stream` (though it should certainly be changed to an `import type`, and a devDependency).

The problem here is that `yarn`, when running in PnP mode, will check that anything you `require/import` is indeed declared in the `dependencies` or `peerDependencies` of the module (`package.json`).

Here, we have the following `pg-query-stream` dependency tree:
  - dependency: `pg-cursor`
    - peerDependency: `pg`

Because `pg-query-stream` does not declare `pg` as either a dependency or a peer dependency, yarn's PnP script will yield an error when importing `pg-query-stream`. This might be an error in yarn's PnP which should not behave that way regarding peer dependencies of child modules.

I did report the "bug" here because:
1) from a conceptual point of view, `pg-query-stream` *does* have a peer dependency on `pg`
2) from the 2215 other modules in my `node_modules`, `pg-query-stream` is the only one causing an error when yarn is in PnP mode.

charmander (approved):

matthieusieben:
❤️

PR #2815, opened 2022-09-16
use crypto.pbkdf2Sync in sasl auth

### LINKED ISSUE #2814: cache password hash in SASL auth
In SASL auth handling, `Hi` function is [called](https://github.com/brianc/node-postgres/blob/3e53d06cd891797469ebdd2f8a669183ba6224f6/packages/pg/lib/sasl.js#L40) for every established connection although the password and hash are same. In my measurements, it takes 10-20ms to calculate the hash with default 4096 iterations. Caching the result of this function would significantly improve performance. 
What do you think? I can prepare PR.

matej21:
fixes #2814

### DIFF
--- a/packages/pg/lib/sasl.js
+++ b/packages/pg/lib/sasl.js
@@ -37,7 +37,7 @@ function continueSession(session, password, serverData) {
 
   var saltBytes = Buffer.from(sv.salt, 'base64')
 
-  var saltedPassword = Hi(password, saltBytes, sv.iteration)
+  var saltedPassword = crypto.pbkdf2Sync(password, saltBytes, sv.iteration, 32, 'sha256')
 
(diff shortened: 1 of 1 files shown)


charmander (approved):

brianc (approved):
This is great - thanks for diving in here!  Really appreciate the community attention on the parts of the library my venture doesn't use on a day to day basis.  I need to make a note to myself to eventually convert this function to async as doing anything with `*Sync` in the name is probably not as good long term as using the async version....though connections to postgres are already rather slow (due to the protocol and handshake itself) and not extremely common (and the previous code was blocking) so I'm not immediately highly concerned.

matej21:
@brianc hi, when do you plan to release a new version of the `pg` package?

PR #2831, opened 2022-10-05
pg-cursor: Fix errors only being sent to half the queue

charmander:
`pop` decreases the length while `i` is still being checked against it.


charmander:
semver-patch?

PR #2832, opened 2022-10-05
pg-pool: don't allow connection reuse after streaming

alxndrsn:
This might be a fix for https://github.com/brianc/node-postgres/issues/1674.

I would expect that connections _could_ be re-used safely after streaming though, so there should be a more correct way of approaching this...


charmander (changes requested):
This isn’t a good hack. I don’t think it should be merged.

alxndrsn:
Closing in favour of https://github.com/brianc/node-postgres/pull/2836.

PR #2833, opened 2022-10-05
pg-connection-string: avoid clobbering port from queryparams

rafiss:
If the connection string is something like:
    postgresql://demo:password@/postgres?host=localhost&port=26258

Then the port from the query parameters should be used. Previously, the parsing function would end up with a null port, and the default port would end up being used by the connection package.

### DIFF
--- a/packages/pg-connection-string/index.js
+++ b/packages/pg-connection-string/index.js
@@ -31,7 +31,6 @@ function parse(str) {
   config.user = auth[0]
   config.password = auth.splice(1).join(':')
 
-  config.port = result.port
   if (result.protocol == 'socket:') {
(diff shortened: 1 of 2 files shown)


rafiss:
Hi @brianc! Any chance this can get reviewed? I believe I've followed the [guidelines](https://github.com/brianc/node-postgres#contributing) but let me know if I should do something else to move this forward.

charmander (approved):

rafiss:
@charmander thanks for the review! is anything else needed to merge this?

rafiss:
Hi @brianc or @charmander - just checking in to see if this can get merged, since it has been approved already.

PR #2861, opened 2022-11-14
Make Result iterable

vitaly-t:
Make `Result` automatically iterable, so one can iterate through data without having to access `rows` explicitly:

```js
client.query('select...', (err, res) => {
    // below we can use just "res", not "res.rows"
    for(const a of res) {
        console.log(a); // print the row
    }
})
```

### DIFF
--- a/packages/pg/lib/result.js
+++ b/packages/pg/lib/result.js
@@ -23,6 +23,10 @@ class Result {
     }
   }
 
+  [Symbol.iterator]() {
+    return this.rows.values()
+  }
+  
   // adds a command complete message
   addCommandComplete(msg) {
     var match


vitaly-t:
In an afterthought, it would make the library perform better, if `rows` array is retired, and replaced with an iterator, because it doesn't make sense executing all that data-type conversion unless it is actually used. In many cases, particularly with large data results, the client ends up not ever using the data, rather a small portion of it.

vitaly-t:
[Added the same change into pg-native, to be merged at the same time.](https://github.com/brianc/node-pg-native/pull/113).

charmander:
I don’t think this is worth changing now.

- Multi-statement queries return an array of results that is also iterable, which could be confusing.
- `res.rows`, `{rows}`, or a wrapper function are easy to write.
- People will often want to use array methods, which aren’t available on the plain iterable anyway.

It’s nice when there’s just one way to do it.

vitaly-t:
This caters for all coding styles, without any change in functionality. It also opens a path to changing types conversion, to become dynamic. There are lots of libraries these days that handle iterables well.

PR #2879, opened 2022-12-07
Changed docs header title to projects name

alejandroperezpujante:
Noticed that the title in the docs displays a generic text from Nextra.

Tought it would be nice to change it to the projects name.


brianc:
nice!

PR #2898, opened 2023-01-14
Add support to stream factory

ruyadorno:
This changeset enables declaring the `stream` config value as a factory method. Providing a much more flexible control of the socket connection.

Defining a custom `stream` config value allows the postgres driver to support a larger variety of environments/setups such as proxy servers and secure socket connections that are used by cloud providers such as GCP.

Currently, usage of the `stream` config value is only viable for single connections given that it's only possible to define a single socket stream instance per new Client/Pool instance. By adding support to a factory function, it becomes possible to enable usage of custom socket streams for connection pools.

For reference, see the **mysql2** driver for MySQL (linked below) for prior art example of this pattern.

**Refs:** https://github.com/sidorares/node-mysql2/blob/ba15fe25703665e516ab0a23af8d828d1473b8c3/lib/connection.js#L63-L65
**Refs:** https://cloud.google.com/sql/docs/postgres/connect-overview
**Signed-off-by:** Ruy Adorno <ruyadorno@google.com>

### DIFF
--- a/packages/pg/lib/connection.js
+++ b/packages/pg/lib/connection.js
@@ -14,7 +14,12 @@ class Connection extends EventEmitter {
   constructor(config) {
     super()
     config = config || {}
+
     this.stream = config.stream || new net.Socket()
+    if (typeof this.stream === 'function') {
+      this.stream = this.stream(config)
+    }
+
     this._keepAlive = config.keepAlive
     this._keepAliveInitialDelayMillis = config.keepAliveInitialDelayMillis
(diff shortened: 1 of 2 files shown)


brianc:
Hey sorry for missing this! Been on the road traveling for my dad's 70th birthday.  I'll get to this & a couple other PRs merged up here & do a release with these cool features & fixes.

ruyadorno:
awesome to hear @brianc! thank you so much 😄

PR #2899, opened 2023-01-19
Add informative SASL empty password error message

### LINKED ISSUE #2757: Error: SASL: SCRAM-SERVER-FIRST-MESSAGE: client password must be a string

sehrope:
First commit adds a test for non-string password values being used for SASL authentication. The main code already throws an error for it but it wasn't tested anywhere.

Second commit adds a new error message when SASL authentication is requested by the server but the password is a zero-length string. This would already be a failure situation as you cannot have a zero-length password with SASL auth. This commit just changes it so the user gets an error of: `SASL: SCRAM-SERVER-FIRST-MESSAGE: client password must be a non-empty string`. Also adds a unit test for it.

I didn't merge the two checks and error messages as they're more informative for the end user separately.

Addresses and somewhat closes #2757.

### DIFF
--- a/packages/pg/lib/sasl.js
+++ b/packages/pg/lib/sasl.js
@@ -23,6 +23,9 @@ function continueSession(session, password, serverData) {
   if (typeof password !== 'string') {
     throw new Error('SASL: SCRAM-SERVER-FIRST-MESSAGE: client password must be a string')
   }
+  if (password === '') {
+    throw new Error('SASL: SCRAM-SERVER-FIRST-MESSAGE: client password must be a non-empty string')
+  }
   if (typeof serverData !== 'string') {
     throw new Error('SASL: SCRAM-SERVER-FIRST-MESSAGE: serverData must be a string')
   }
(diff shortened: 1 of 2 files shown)


packages/pg/lib/sasl.js
```diff
@@ -23,6 +23,9 @@ function continueSession(session, password, serverData) {
   if (typeof password !== 'string') {
     throw new Error('SASL: SCRAM-SERVER-FIRST-MESSAGE: client password must be a string')
   }
+  if (password.length === 0) {
```
charmander, line 26: More typical?
```suggestion
  if (password === '') {
```

sehrope: Yes that's better. Fixed and pushed.

charmander:
Looks good, but I don’t understand how it can help with #2757. Anyone who’s hitting that error message won’t hit this one, right?

sehrope:
@charmander As usual, you're right :D

Studying this a bit more I realized the SASL integration tests weren't even being run in CI. I've created a new PR, #2901, that fixes that and all this too. Closing this out in favor of that newer PR.

Issue #1927, opened 2019-07-19
No handlers for SASL errors


sehrope:
Leaving a note here as a reminder to take another look at the SASL error handling. At first glance I don't think the errors would bubble up as they're thrown synchronously in the helper functions and the caller does not catch or emit them up to the client.

https://github.com/brianc/node-postgres/blob/0acaf9d8ff3e96fd2a71c786c8ee883c2b2006d2/lib/sasl.js#L4-L17

https://github.com/brianc/node-postgres/blob/0acaf9d8ff3e96fd2a71c786c8ee883c2b2006d2/lib/client.js#L142-L146

gabegorelick:
> At first glance I don't think the errors would bubble up as they're thrown synchronously in the helper functions and the caller does not catch or emit them up to the client.

This is still an issue as of pg 8.7.1. 

To reproduce, attempt to connect to a PG instance that uses scram and don't specify a password when connecting:
```js
const client = new Client();
try {
  await client.connect();
} catch (err) {
  // this is never reached
}
```

You'll get an uncaught exception: `Error: SASL: SCRAM-SERVER-FIRST-MESSAGE: client password must be a string`. But this really should be raised by `client.connect` and not be unhandled.

rpominov:
I’m building a CLI app that uses `node-postgres`.

When I leave all connection arguments empty as a user of the app, it crashes with this very confusing error.

As a developer of the app, I don’t see any way to catch this error to provide any context to the user as to why it crashed and exit gracefully.

If it’s possible to prioritise fixing this, I would appreciate it greatly.

charmander:
@rpominov Check if it’s missing before making a connection attempt.

rpominov:
I think it should be possible to setup the server so that you can connect without a password.

Also checking the password is not trivial, as it can come from at least 4 places that I'm aware of: 
 - password option, 
 - connectionString option, 
 - env variable,
 - .pgpass file.

In general, there seem to be a lot of authentication options that node-postgres supports that I don't fully understand. Ideally, I want to just pass the options to node-postgres and show any errors that it produces back to the user, while stating in the documentation that they can use whatever authentication method that the library supports.

ephys:
We're encountering an issue over at Sequelize after updating our CI to use postgres 14 (PR https://github.com/sequelize/sequelize/pull/14737)

[...]

PR #2901, opened 2023-01-23
Fix SASL to bubble up errors, enable SASL tests in CI, and add informative empty SASL password message

sehrope:
Couple of SASL related commits here:

1. Enables SASL testing in GitHub actions CI.
2. Adds a basic unit test for SASL password being a string.
3. Catches SASL handling errors and bubbles them up as error events.
4. Adds an explicit check to the SASL handling for an empty password so the user gets an informative email.

Number 3 should actually close out https://github.com/brianc/node-postgres/issues/2757. Interestingly this was brought up in 2019 when the SASL code was first added: https://github.com/brianc/node-postgres/issues/1927. This should close that out too.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,7 @@ jobs:
           POSTGRES_USER: postgres
           POSTGRES_PASSWORD: postgres
           POSTGRES_DB: ci_db_test
+          POSTGRES_HOST_AUTH_METHOD: 'md5'
         ports:
           - 5432:5432
         options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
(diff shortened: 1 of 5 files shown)


brianc:
aww thank you, this is an awesome change! 💃 Will get this merged & pushed out ASAP!

verycosy:
AWESOME!

Prasad1-S:
I hit this exact error today when my DB_PASSWORD env var was undefined. The current message `client password must be a string` sent me in the wrong direction — I was looking at my login logic instead of my DB connection config.

The empty string improvement in this PR is great, but the `undefined` case still has the same unhelpful message. Would a PR that improves it to something like `client password must be a string (received undefined — is your DB_PASSWORD environment variable set?)` be welcome?

PR #2908, opened 2023-02-04
Destroy connection if the server does not support SSL

### LINKED ISSUE #2907: Connection failure leaks an open socket
When trying to connect to a Postgres instance with SSL enabled, but the server didn't enable SSL, we get a "The server does not support SSL connections" as expected. But it does not close the underlying socket and it is leaked, preventing the node process from closing. It will hang for a few seconds/milliseconds until the next GC cycle runs to collect and close the socket.

```js
import pg from "pg";

const client = new pg.Client({
  connectionString: process.env.CONNECTION_STRING,
  ssl: { rejectUnauthorized: false },
});

try {
  await client.connect();
} catch (e) {
  console.log(e);
} finally {
  await client.end();
}

console.log("done");
```

This snippet should run and exit immediatelly after showing "done" on the console.

This issue relates to #2720. There are situations where we don't know ahead of time if the database supports SSL or not, and we must try connecting with and without ssl. This is an issue for short-lived applications, such as a migration runner.

lbguilherme:
Fix #2907 by removing the delay of ending the connection normally, as this is an error condition.

### DIFF
--- a/packages/pg/lib/connection.js
+++ b/packages/pg/lib/connection.js
@@ -72,11 +72,11 @@ class Connection extends EventEmitter {
         case 'S': // Server supports SSL connections, continue with a secure connection
           break
         case 'N': // Server does not support SSL connections
-          self.stream.end()
+          self.stream.destroy()
           return self.emit('error', new Error('The server does not support SSL connections'))
         default:
           // Any other response byte, including 'E' (ErrorResponse) indicating a server error
-          self.stream.end()
+          self.stream.destroy()
           return self.emit('error', new Error('There was an error establishing an SSL connection'))
       }
       var tls = require('tls')


brianc:
Thanks for this PR!  I really appreciate it.  Would it be possible to include a test? With github issues like #2907 sometimes I just make a file [here](https://github.com/brianc/node-postgres/tree/master/packages/pg/test/integration/gh-issues) with the issue number and you can basically copy/paste the code that's supposed to cleanly exit into the file...if the file never exits the test is considered a failure so it'd be a quick way to create a test for this in the future.  If you don't want to run the full test suite locally you can run any individual file in the suite with `node packages/pg/test/integration/gh-issues/2907-tests.js` for example.  pg was written before any of the now popular node test frameworks existed so the tests are basically each file is a self-contained script which must exit with `0` or the test suite is a fail, so its easy to just run one test (or file full of tests) at a time.

PR #2913, opened 2023-02-21
Fix typo in URL on Pooling page

zegerius:
The [features/pooling page](https://node-postgres.com/features/pooling#good-news) links to `https://node-postgres.com/api/pool` (404) which I believe should be `https://node-postgres.com/apis/pool`.

So here a small PR to fix that!


brianc:
Thanks!

PR #2845, opened 2022-10-17
Emit a 'release' event when a connection is released back to the pool

nihonjinrxs:
This PR has the `Pool` object emit a `'release'` event upon the release of a client/connection back to the pool.

We're experiencing an issue that feels quite similar to what's described in brianc/node-postgres#2262, and in reading through the comments there, I was interested to see @brianc's suggestion of [adding a `'release'` event as the corresponding bookend to the `'connect'` and `'acquire'` events](https://github.com/brianc/node-postgres/issues/2262#issuecomment-656751203). However, in searching the code, it seems this was never implemented, so I've implemented that here.

PR #2928, opened 2023-03-10
Add release event to Pool API docs

nihonjinrxs:
@brianc I realized with your release including #2845 and the update of the docs link in #2925 that I hadn't done the documentation update to match the feature change for the new event in pg-pool.

This small PR makes that documentation update.

Apologies for the prior oversight!


brianc:
hey no worries at all! Really appreciate it!! 😄

PR #1448, opened 2017-09-07
Support Uint8Array values

jscissr:
Fixes #1166

PR #2930, opened 2023-03-13
Serialize arrays of Uint8Array objects as hex escape sequences

sds:
Previously, if you attempted to pass an array of `Uint8Array` objects to a prepared statement, it would render each literal numeric value of that array.

Since `Uint8Array` (and other `TypedArray` types) represent views over raw bytes, ensure these are serialized to Postgres as bytes via hex escape sequence.

### DIFF
--- a/packages/pg/lib/utils.js
+++ b/packages/pg/lib/utils.js
@@ -23,8 +23,17 @@ function arrayString(val) {
       result = result + 'NULL'
     } else if (Array.isArray(val[i])) {
       result = result + arrayString(val[i])
-    } else if (val[i] instanceof Buffer) {
-      result += '\\\\x' + val[i].toString('hex')
+    } else if (ArrayBuffer.isView(val[i])) {
+      var item = val[i]
+      if (!(item instanceof Buffer)) {
+        var buf = Buffer.from(item.buffer, item.byteOffset, item.byteLength)
+        if (buf.length === item.byteLength) {
+          item = buf
+        } else {
+          item = buf.slice(item.byteOffset, item.byteOffset + item.byteLength)
+        }
+      }
+      result += '\\\\x' + item.toString('hex')
     } else {
       result += escapeElement(prepareValue(val[i]))
     }
--- a/packages/pg/test/unit/utils-tests.js
+++ b/packages/pg/test/unit/utils-tests.js
@@ -175,6 +175,13 @@ test('prepareValue: buffer array prepared properly', function () {
   assert.strictEqual(out, '{\\\\xdead,\\\\xbeef}')
 })
 
+test('prepareValue: Uint8Array array prepared properly', function () {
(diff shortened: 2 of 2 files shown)


packages/pg/lib/utils.js
```diff
@@ -23,8 +23,17 @@ function arrayString(val) {
       result = result + 'NULL'
     } else if (Array.isArray(val[i])) {
       result = result + arrayString(val[i])
-    } else if (val[i] instanceof Buffer) {
-      result += '\\\\x' + val[i].toString('hex')
+    } else if (ArrayBuffer.isView(val[i])) {
+      var item = val[i]
+      if (!(item instanceof Buffer)) {
```
charmander, line 28:
```suggestion
      if (!Buffer.isBuffer(item)) {
```

packages/pg/lib/utils.js
```diff
@@ -23,8 +23,17 @@ function arrayString(val) {
...
+        var buf = Buffer.from(item.buffer, item.byteOffset, item.byteLength)
+        if (buf.length === item.byteLength) {
```
charmander, line 30: Is this for Node 4.x compatibility? A comment might be useful.

charmander: Actually, pg requires Node 8.x as a minimum now.

sds: Good to know! This was copy/pasted from below, which specifically calls out the Node.js V4 requirement.

I didn't want to diverge from established practice. If you're saying it's safe to drop code supporting Node.js V4, perhaps that should be done separately?

packages/pg/lib/utils.js
sds: Was trying to follow the style used elsewhere in this file (see above). Happy to change if we're otherwise supportive of this direction.

sds:
Thanks for the review, @charmander. 

Would you like me to combine your recommendations into this PR, or would you prefer those changes be made in separate PRs? Happy to make the additional changes, just looking for confirmation. Thanks!

sds:
Polite ping, @charmander. Let me know if we'd like your requested changes combined into a single PR, or implemented in separate PRs. Happy to do so once I get clarity on preference. Thanks!

SevInf:
We've just stumbled upon this bug in Prisma too and would like to see this merged. @charmander is there any way we could help? If there are some changes need we can spend some time on it.

charmander:
(Is it really a bug if typed arrays didn’t exist in Node back when the code was written?)

Anyway, I’ll merge this for convenience, but it’s up to @brianc to release it. (semver minor)

SevInf:
>(Is it really a bug if typed arrays didn’t exist in Node back when the code was written?)

Singular `typed is correctly encoded by `pg` since [2017](https://github.com/brianc/node-postgres/pull/1448), only a list of typed arrays is not until this PR.

PR #2941, opened 2023-03-29
feat: add connection parameter nativeConnectionString

NovikovEvgeny:
This is a follow-up of MR https://github.com/brianc/node-postgres/pull/2751

I copy-paster the solution itself, but also I have added tests as you asked

Support of nativeConnectionString in some cases could be really useful like
* specifying multihost URI
* specifying additional params (like target_session_attrs)
* having different formats of the connection URI, anything what is described in the [libpq docs](https://www.postgresql.org/docs/9.3/libpq-connect.html#LIBPQ-CONNSTRING)

using ConnectionParameters class not always be helpful since it doesn't support all possible variants of the connection string

### DIFF
--- a/packages/pg/lib/native/client.js
+++ b/packages/pg/lib/native/client.js
@@ -29,6 +29,7 @@ var Client = (module.exports = function (config) {
   // keep these on the object for legacy reasons
   // for the time being. TODO: deprecate all this jazz
   var cp = (this.connectionParameters = new ConnectionParameters(config))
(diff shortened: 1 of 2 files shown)


alexkvak:
@brianc here is the same as https://github.com/brianc/node-postgres/pull/2751 but with tests. Could you review and merge it?

brianc:
Nice! Thanks - sorry for the delay. Will release this today. 😄

PR #2942, opened 2023-03-29
Implement Batch query package

### LINKED ISSUE #2257: Extended Query: Support Batch Execution
Hi!

The [Extended Query](https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-EXT-QUERY) protocol enables drivers to submit multiple BIND messages before SYNC. One of the big benefits of using Extended Queries is that you can efficiently send a batch of executions without waiting for round trips for each execution. Pairing that with prepared statements and some simplifications: you send a single PARSE, a bunch of BIND/EXECUTE and a SYNC to find out how things went.

In other words, you'd be able to support something like the following without needing 4 entire round trips. (I'm not _recommending_ this API since it would be a terrible breaking change.)

```javascript
await client.query({
  name: 'my_query',
  text: 'insert sometable (id, val) values ($1, $2)'
  values: [
    [ 1, 'asdf' ],
    [ 2, 'fdsa' ],
    [ 3, 'qwer' ],
    [ 4, 'uytr' ]
  ]
})
```

For more information, check out how the JDBC Postgres driver handles a batched execution. There are a few layers to dig through, but [this appears to be the core of the code that sends a batch of messages](https://github.com/pgjdbc/pgjdbc/blob/b80ebcef33f9c3801f01e28994746bc36abb2d21/pgjdbc/src/main/java/org/postgresql/core/v3/QueryExecutorImpl.java#L490-L509) and subsequently sends a single SYNC. NOTE: their driver imposes a limit of 128 records per batch as ([apparently](https://github.com/pgjdbc/pgjdbc/blob/b80ebcef33f9c3801f01e28994746bc36abb2d21/pgjdbc/src/main/java/org/postgresql/jdbc/PgPreparedStatement.java#L1604-L1608)) further batching does not improve performance.

iamkhush:
Attempts to fix #2257

### DIFF
--- a/packages/pg-batch-query/README.md
+++ b/packages/pg-batch-query/README.md
@@ -1,0 +1,65 @@
+# pg-batch-query
+
+
+
+## installation
+
+```bash
+$ npm install pg --save
+$ npm install pg-batch-query --save
+```
+
+## use
+
+```js
+const pg = require('pg')
+var pool = new pg.Pool()
+const BatchQuery = require('pg-batch-query')
+
+const batch = new BatchQuery({
+  name: 'optional',
+  text: 'INSERT INTO foo (bar) VALUES ($1)',
+  values: [
+    ['first'],
+    ['second']
+  ]
+})
+
+pool.connect((err, client, done) => {
+  if (err) throw err
+  const result = client.query(batch).execute()
+  for (const res of result) {
+    for (const row of res) {
+      consolel.log(row)
+    }
+  }
+})
+```
+
+## contribution
+
+I'm very open to contribution! Open a pull request with your code or idea and we'll talk about it. If it's not way insane we'll merge it in too: isn't open source awesome?
+
+## license
+
+The MIT License (MIT)
+
+Copyright (c) 2013-2020 Ankush Chadda
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
--- a/packages/pg-batch-query/README.md
+++ b/packages/pg-batch-query/README.md
@@ -18,7 +18,7 @@ const BatchQuery = require('pg-batch-query')
 
 const batch = new BatchQuery({
   name: 'optional',
-  text: 'INSERT INTO foo (bar) VALUES ($1)',
+  text: 'SELECT from foo where bar = $1',
   values: [
     ['first'],
     ['second']
--- a/packages/pg-batch-query/bench.ts
+++ b/packages/pg-batch-query/bench.ts
@@ -1,0 +1,79 @@
+import pg from 'pg'
+import BatchQuery from './dist'
+
+const insert = (value) => ({
+  text: 'INSERT INTO foobar(name, age) VALUES ($1, $2)',
+  values: ['brian', value],
+})
+
+const seq = {
+  text: 'SELECT * FROM generate_series(1, 1000)',
+}
+
+let counter = 0
+
+const simpleExec = async (client, getQuery, count) => {
+  const query = getQuery(count)
+  await client.query({
+    text: query.text,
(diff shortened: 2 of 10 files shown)


packages/pg-batch-query/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+## contribution
+
+I'm very open to contribution! Open a pull request with your code or idea and we'll talk about it. If it's not way insane we'll merge it in too: isn't open source awesome?
+
+## license
+
+The MIT License (MIT)
+
+Copyright (c) 2013-2020 Ankush Chadda
```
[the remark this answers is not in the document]
brianc: yea absolutely okay. you can get boilerplate MIT license text from github or other places 👍

brianc:
whoah this is really awesome! I'm excited to help get this over the finishline & release it.

brianc:
@charmander any thoughts on this? I think it looks p great!

packages/pg-batch-query/bench.ts
```diff
@@ -0,0 +1,79 @@
+import pg from 'pg'
+import BatchQuery from './dist'
```
brianc, line 2: question: is it a bit strange to require the dist folder here? Why not require from `./src`?

packages/pg-batch-query/src/index.ts
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  execute(): Promise<QueryResult[]> {
+    return new Promise((resolve, reject) => {
+      this.callback = (err, rows) => (err ? reject(err) : resolve(rows))
+    })
+  }
+
+  handleError(err, connection) {
+    console.log(err)
```
brianc, line 85: This should be removed.

packages/pg-batch-query/src/index.ts
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  handleReadyForQuery(con) {
+    if (this.callback) {
+      try {
+        this.callback(null, this._results)
+      }
+      catch(err) {
+        process.nextTick(() => {
+          throw err
```
brianc, line 96: why catcy & throw on next tick? Doesn't this create an uncatchable error which will crash a node process?

packages/pg-batch-query/test/test-query.ts
```diff
@@ -0,0 +1,43 @@
...
+            ['first'],
+            ['second']
+        ],
+        name: 'optional'
+    })).execute()
+    for ( const response of responses) {
+      assert.strictEqual(response.rowCount, 1)
+    }
+  })
+})
```
brianc, line 43: would be good to see some tests for errors / error handling / error recovery

brianc (commented):
Some initial comments on things we should change.  And needs some tests around errors, error handling, and ensuring the client is still usable after an error (you typically need to flush after an error I believe? Could be wrong there...tests will uncover these things though 😄 )

packages/pg-batch-query/bench.ts
iamkhush: Good catch. It seems IDE auto-imported it from `dist` . Will fix.

charmander:
> @charmander any thoughts on this? I think it looks p great!

If it’s going to be a separate package instead of an API change to pg’s default query type – why in this repo?

packages/pg-batch-query/test/test-query.ts
iamkhush: added `test-error-handling.ts`

iamkhush:
@brianc @charmander please have a look at the PR and provide feedback

abenhamdine:
waouh, so happy to see that !

packages/pg-batch-query/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "devDependencies": {
+    "@types/chai": "^4.2.13",
+    "@types/mocha": "^8.0.3",
+    "@types/node": "^14.0.0",
+    "@types/pg": "^7.14.5",
+    "eslint-plugin-promise": "^6.0.1",
+    "mocha": "^7.1.2",
+    "pg": "^8.10.0",
+    "ts-node": "^8.5.4",
+    "typescript": "^4.0.3"
```
abenhamdine, line 40: typescript is now 5.0.4, why keep ts 4 ?

packages/pg-batch-query/README.md
abenhamdine, line 1: IMHO, it would be great to add explanations about : 
- the benefits of batched queries vs  non-batched
- the difference with pipelining (not implemented yet)
- perhaps also some hints on the underlying mechanism (multiple BIND etc) for future contributors

I also wonder if any advice about the number of parameters is needed ? is there a hard limit ?

packages/pg-batch-query/package.json
```diff
@@ -0,0 +1,45 @@
...
+  },
+  "devDependencies": {
+    "@types/chai": "^4.2.13",
+    "@types/mocha": "^8.0.3",
+    "@types/node": "^14.0.0",
+    "@types/pg": "^7.14.5",
+    "eslint-plugin-promise": "^6.0.1",
+    "mocha": "^7.1.2",
+    "pg": "^8.10.0",
+    "ts-node": "^8.5.4",
```
abenhamdine, line 39: ts-node is now version 10

abenhamdine:
I think this important question remains unanswered 

> If it’s going to be a separate package instead of an API change to pg’s default query type – why in this repo

in other words : are maintainers of this repo willing to take responsability for this new package ?

iamkhush:
>In other words : are maintainers of this repo willing to take responsability for this new package ?

@abenhamdine I have followed what @brianc advised me [here](https://github.com/brianc/node-postgres/issues/2257#issuecomment-1230860536).
But ofcourse it will be good to have confirmation again

iamkhush:
Moving it to draft to add typescript definitions

iamkhush:
Hey @hillac , happy to continue the work if maintainers would confirm on what was [asked above](https://github.com/brianc/node-postgres/pull/2942#issuecomment-1557806854)

iamkhush:
@hillac could you show the code where it didnt work for you ? Im curious since the bench.ts file showed triple the qps I had seen otherwise.

PR #2947, opened 2023-04-01
Fix: deal with bigint types in objects stored as jsonb

ales-albert-kilbergr:
The pg library fails to serialize query values in case when an object for jsonb column contains bigint. 

The failed error message is as follows because JSON.serialize does not know how to deal with bigint.

```
TypeError: Do not know how to serialize a BigInt
```

Reproducible example could look like this:
```ts
const queryConfig = `INSERT INTO my_table (id, data) VALUES ($1, $2)`;
const values = ['someId', { someValue: BigInt(1) }]
await pgClient.query(queryConfig, values);
```

### DIFF
--- a/packages/pg/lib/utils.js
+++ b/packages/pg/lib/utils.js
@@ -78,7 +78,7 @@ function prepareObject(val, seen) {
 
     return prepareValue(val.toPostgres(prepareValue), seen)
   }
-  return JSON.stringify(val)
+  return JSON.stringify(val, (_, val) => typeof val === 'bigint' ? val.toString() : val)
 }
 
 function pad(number, digits) {


charmander:
This introduces overhead to every JSON encoding in order to accommodate one specific type. I would suggest moving the `JSON.stringify` step to user code when you need to stringify a `BigInt`; I don’t think it belongs in pg.

Also, I wouldn’t expect `BigInt` to be serialized as a JSON *string*.

ales-albert-kilbergr:
> Also, I wouldn’t expect BigInt to be serialized as a JSON string.

A big int can exceed a range for number in javascript therefor there is not much other options how to serialize it in JSON than a string currently.

> I would suggest moving the JSON.stringify step to user code when you need to stringify a BigInt

That would be a good option.

>  I don’t think it belongs in pg.

Regarding a fact, that it most correctly represents BIGSERIAL or BIGINT in postgres, I would expect the pg library to deal with it without  any user  custom code. It is just one standardized javascript primitive type the same way as boolean or number.

charmander:
> BigInt is standartized javascript primitive type. Why shouldn`t be possible to pass it as a property for JSONB object?

For the reasons I mentioned. To expand on the second one: it doesn’t survive the round trip through `JSON.serialize`. To emphasize the first: this affects all JSON serialization and is unrelated to pg, so why should pg have its own special behaviour?

PR #2952, opened 2023-04-06
fix(theme.config.js): Replace default meta description and social title

janpio:
Currently still nextra default.

Those are shown in Slack and other social apps when sharing the website.
image


brianc (approved):
beautiful, i love it. thank you!!

Issue #1978, opened 2019-09-19
Document `escapeLiteral` and `escapeIdentifier`


jmm:
Hello, thanks for the work on this.

I'm new to this library and was surprised not to be finding any escaping functions. But I managed to find https://github.com/brianc/node-postgres/pull/396.

It seems these are part of the public API and just seemingly aren't documented, so if that's correct this is a request to document them.

Also, since `Pool` otherwise appears to be a superset of the `Client` API, would it make sense for them to be available there as well?

Thanks

brianc:
The docs are open source and open to contribution, would love to help you add docs about this over there! https://github.com/brianc/node-postgres-docs

As far as adding it to pool as well? Nah, I thinking having only 1 place to access a function is probably overall easier to reason about than putting it on the pool _as well_ though I'll admit having those functions be on the client _prototype_ was a weird decision made long ago.  Probably should just be exported as top level funcs or something nowadays...but anyways it is what it is.

One thing to note about escaping parameters - you're better off using parameterized SQL.  I realize that's not _always_ an option.

PR #2954, opened 2023-04-10
Document client.escapeIdentifier and client.escapeLiteral

TheConner:
Per #1978 it seems that these client APIs are undocumented. I added documentation for these methods along with some examples and relevant links.

I wasn't too sure what kind of structure to take with the documentation, such as documenting the usage via a client instance and the Client prototype. I can shuffle things around if there is a better way of expressing the usage.

### DIFF
--- a/docs/pages/apis/client.mdx
+++ b/docs/pages/apis/client.mdx
@@ -1,6 +1,7 @@
 ---
 title: pg.Client
 ---
+import { Alert } from '/components/alert.tsx'
 
 ## new Client
 
@@ -255,6 +256,44 @@ client
 ```
 
 _note: end returning a promise is only available in pg7.0 and above_
+
+## client.escapeIdentifier
+
+Escapes a string as a [SQL identifier](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS). There are two ways of using this method: 
+
+First, this can be used on an existing instance of `Client`:
+```js
+const escpaedIdentifier = client.escapeIdentifier('FooIdentifier')
+console.log(escpaedIdentifier) // '"FooIdentifier"'
+```
+
+Alternatively, this can be used via `Client.prototype`:
+```js
+const { Client } = require('pg')
+const escpaedIdentifier = Client.prototype.escapeIdentifier('Bar"Identifier')
+console.log(escpaedIdentifier) // '"Bar""Identifier"'
+```
+
+## client.escapeLiteral
+
+Escapes a string as a [SQL literal](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-CONSTANTS). There are two ways of using this method:
+
+<Alert>
+  **Note**: Instead of manually escaping SQL literals, it is recommended to use parameterized queries. Refer to [parameterized queries](/features/queries#parameterized-query) and the [client.query](#clientquery) API for more information.
+</Alert>
+
+First, this can be used on an existing instance of `Client`:
+```js
+const escapedLiteral = client.escapeLiteral("hello 'world'")
+console.log(escapedLiteral) // "Hello ''world''"
+```
+
+Alternatively, this can be used via `Client.prototype`:
+```js
+const { Client } = require('pg')
+const escapedLiteral = Client.prototype.escapeLiteral("hello \\ ' world")
+console.log(escapedLiteral) // " E'hello \\\\ '' world'"
+```
 
 ## events
 


docs/pages/apis/client.mdx
```diff
@@ -256,6 +257,44 @@ client
...
 _note: end returning a promise is only available in pg7.0 and above_
 
+## client.escapeIdentifier
+
+Escapes a string as a [SQL identifier](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS). There are two ways of using this method: 
+
+First, this can be used on an existing instance of `Client`:
+```js
+const escpaedIdentifier = client.escapeIdentifier('FooIdentifier')
+console.log(escpaedIdentifier) // '"FooIdentifier"'
```
charmander, line 267:
```suggestion
const escapedIdentifier = client.escapeIdentifier('FooIdentifier')
console.log(escapedIdentifier) // '"FooIdentifier"'
```

docs/pages/apis/client.mdx
```diff
@@ -256,6 +257,44 @@ client
...
+Escapes a string as a [SQL literal](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-CONSTANTS). There are two ways of using this method:
+
+<Alert>
+  **Note**: Instead of manually escaping SQL literals, it is recommended to use parameterized queries. Refer to [parameterized queries](/features/queries#parameterized-query) and the [client.query](#clientquery) API for more information.
+</Alert>
+
+First, this can be used on an existing instance of `Client`:
+```js
+const escapedLiteral = client.escapeLiteral("hello 'world'")
+console.log(escapedLiteral) // "Hello ''world''"
```
charmander, line 288:
```suggestion
console.log(escapedLiteral) // "'hello ''world'''"
```

docs/pages/apis/client.mdx
```diff
@@ -256,6 +257,44 @@ client
...
+```js
+const escpaedIdentifier = client.escapeIdentifier('FooIdentifier')
+console.log(escpaedIdentifier) // '"FooIdentifier"'
+```
+
+Alternatively, this can be used via `Client.prototype`:
+```js
+const { Client } = require('pg')
+const escpaedIdentifier = Client.prototype.escapeIdentifier('Bar"Identifier')
+console.log(escpaedIdentifier) // '"Bar""Identifier"'
```
charmander, line 274:
```suggestion
const escapedIdentifier = Client.prototype.escapeIdentifier('Bar"Identifier')
console.log(escapedIdentifier) // '"Bar""Identifier"'
```

charmander (changes requested):
I think we should export these as `pg.escapeIdentifier` and `pg.escapeLiteral` instead of documenting the `Client.prototype` workaround.

TheConner:
Fixed the typos, thank you for pointing them out! 

> I think we should export these as `pg.escapeIdentifier` and `pg.escapeLiteral` instead of documenting the `Client.prototype` workaround.

Agreed. It should be fairly simple to do this, and it would be cleaner from a documentation perspective. I'll follow up with those changes

packages/pg/test/unit/client/escape-tests.js
```diff
@@ -1,65 +0,0 @@
-'use strict'
```
brianc, line 1: deleting tests means breaking, non-backwards compatible changes.  We shouldn't do this.  For legacy reasons you should leave the methods on `client` but only document them on the main export `pg` object. They need to be properly deprecated and removed in a multi-month long release cycle.

brianc (changes requested):
I like the thinking here, but the best thing to do is leave them on both `client` and also export them on the root `pg` object and only document them on the `pg.` object. Removing stuff & causing backwards incompatible changes is a months long process w/ deprecation warnings and coordinating other breaking changes and weighing their impact. For the sake of the millions of users the best thing to do is keep it on both `Client.prototype` and add it to `pg` and then in another PR we could deprecate the old usage.

packages/pg/test/unit/client/escape-tests.js
TheConner: Works for me. For what it is worth the tests were moved from client to util, but it is still a breaking change. I wasn't sure if implementing charmander's suggestions would be applicable to this PR, since it was a breaking API change. I'll make these changes backwards compatible, thus no changes to unit tests will be needed.

brianc:
awesome! thank you - yeah we can just have the functions available in both places (with tests for both access paths) & we're good to goooo! I appreciate the work here - over the years I just got very, very cautious w/ breaking changes. :)

TheConner:
@brianc no worries! a future PR or issue can be created for the breaking changes

Updated with requested changes -- these updates are now backwards compatible. This is a matter of preference, but I chose to keep the functions in utils but re-export on client with a disclaimer that it is for backwards compatibility. Restored original tests for client escapes, added an additional test step for the other entry point in utils. Not sure how I feel about testing the second entry point in the client test suite, but the alternative is duplicating the tests for the utils test suite.

packages/pg/lib/client.js
```diff
@@ -456,36 +456,12 @@ class Client extends EventEmitter {
...
-      } else {
-        escaped += c
-      }
-    }
-
-    escaped += "'"
+  // escapeIdentifier and escapeLiteral moved to utility functions & exported
+  // on PG 
+  // re-exported here for backwards compatibility
+  escapeIdentifier = utils.escapeIdentifier
```
charmander, line 462: Need to be defined on the prototype, not as a class field, to avoid breaking backwards compatibility with people who were already using `Client.prototype` directly. Sorry, I should have mentioned that initially.

packages/pg/lib/index.js
```diff
@@ -23,6 +24,8 @@ var PG = function (clientConstructor) {
   this.Connection = Connection
   this.types = require('pg-types')
   this.DatabaseError = DatabaseError
+  this.escapeIdentifier = escapeIdentifier
+  this.escapeLiteral = escapeLiteral
```
charmander, line 28: This is just new though, right? Why?

packages/pg/lib/client.js
TheConner: Right, that was one of the workarounds people are using. I'll fix that, and also add test cases that cover the usage of Client.prototype.escape*, as currently they only cover usages of instantiated clients

packages/pg/lib/index.js
TheConner: This is new yes, as for why [your previous comment](https://github.com/brianc/node-postgres/pull/2954#pullrequestreview-1378415905) suggested that the escape* methods be exported on pg, this adds them as exports to the pg object. On my end it looks like the github diff viewer annoyingly cuts off the part where it shows these belonging to PG, but looking at the [full file](https://github.com/brianc/node-postgres/blob/fe01e3db7f578039a9f4a1daa4cc3d5816c4cb2e/packages/pg/lib/index.js#L18-L29) it shows the rest of the context.

Is there an issue with exporting them this way?

brianc (approved):

PR #2964, opened 2023-04-21
Update transactions.mdx

abel1534:
Hi Brian,
I was implementing transactions and the documentation says we shouldn't use pool.query but the sample code provided looks like it is using "const pool = new Pool()". Shouldn't it be using new Client() instead of new Pool() ? What am I missing?

Thank You so much!

### DIFF
--- a/docs/pages/features/transactions.mdx
+++ b/docs/pages/features/transactions.mdx
@@ -20,8 +20,8 @@ To execute a transaction with node-postgres you simply execute `BEGIN / COMMIT /
 Things are considerably more straightforward if you're using async/await:
 
 ```js
-const { Pool } = require('pg')
-const pool = new Pool()
+const { Client } = require('pg')
+const client = new Client()
 
 // note: we don't try/catch this because if connecting throws an exception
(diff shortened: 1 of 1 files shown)


charmander:
`pool.query` can’t use a transaction, but running a transaction on a client obtained by `pool.connect` is fine.

abel1534:
@brianc Thank You so much for your response. That makes sense. 
When I do try the example in the documentation, in the finally block,  where I have my pgClient.release() method, I get error saying "Property 'release' does not exist on type 'Pool'.

Sample code
try {
    await pgClient.query('BEGIN')
     ......
     await pgClient.query('BEGIN')
} catch(e) {
   await pgClient.query('ROLLBACK')
   throw e
} finally {
   pgClient.release()
}

charmander:
`pgClient` should be a client obtained from the pool with `pool.connect`, not the pool itself.

AbelAbrah:
@charmander Thank you so much. I have implemented it according to your recommendation and it is working. I appreciate it.

PR #2971, opened 2023-05-04
Cloudflare Workers support

petebacondarwin:
CI is green. Please take a look at the code. Probably easiest commit by commit

### DIFF
--- a/packages/pg-connection-string/index.js
+++ b/packages/pg-connection-string/index.js
@@ -1,8 +1,5 @@
 'use strict'
 
-var url = require('url')
-var fs = require('fs')
-
 //Parse method copied from https://github.com/brianc/node-postgres
 //Copyright (c) 2010-2014 Brian Carlson (brian.m.carlson@gmail.com)
 //MIT License
@@ -11,51 +8,53 @@ var fs = require('fs')
 function parse(str) {
   //unix socket
   if (str.charAt(0) === '/') {
-    var config = str.split(' ')
+    const config = str.split(' ')
     return { host: config[0], database: config[1] }
   }
 
-  // url parse expects spaces encoded as %20
-  var result = url.parse(
-    / |%[^a-f0-9]|%[a-f0-9][^a-f0-9]/i.test(str) ? encodeURI(str).replace(/\%25(\d\d)/g, '%$1') : str,
-    true
-  )
-  var config = result.query
-  for (var k in config) {
-    if (Array.isArray(config[k])) {
-      config[k] = config[k][config[k].length - 1]
-    }
+  // Check for empty host in URL
+
+  const config = {}
+  let result
+  let dummyHost = false
+  if (/ |%[^a-f0-9]|%[a-f0-9][^a-f0-9]/i.test(str)) {
+    // Ensure spaces are encoded as %20
+    str = encodeURI(str).replace(/\%25(\d\d)/g, '%$1')
   }
 
-  var auth = (result.auth || ':').split(':')
-  config.user = auth[0]
-  config.password = auth.splice(1).join(':')
+  try {
+    result = new URL(str, 'postgres://base')
+  } catch (e) {
+    // The URL is invalid so try again with a dummy host
+    result = new URL(str.replace('@/', '@___DUMMY___/'), 'postgres://base')
+    dummyHost = true
+  }
+
+  // We'd like to use Object.fromEntries() here but Node.js 10 does not support it
+  for (const entry of result.searchParams.entries()) {
+    config[entry[0]] = entry[1]
+  }
+
+  config.user = config.user || decodeURIComponent(result.username)
+  config.password = config.password || decodeURIComponent(result.password)
 
   config.port = result.port
   if (result.protocol == 'socket:') {
     config.host = decodeURI(result.pathname)
-    config.database = result.query.db
-    config.client_encoding = result.query.encoding
+    config.database = result.searchParams.get('db')
+    config.client_encoding = result.searchParams.get('encoding')
     return config
   }
+  const hostname = dummyHost ? '' : result.hostname
   if (!config.host) {
     // Only set the host if there is no equivalent query param.
-    config.host = result.hostname
+    config.host = decodeURIComponent(hostname)
+  } else if (hostname) {
+    result.pathname = hostname + result.pathname
   }
 
-  // If the host is missing it might be a URL-encoded path to a socket.
-  var pathname = result.pathname
-  if (!config.host && pathname && /^%2f/i.test(pathname)) {
-    var pathnameSplit = pathname.split('/')
-    config.host = decodeURIComponent(pathnameSplit[0])
-    pathname = pathnameSplit.splice(1).join('/')
-  }
-  // result.pathname is not always guaranteed to have a '/' prefix (e.g. relative urls)
-  // only strip the slash if it is present.
-  if (pathname && pathname.charAt(0) === '/') {
-    pathname = pathname.slice(1) || null
-  }
-  config.database = pathname && decodeURI(pathname)
+  const pathname = result.pathname.slice(1) || null
+  config.database = pathname ? decodeURI(pathname) : null
 
   if (config.ssl === 'true' || config.ssl === '1') {
     config.ssl = true
@@ -68,6 +67,9 @@ function parse(str) {
   if (config.sslcert || config.sslkey || config.sslrootcert || config.sslmode) {
     config.ssl = {}
   }
+
+  // Only try to load fs if we expect to read from the disk
+  const fs = config.sslcert || config.sslkey || config.sslrootcert ? require('fs') : null
 
   if (config.sslcert) {
     config.ssl.cert = fs.readFileSync(config.sslcert).toString()
--- a/packages/pg/lib/connection.js
+++ b/packages/pg/lib/connection.js
@@ -92,7 +92,8 @@ class Connection extends EventEmitter {
         }
       }
 
-      if (net.isIP(host) === 0) {
+      var net = require('net')
+      if (net.isIP && net.isIP(host) === 0) {
         options.servername = host
       }
       try {
--- a/packages/pg/lib/connection.js
+++ b/packages/pg/lib/connection.js
@@ -4,6 +4,7 @@ var net = require('net')
 var EventEmitter = require('events').EventEmitter
 
 const { parse, serialize } = require('pg-protocol')
+const { getStream, getSecureStream } = require('./stream')
 
 const flushBuffer = serialize.flush()
 const syncBuffer = serialize.sync()
@@ -15,7 +16,7 @@ class Connection extends EventEmitter {
     super()
     config = config || {}
 
-    this.stream = config.stream || new net.Socket()
+    this.stream = config.stream || getStream(config.ssl)
     if (typeof this.stream === 'function') {
       this.stream = this.stream(config)
     }
@@ -79,7 +80,6 @@ class Connection extends EventEmitter {
           self.stream.end()
           return self.emit('error', new Error('There was an error establishing an SSL connection'))
       }
-      var tls = require('tls')
       const options = {
         socket: self.stream,
       }
@@ -97,7 +97,7 @@ class Connection extends EventEmitter {
         options.servername = host
       }
       try {
-        self.stream = tls.connect(options)
+        self.stream = getSecureStream(options)
       } catch (err) {
         return self.emit('error', err)
       }
--- a/packages/pg/lib/crypto/utils.js
+++ b/packages/pg/lib/crypto/utils.js
@@ -1,0 +1,92 @@
+'use strict'
+
+const useLegacyCrypto = parseInt(process.versions && process.versions.node && process.versions.node.split('.')[0]) < 15
+if (useLegacyCrypto) {
+  // We are on an old version of Node.js that requires legacy crypto utilities.
+  module.exports = require('./utils-legacy')
+  return
+}
+
+const nodeCrypto = require('crypto')
+
+module.exports = {
+  postgresMd5PasswordHash,
+  randomBytes,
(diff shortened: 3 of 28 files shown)


packages/pg/lib/stream.js
```diff
@@ -0,0 +1,176 @@
...
+  if (tls.connect) {
+    return tls.connect(options)
+  } else {
+    options.socket.startTls(options)
+    return options.socket
+  }
+}
+
+/**
+ * Wrapper around the Cloudflare built-in socket that can be used by the `Connection`.
```
charmander, line 30: Looks like something that could go in another package.

packages/pg/lib/stream.js
```diff
@@ -0,0 +1,176 @@
...
+    return this
+  }
+
+  destroy(reason) {
+    log('destroying CF socket', reason)
+    this.destroyed = true
+    return this.end()
+  }
+
+  startTls(options) {
```
jasnell, line 151: I'd recommend setting an internal state that prevents `startTls` from being called twice here. The `this._cfSocket.startTls()` call will throw but only after you've released the locks. Better to check that before releasing.

packages/pg-connection-string/index.js
```diff
@@ -69,16 +68,23 @@ function parse(str) {
...
+    }
 
-  if (config.sslrootcert) {
-    config.ssl.ca = fs.readFileSync(config.sslrootcert).toString()
+    if (config.sslrootcert) {
+      const fs = require('fs')
+      config.ssl.ca = fs.readFileSync(config.sslrootcert).toString()
+    }
+  } catch (e) {
+    throw new Error('filesystem not supported')
```
charmander, line 87: Or any of the files wasn’t readable. Just the `require` should go in `try`, ideally.

packages/pg/lib/connection.js
```diff
@@ -92,11 +91,17 @@ class Connection extends EventEmitter {
...
 
-      if (net.isIP(host) === 0) {
-        options.servername = host
+      try {
+        var net = require('net')
+        if (net.isIP && net.isIP(host) === 0) {
+          options.servername = host
+        }
+      } catch (e) {
+        this.emit('error', e)
```
charmander, line 100: Why catch this?

packages/pg/lib/stream.js
```diff
@@ -0,0 +1,176 @@
...
+      const options = this.ssl ? { secureTransport: 'allow' } : {}
+      const { connect } = await import('cloudflare:sockets')
+      this._cfSocket = connect(`${host}:${port}`, options)
+      this._cfWriter = this._cfSocket.writable.getWriter()
+
+      this._cfSocket.closed.then(() => {
+        if (!this._upgrading) {
+          log('CF socket closed')
+          this._cfSocket = null
+          this.emit('close')
```
jasnell, line 74: If any of the registered `close` event handlers throw an error synchronously, the promise returned by this then call is going to register an unhandled rejection. It would be good to attach a catch handler to this that forwards the error to an `error` event on this class.

packages/pg/lib/stream.js
```diff
@@ -0,0 +1,176 @@
...
+    log('destroying CF socket', reason)
+    this.destroyed = true
+    return this.end()
+  }
+
+  startTls(options) {
+    this._cfWriter.releaseLock()
+    this._cfReader.releaseLock()
+    this._upgrading = true
+    this._cfSocket = this._cfSocket.startTls(options)
```
jasnell, line 155: This will need to `this._cfSocket.closed` then handler to be reapplied to the replacement socket correct?

packages/pg/lib/crypto/crypto-shim.js
```diff
@@ -0,0 +1,101 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+const sha256Sync = require('./sha256')
+
+/**
+ * Patch the Node.js crypto APIs to work outside Node.js
```
charmander, line 24: This looks very hacky. What’s in the way of using the Web Crypto API here?

charmander (changes requested):

packages/pg/lib/crypto/crypto-shim.js
jasnell: I have to agree, this is definitely suboptimal in a number of ways.

packages/pg/lib/crypto/crypto-shim.js
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  crypto.createHmac = function createHmac(type, key) {
+    if (type !== 'sha256') throw new Error('Only sha256 is supported')
+    return {
+      update: function (data) {
+        return {
+          digest: function () {
+            if (typeof key === 'string') key = new TextEncoder().encode(key)
+            if (typeof data === 'string') data = new TextEncoder().encode(data)
```
jasnell, line 51: Nit: there's no reason to create two separate `TextEncoder` instances here. Just create one at top level and reuse it.

packages/pg/lib/crypto/crypto-shim.js
```diff
@@ -0,0 +1,101 @@
...
+    }
+  }
+
+  crypto.createHmac = function createHmac(type, key) {
+    if (type !== 'sha256') throw new Error('Only sha256 is supported')
+    return {
+      update: function (data) {
+        return {
+          digest: function () {
+            if (typeof key === 'string') key = new TextEncoder().encode(key)
```
jasnell, line 50: I think what you want here, actually, is `Buffer.from(key)`

packages/pg/lib/crypto/crypto-shim.js
```diff
@@ -0,0 +1,101 @@
...
+            result.set(sha256Sync(msg), 64)
+
+            return Buffer.from(sha256Sync(result))
+          },
+        }
+      },
+    }
+  }
+
+  crypto.pbkdf2Sync = function pbkdf2Sync(password, saltBytes, iterations, keylen, digest) {
```
jasnell, line 85: One limitation that Web Crypto does have here is that it cannot do PBKDF2 synchronously.

charmander: Right, but as far as I know, we don’t need it to be synchronous (and it’s possible that it never should have been synchronous in the first place?).

brianc: > (and it’s possible that it never should have been synchronous in the first place?).

yeah probably should have been async since blocking calls in a code path like this === bad

packages/pg/lib/crypto/crypto-shim.js
petebacondarwin: Probably, I just copied this over verbatim from the Neon serverless shims.
I can spend some time optmizing all of this now that it is proved to work.

packages/pg/lib/crypto/crypto-shim.js
petebacondarwin: Yep!

packages/pg/lib/crypto/crypto-shim.js
petebacondarwin: Yep - I was trying to touch the original code as little as possible.
But now that this PR has shown this to be working I can look at further refactoring and improvements.

packages/pg/lib/crypto/crypto-shim.js
```diff
@@ -0,0 +1,101 @@
...
+    return globalThis.crypto.getRandomValues(Buffer.alloc(length))
+  }
+
+  crypto.createHash = function (type) {
+    if (type !== 'sha256') throw new Error('Only sha256 is supported')
+    return {
+      update: function (data) {
+        return {
+          digest: function () {
+            return Buffer.from(sha256Sync(data))
```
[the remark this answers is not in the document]
petebacondarwin: It was a very narrow hacky shim for this particular use case.
I think the best way forward is to try to move off this API altogether.

packages/pg/lib/connection.js
petebacondarwin: There are two approaches to Node.js compatibility in Cloudflare Workers right now.
One uses compile-time shims; the other uses packages built in to the runtime.
The former case, I think, provides an empty object for this import; the latter will throw.
Currently I could not quite get the latter approach working, so I reverted to the former, which probably does not need this try-catch.

I'll see if we can remove it.

packages/pg/lib/stream.js
petebacondarwin: Good spot!

petebacondarwin:
Converting back to draft while I refactor things on the back of the reviews. Thanks @brianc @jasnell and @charmander

elithrar:
> That would involve pulling out the crypto and fs access functions and delegating them to something passed in the connection config object (defaulting to loading node crypto and fs).

Can you expand on this? If this becomes an explicit config option, it becomes much higher friction for users & ORMs that depend on `pg` end up requiring changes as well, which doesn't seem scalable.

Note that this should allow node-postgres to work in [other serverless environments](https://wintercg.org/), like Deno and Bun, once they also allow TCP socket creation. The crypto & FS shims are relevant there too.

packages/pg-connection-string/index.js
petebacondarwin: Yep! Fixing.

packages/pg/lib/crypto/utils.js
```diff
@@ -0,0 +1,85 @@
...
+  sha256,
+  hmacSha256,
+  md5,
+}
+
+/**
+ * The Web Crypto API - grabbed from the Node.js library or the global
+ * @type Crypto
+ */
+const webCrypto = nodeCrypto.webcrypto || globalThis.crypto
```
petebacondarwin, line 18: This is basically the decision point between using internal Node.js APIs and built-in serverless (i.e. Cloudflare) ones.

petebacondarwin:
@sehrope - thanks for your comments. That and the other reviews gave me enough motivation to go and do further refactoring to avoid the Node.js API shims altogether. Now the code is mostly using the more modern WebCrypto APIs.

packages/pg/lib/stream.js
petebacondarwin: Sure. Thanks.

sehrope:
> Can you expand on this? If this becomes an explicit config option, it becomes much higher friction for users & ORMs that depend on pg end up requiring changes as well, which doesn't seem scalable.

Of course it's higher friction if you're going to compare it to zero friction. But that's what you get for running on something that purports to be node but doesn't provide the entire userspace.

The Cloudflare worker docs already instruct the end user they need to explicitly enable a flag to turn on nodejs compatibility for things like Buffer:

> https://developers.cloudflare.com/workers/runtime-apis/nodejs/buffer/
>
> To use Node.js APIs in your Worker, add the nodejs_compat compatibility flag to your wrangler.toml file.

Ideally any special handling to make the environment node-like should be happening on Cloudflare's end.

If need be, we can expand the configurability of this module to facilitate that. But whatever leverages those options to provide support for a specific provider should be owned by that provider.

> Note that this should allow node-postgres to work in other serverless environments, like Deno and Bun, once they also allow TCP socket creation. The crypto & FS shims are relevant there too.

If there's an actual standard or API then sure, but rolling our own sha256 or adding a service providers specific socket implementation (e.g. `class CFSocket`) is not sustainable. We're not going to support N such implementations and be worried about subtlety breaking one of them.

Again, I'm not against adding support for this use case, I just think it should be done in a way that minimizes code ownership for this project and allows multiple implementations.

The PR for what I'm describing would not have any provider specific code. It'd be a series of commits to pull out internal references to things like crypto operations (for md5 and sha256), reading from the filesystem / pgpass, and creating the socket streams. The defaults would be using the nodejs built-ins for crypto, fs, and net so there would be no net change to the module. Anybody that wants to override it for a non-nodejs platform would shim those fields, either globally or in the config object passed to pg.

A good analogy is the addition of dynamic passwords to this module. IIRC, the original request was someone wanting to authenticate with AWS IAM auth against an RDS database. Rather than adding in the AWS SDK and baking in a provider specific option for signing auth requests, we have a generic interface that allows specifying an async function to provide a password dynamically. That's applicable to N providers and it's up to each of them to support however they work internally.

packages/pg/lib/stream.js
petebacondarwin: Moved to a new pg-cloudflare package

petebacondarwin:
Ideally we would have `pg-cloudflare` as a simple dependency of `pg`, so that there is nothing more for the user to do other than add the library to their Worker project.

petebacondarwin:
Oh no 😭 WebCrypto.subtle only became available in node 15.

petebacondarwin:
All green @brianc - please take a look when you have some time.

brianc:
This is a huge PR - thank you so much for putting in the time and effort here from everyone involved. 😄  For a bit of background I've been meeting with cloudflare about this both in person here in Austin, TX and on video calls off and on for a month discussing this & so on.  Really cool to see how little actual code change was involved.  I hear the concern about making things more generic and accepting shims being injected in...there is still a bunch of churn on some of these web crypto and stream modules...hopefully they settle down & get "standardized" soon and we can look at making some of these injectables more dynamic. I'm cool w/ having them more bespoke when there are only node & CF workers being supported & can work towards more dynamic injectable things in the future if/when a 3rd supported platform wants to add their stuff.

brianc (approved):

PR #2977, opened 2023-05-16
Fix connection string parsing for overriden hosts

### LINKED ISSUE #2976: Change in behaviour of connection string parsing
Bug hasn't been fully confirmed yet, but I suspect there is an issue with the recent changes to pg-connection-string.

I think https://github.com/brianc/node-postgres/commit/f305419676afe79a91dc7863289f7f97ac6db3d5 may have inadvertently changed the behaviour of connection string parsing, for strings of the form:

```
postgresql://myuser:mypass@localhost/mydb?host=/cloudsql/my-project:europe-west2:my-sql-instance
```

This style connection string I believe was previously recommended in the Google Cloud docs for connecting to Google Cloud SQL instances, because the host includes forward slashes (although might not not have been, and doesn't seem to be the case any more).

domdomegg:
Fixes #2976

### DIFF
--- a/packages/pg-connection-string/index.js
+++ b/packages/pg-connection-string/index.js
@@ -50,6 +50,7 @@ function parse(str) {
     // Only set the host if there is no equivalent query param.
     config.host = decodeURIComponent(hostname)
   } else if (hostname && /^%2f/i.test(hostname)) {
+    // Only prepend the hostname to the pathname if it is not a URL encoded Unix socket host.
     result.pathname = hostname + result.pathname
   }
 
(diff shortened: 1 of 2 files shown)


packages/pg-connection-string/index.js
```diff
@@ -49,7 +49,7 @@ function parse(str) {
   if (!config.host) {
     // Only set the host if there is no equivalent query param.
     config.host = decodeURIComponent(hostname)
-  } else if (hostname) {
+  } else if (hostname && /^%2f/i.test(hostname)) {
```
petebacondarwin, line 52: Nice fix. Should probably add a comment here explaining the logic.

Eg.

```suggestion
  } else if (hostname && /^%2f/i.test(hostname)) {
    // Only prepend the hostname to the pathname if it is not a URL encoded Unix socket host.
```

petebacondarwin (approved):

brianc:
Hey thanks for this! ❤️

brianc:
For sure! Thought I did, sorry!

PR #2978, opened 2023-05-16
fix: ensure that pg-cloudflare can be used with bundlers that don't know about Cloudflare sockets

### LINKED ISSUE #2975: v8.11.0 UnhandledSchemeError when bundling with Webpack
This [line](https://github.com/brianc/node-postgres/blob/14b840e96e57fc0617b5c4758f6318f774148ee4/packages/pg-cloudflare/src/index.ts#L1) introduced in [this PR](https://github.com/brianc/node-postgres/pull/2971) is causing errors in Webpack builds.

```
Webpack compilation failed:
in cloudflare:sockets
  Module build failed: UnhandledSchemeError: Reading from "cloudflare:sockets" is not handled by plugins (Unhandled scheme).
  Webpack supports "data:" and "file:" URIs by default.
  You may need an additional plugin to handle "cloudflare:" URIs.
```

This occurred after a minor update to the `pg` package from `v8.10.0` to `v8.11.0` when the `pg-cloudflare` package was added as an optional dependency.

For now, we've reverted to `v8.10.0`.

I expect this will affect a lot of people using Webpack as they start to update.

petebacondarwin:
Fixes #2975

By implementing package.json `exports` we can avoid processing the Cloudflare specific code, which contains `import ... from "cloudflare:sockets"`, in bundlers such as Webpack.

If you are bundling for a Worker environment using Webpack then you need to add the `workerd` condition and ignore `cloudflare:sockets` imports:

**webpack.config.js**
```js
  resolve: { conditionNames: ["require", "node", "workerd"] },
  plugins: [
    new webpack.IgnorePlugin({
      resourceRegExp: /^cloudflare:sockets$/,
    }),
  ],
```

### DIFF
--- a/packages/pg-cloudflare/package.json
+++ b/packages/pg-cloudflare/package.json
@@ -2,12 +2,16 @@
   "name": "pg-cloudflare",
   "version": "1.1.0",
   "description": "A socket implementation that can run on Cloudflare Workers using native TCP connections.",
-  "main": "dist/index.js",
+  "main": "dist/empty.js",
   "types": "dist/index.d.ts",
   "license": "MIT",
   "devDependencies": {
     "ts-node": "^8.5.4",
     "typescript": "^4.0.3"
+  },
+  "exports": {
+    "workerd": "./dist/index.js",
+    "default": "./dist/empty.js"
   },
   "scripts": {
     "build": "tsc",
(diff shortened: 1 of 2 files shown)


brianc:
@petebacondarwin you still working on this (it's set to draft) - getting repeated issues about this & I'd love to help get this one merged in.

petebacondarwin:
Oh sorry. I thought it was set to ready.
This works for me. It would be good if those hitting the issue could try it

brianc:
Ah great thanks!  I'll go ahead & get it released & follow up on the linked issues. ❤️

IqbalLx:
what about vite ?

Feuerhamster:
> what about vite ?

+1

Issue #2622, opened 2021-09-24
Stack trace not including the calling location.


agirorn:
When an error is rejected from a query, then the stack trace is not complete. It does not include the calling code and that makes finding and fixing the errors harder than it needs to be. Could we not improve this in the module instead of having users add some wrapping try-catch logic to try to improve this? 

**Sample of and error I just got"
```shell
  error: operator does not exist: text = integer
        at Parser.parseErrorMessage (/Users/aegiro/code/project/node_modules/pg-protocol/src/parser.ts:369:69)
        at Parser.handlePacket (/Users/aegiro/code/project/node_modules/pg-protocol/src/parser.ts:188:21)
        at Parser.parse (/Users/aegiro/code/project/node_modules/pg-protocol/src/parser.ts:103:30)
        at Socket.<anonymous> (/Users/aegiro/code/project/node_modules/pg-protocol/src/index.ts:7:48)
        at Socket.emit (events.js:400:28)
        at addChunk (internal/streams/readable.js:290:12)
        at readableAddChunk (internal/streams/readable.js:265:9)
        at Socket.Readable.push (internal/streams/readable.js:204:10)
        at TCP.onStreamRead (internal/stream_base_commons.js:188:23) {
      length: 200,
      severity: 'ERROR',
      code: '42883',
      detail: undefined,
      hint: 'No operator matches the given name and argument types. You might need to add explicit type casts.',
      position: '295',
      internalPosition: undefined,
      internalQuery: undefined,
      where: undefined,
      schema: undefined,
      table: undefined,
      column: undefined,
      dataType: undefined,
      constraint: undefined,
      file: 'parse_oper.c',
      line: '731',
      routine: 'op_error'
    }
```

Yuval-Peled:
We've also encountered this. This is something that the node.js ecosystem really suffers from, and node.js does not make it easy for library writers in a lot of places.
However, I think this really deserves focus. This is a library that a lot of other libraries depend on, which means that in a lot of cases a DB query error is really hard to debug.

Issue #2484, opened 2021-03-04
preferred package for debugging query text?


jmrlib:
Hi, I'm wondering if there is a preferred package, extension, or method for providing more detail when you get a syntax error back from the parser (for instance, one that listed line number, column of the error) 

for instance, right now:
```
error: syntax error at or near "as"
    at Parser.parseErrorMessage (/home/collspec/projects/staff-portal/sprint-server/node_modules/pg-protocol/dist/parser.js:278:15)
```

desired behavior:
```
error: syntax error at or near "as", line 5, column 7
    at Parser.parseErrorMessage (/home/collspec/projects/staff-portal/sprint-server/node_modules/pg-protocol/dist/parser.js:278:15)
```

vitaly-t:
That error is all the server provides, you can't ask for more, 'cos it just ain't there.

sehrope:
There are a bunch of additional fields on Error objects populated by the driver. If you log the error object you can see them. They correspond to the error fields returned by the server: https://www.postgresql.org/docs/current/protocol-error-fields.html

For example with the command:

```sql
SELECT foo
FROM bar
```

You can get an error like this:

```text
{
  length: 102,
  severity: 'ERROR',
  code: '42P01',
  detail: undefined,
  hint: undefined,
  position: '17',
  internalPosition: undefined,
  internalQuery: undefined,
  where: undefined,
  schema: undefined,
  table: undefined,
  column: undefined,
  dataType: undefined,
  constraint: undefined,
  file: 'parse_relation.c',
  line: '1180',
  routine: 'parserOpenTable'
}
```

The one you want is `position`. It gives you the character offset in the SQL of the error. In this example the position value of "17" refers to the start of the `bar` token in the SQL. It's not always populated though as it depends on what caused the error (generally just parse errors).

abenhamdine:
For information, since version 8.4.0, you can import `DatabaseError` interface (see brianc/node-postgres@54b8752/CHANGELOG.md#pg840)  

the DatabaseError has the following fields :
brianc/node-postgres@6121bd3/packages/pg-protocol/src/messages.ts#L97
unit tests are there :
brianc/node-postgres@4b22927/packages/pg/test/integration/client/query-error-handling-tests.js#L81

offical docs about error properties :
postgresql.org/docs/13/protocol-error-fields.html

official error codes :
postgresql.org/docs/current/errcodes-appendix.html  

Somebody has already converted these on js constants here  :
LinusU/pg-error-constants
[...]

Issue #2470, opened 2021-02-12
Stack traces are truncated / lack client side stack trace


bluenote10:
When `pg` throws an error, I often have trouble finding the place in my code where this is actually coming from. For some reasons the exceptions thrown by `pg` seem to contain strange stack trace, and in particular, the stack trace doesn't contain any client side code. I'm actually on TypeScript running through `ts-node`, but I think I can reproduce the behavior with plain JS. Here is a minimal example:

```js
const pg = require("pg");

async function provokeExceptionPg() {
  let client = new pg.Client();
  await client.connect();
}

provokeExceptionPg().catch((err) => console.log(err));
```

The output of running `node index.js` is just:

```
Error: connect ECONNREFUSED 127.0.0.1:5432
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1146:16) {
  errno: -111,
  code: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 5432
}
```

Note that the output is completely lacking any information of client side code, which in this case would just be my `index.js`. To demonstrate the kind of traceback I'd expect, I can for instance replace `new pg.Client()` by something broken like `new pg.ClientNonexisting()`. This results in a stack trace that clearly shows the client side stack leading to the error:

```
TypeError: pg.ClientNonexisting is not a constructor
    at provokeExceptionPg (/tmp/test_project/index.js:4:16)
    at Object.<anonymous> (/tmp/test_project/index.js:8:1)
    [...]
```

The lack of client side stack traces makes it very hard to understand where errors are coming from in a more complex code base. Is there any way to get more meaningful stack traces out of `pg` or would that make sense as a feature request?

vitaly-t:
[Bring your own promise](https://github.com/brianc/node-postgres/blob/a02dfac5ad2e2abf0dc3a9817f953938acdc19b1/packages/pg-pool/README.md#bring-your-own-promise) - `Bluebird`, and enable [Long Stack Traces](http://bluebirdjs.com/docs/api/promise.longstacktraces.html).

bluenote10:
I had the impression that this is no longer necessary now that `--async-stack-traces` is enabled by default. At least I often get meaningful stack traces from other libraries (presumably using default promises), and e.g. basic examples [like this](https://thecodebarbarian.com/async-stack-traces-in-node-js-12) also seem to contain the full trace by default.

[...]

Issue #1762, opened 2018-11-05
Pool.query() does not throw error with full stack trace


kevin-leptons:
Issue of  **Pool.query()**.

Errors which are throw form Pool.query() does not contains full error stack trace. Please throw standard error with full stack trace.

joelmukuthu:
Hi, could you post the incomplete stack trace and the code that generates the error? Without knowing more about your case however, I’m guessing that it might have something to do with [this infamous bug in node](https://github.com/nodejs/node/issues/11865), which can’t be fixed in this library

kevin-leptons:
Hi Joe,

Yes, I got it, my case is similar like that. Now I know that we can't fix it in this library. What can we do in this case?, or just wait for node.js team?.

joelmukuthu:
You can try some of the workarounds suggested in that issue:

- https://github.com/AndreasMadsen/trace
- https://github.com/artdecocode/erotic
- Bluebird promises with `BLUEBIRD_LONG_STACK_TRACES=1`. You can configure the promise constructor to use and then set the BLUEBIRD_LONG_STACK_TRACES environment variable when running the application:

```js
const bluebird = require('bluebird');
const { Client, Pool } = require('pg');

const pool = new Pool({
  Promise: bluebird
});

// for this one, only available in pg@7.6.0 and up:
const client = new Client({
  Promise: bluebird
});
```

```bash

kevin-leptons:
Thank Joel, I will try all of it. I also let this issue is open to show that it should be solve, some how.

charmander:
@joelmukuthu’s covered it all; stack traces get lost by default in JavaScript when asynchronous tasks aren’t direct calls.

phiresky:
Now that [zero-cost async stack traces](https://v8.dev/blog/fast-async) are part of Node (v12) by default, it would be great if this could be reconsidered. I'm not sure what the reason is that this is still broken in node-pg, maybe because node-pg doesn't use native promises or native async functions?

brianc:
> I'm not sure what the reason is that this is still broken in node-pg

the reason is this project has been maintained almost entirely by myself and 1 other person for the past decade and we're buried under bug reports, feature requests, internal priorities.  Also, the world has fallen into a global pandemic causing me significant anxiety attacks.  Also, we still support versions of node back to 6.0.

brianc:
> maybe because node-pg doesn't use native promises or native async functions?

[...]

PR #2983, opened 2023-05-22
fix stack traces of query() to include the async context (#1762)

phiresky:
This PR replaces the stack trace of errors in pool.query() and client.query() with a more useful stack trace including the callsite:

The following code:

```js
async function innerFunction() {
	const pool = new pg.Pool();
	await pool.query("SELECT test from nonexistent");
}

async function outerFunction() {
	await innerFunction();
}

outerFunction();
```

Throws this with the current version of pg:

```
error: relation "nonexistent" does not exist
    at Parser.parseErrorMessage (.../node_modules/pg-protocol/dist/parser.js:287:98)
    at Parser.handlePacket (.../node_modules/pg-protocol/dist/parser.js:126:29)
    at Parser.parse (.../node_modules/pg-protocol/dist/parser.js:39:38)
    at Socket.<anonymous> (.../node_modules/pg-protocol/dist/index.js:11:42)
    at Socket.emit (node:events:511:28)
    at addChunk (node:internal/streams/readable:332:12)
    at readableAddChunk (node:internal/streams/readable:305:9)
    at Readable.push (node:internal/streams/readable:242:10)
    at TCP.onStreamRead (node:internal/stream_base_commons:190:23)
```

With this PR, it throws this much more useful error:

```
error: relation "nonexistent" does not exist
    at .../node_modules/pg-pool/index.js:45:11
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async innerFunction (my-own-code.js:10:3)
    at async outerFunction (my-own-code.js:14:3)
```

## Why

For larger code-bases it has been a long-standing issue of not being able to find out which query or code a PG error comes from. I have years of logs with errors like `error: invalid input syntax for type bigint: "NaN"` or `error: could not serialize access due to concurrent update` with no way to know where they are coming from. This should fix that by directly linking the calling functions in the stack trace.

## How it works

Since Node 16, Node has included stack tracing for async functions. The reason these don't work out of the box in node-pg is that the response to a query comes from a different stack resulting from the message received via TCP, and this stack is not linked anywhere to the calling stack. As far as I know there's no way to directly do this, there's an API to do this for the async_context API (https://nodejs.org/api/async_context.html#static-method-asyncresourcebindfn-type-thisarg) that AFAIK uses the same technology as the zero-cost async stack traces, but it doesn't work for the integrated Error stack tracing.

Instead, the solution is to simply replace the Error stack at the first point where the async execution context is the "correct" one by catching and rethrowing the error. This should come at no cost for the happy path.

## Limitations

-   This only works for promise-based code. I'm not sure if it works for mixes of async functions with `.then(), .catch()` style calls but it definitely doesn't work for purely callback-based code. (this doesn't break anything, worst case you just won't get useful error traces as before)
-   This code fully replaces the stack trace, losing the original "parseErrorMessage" stack. I've never found that part useful so I didn't preserve it. It would be neatly possible though e.g. through the error.cause property: `...catch(e => {e.cause = Object.assign(Object.create(Error.prototype), { stack: e.stack }); Error.captureStackTrace(e); throw e});`
- I didn't implement it for pg-native though the same thing should work?

## Compatibility

This code really only uses the Error.captureStackTrace function which has been there since [Node 8](https://nodejs.org/docs/latest-v8.x/api/errors.html#errors_error_capturestacktrace_targetobject_constructoropt). So this should be compatible starting with node 8+ and useful starting with Node 14 (with `--async-stack-traces`) and Node 16 (out of the box).

## References

- fixes https://github.com/brianc/node-postgres/issues/1762
- fixes https://github.com/brianc/node-postgres/issues/2622
- fixes https://github.com/brianc/node-postgres/issues/2484
- partially fixes https://github.com/brianc/node-postgres/issues/2470
- related issue: https://github.com/brianc/node-postgres/issues/1619
-   [zero-cost async stack traces design document](https://docs.google.com/document/d/13Sy_kBIJGP0XT34V1CV3nkWya4TwYx9L3Yv45LdGB6Q/edit)
-   [V8 blog about async stack traces](https://v8.dev/blog/fast-async)
- [similar issue in http](https://github.com/nodejs/node/issues/31297)
- [open knex issue](https://github.com/knex/knex/issues/3796)
- [closed node issue](https://github.com/nodejs/node/issues/11865)


charmander (approved):

phiresky:
Something's wrong, the tests should be failing on Node 10 and 12 and succeeding on 16+

phiresky:
The tests weren't run because I didn't name the test file *-tests.js. They should run now hopefully.

brianc:
This is pretty fantastic! I'll get this merged up and released today. 😄  Thanks a bunch!!

dpkirchner:
Was this code reverted? It doesn't look like it to me, but I'm still seeing the original TCP.onStreamRead,etc stack traces. I am using knex as an abstraction around pg, so I guess that might be affecting things. If it wasn't reverted, I can open a new issue.

Issue #2723, opened 2022-03-19
Support the PGSSLCERT, PGSSLKEY, PGSSLROOTCERT environment variables


rafiss:
### Summary 

According to [the docs](https://node-postgres.com/features/connecting), node-postgres uses the same [environment variables](https://www.postgresql.org/docs/9.1/static/libpq-envars.html) as libpq to connect to a PostgreSQL server.

However, PGSSLCERT, PGSSLKEY, PGSSLROOTCERT are not supported. 

### To reproduce

Use the following script
```
const {Client, Pool} = require("./packages/pg")

const client = new Client()
client.connect(err => {
if (err) {
    console.error('error connecting', err.stack)
} else {
    console.log('connected')
    client.end()
}
})

const pool = new Pool()
pool
.connect()
.then(client => {
    console.log('connected')
    client.release()
})
.catch(err => console.error('error connecting', err.stack))
.then(() => pool.end())
```

Run it with `PGSSLMODE=require PGSSLCERT=/home/ubuntu/certs/client.testuser.crt PGSSLROOTCERT=/home/ubuntu/certs/ca.crt PGSSLKEY=/home/ubuntu/certs/client.testuser.key PGHOST=localhost PGPORT=26257 PGUSER=root node test.js`

It results in the error
```
error connecting Error: unable to verify the first certificate
    at TLSSocket.onConnectSecure (_tls_wrap.js:1497:34)
    at TLSSocket.emit (events.js:315:20)
    at TLSSocket._finishInit (_tls_wrap.js:932:8)
    at TLSWrap.ssl.onhandshakedone (_tls_wrap.js:706:12)
error connecting Error: unable to verify the first certificate
    at TLSSocket.onConnectSecure (_tls_wrap.js:1497:34)
    at TLSSocket.emit (events.js:315:20)
    at TLSSocket._finishInit (_tls_wrap.js:932:8)
    at TLSWrap.ssl.onhandshakedone (_tls_wrap.js:706:12)
```

This is because the ssl field in ConnectionParameters is simply set to true and the certs fields are not populated

```
ConnectionParameters {
  user: 'root',
  database: 'root',
  port: 26257,
  host: 'localhost',
  binary: false,
  options: undefined,
  ssl: true,
  client_encoding: '',
  replication: undefined,
  isDomainSocket: false,
  application_name: undefined,
  fallback_application_name: undefined,
  statement_timeout: false,
  idle_in_transaction_session_timeout: false,
  query_timeout: false,
[...]

PR #2994, opened 2023-06-01
Update packages/pg/lib/connection-parameters.js. Fix SSL issue (#2723)

dapeleg-dn:


packages/pg/lib/connection-parameters.js
```diff
@@ -26,7 +27,35 @@ var readSSLConfigFromEnvironment = function () {
...
+
+      var keyFileName = val('sslkey', {})
+      if (!keyFileName) {
+        throw new Error('Postgres SSL connection requested using PGSSLMODE environment variable, but no PGSSLKEY environment variable defined')
+      }
+
+      var result = {
+        rejectUnauthorized: false,
+        ca: fs.readFileSync(caFileName).toString(),
+        cert: fs.readFileSync(crtFileName).toString()
```
brianc, line 48: We don't want blocking calls in here

packages/pg/lib/connection-parameters.js
```diff
@@ -26,7 +27,35 @@ var readSSLConfigFromEnvironment = function () {
...
+      }
+
+      // "hiding" the private key so it doesn't show up in stack traces
+      // or if the client is console.logged
+      Object.defineProperty(result, 'key', {
+        enumerable: false,
+        value: fs.readFileSync(keyFileName).toString()
```
brianc, line 55: please don't use blocking calls - this lib is supposed to be 100% non-blocking.

packages/pg/lib/connection-parameters.js
```diff
@@ -26,7 +27,35 @@ var readSSLConfigFromEnvironment = function () {
...
+      }
+
+      var result = {
+        rejectUnauthorized: false,
+        ca: fs.readFileSync(caFileName).toString(),
+        cert: fs.readFileSync(crtFileName).toString()
+      }
+
+      // "hiding" the private key so it doesn't show up in stack traces
+      // or if the client is console.logged
```
brianc, line 52: nice

brianc (changes requested):
I appreciate the effort and thought here! A couple of concerns

1) all code merged needs supporting tests. Having to support this code base long after you may move on from postgres or even node.js its the only way I can keep my sanity

2) don't use sync APIs - any file reading needs to be non-blocking.

packages/pg/lib/connection-parameters.js
dapeleg-dn: Sure. Taken from elsewhere in the code.

packages/pg/lib/connection-parameters.js
dapeleg-dn: Understood.

dapeleg-dn: BTW I already see some other sync file reads like this one: https://github.com/brianc/node-postgres/blob/master/packages/pg-connection-string/index.js#L76

I may fix it as well if you prefer.

dapeleg-dn: I tried to get rid of the synchronous `fs.readFileSync()` by using `fs.promises.readFile()`; however, this makes the whole constructor of the Client `async` as well, which is a significant breaking change and tests are failing as well.

Since there was already `fs.readFileSync()` in the code, as I mentioned in the previous comment, and since these bugs of TLS support are blocking me (and probably many other people), I would like to ask you to approve them as synchronous for the moment. Anyway, this is only a read from a (probably) local file system, this is not a client-server communication, so I guess it is worth it.

As an improvement for the future, I suggest not resolving the `ssl*` settings during connection string parsing and environment variable resolving, which are done in the constructor. They can be deferred to the `connect()` phase, which is already asynchronous. I think this is also the behavior of `psql` and `libpq`; they only get the NAMES of the files, not their content. But this change is a breaking change since existing config options are the TLS file **contents** themselves, not just the TLS file names.

packages/pg/lib/connection-parameters.js
dapeleg-dn: See my other comment later on.

dapeleg-dn:
@brianc, please take a look. I force-pushed the code. I walked in a slightly different way, allowing you to mix TLS definitions in both environment variables and connection string or config options.

Tests have been added as well for your request.

dapeleg-dn:
@brianc, are you okay with the changes?

packages/pg/lib/connection-parameters.js
```diff
@@ -1,6 +1,7 @@
 'use strict'
 
 var dns = require('dns')
+const _ = require('lodash')
```
charmander, line 4: Undeclared dependency, and too heavy just to check if an object is empty (should pull in just https://www.npmjs.com/package/lodash.isempty if really necessary).

packages/pg/test/unit/connection-parameters/environment-variable-tests.js
```diff
@@ -3,6 +3,11 @@ var helper = require('../test-helper')
 const Suite = require('../../suite')
 
 var assert = require('assert')
+const fs = require('fs')
+
+const tmp = require('tmp')
```
charmander, line 8: Another undeclared dependency. Can the files just be part of the repo instead of written dynamically by test code?

packages/pg/lib/connection-parameters.js
```diff
@@ -33,6 +34,27 @@ var readSSLConfigFromEnvironment = function () {
...
+var fillInSSLVariablesFromEnvironment = function(existingSsl) {
+  var caFileName = val('sslrootcert', existingSsl)
+  var crtFileName = val('sslcert', existingSsl)
+  var keyFileName = val('sslkey', existingSsl)
+
+  // Only try to load fs if we expect to read from the disk
+  const fs = caFileName || crtFileName || keyFileName ? require('fs') : null
+  var result = {}
+  if (caFileName) {
+    result.ca = fs.readFileSync(caFileName).toString()
```
charmander, line 46: Implicit sync calls that happen on every connection might not be good (one good reason for user code to be responsible for this…)

packages/pg/lib/connection-parameters.js
```diff
@@ -89,6 +111,23 @@ class ConnectionParameters {
...
+    if (this.ssl === true) {
+      var sslCandidate = fillInSSLVariablesFromEnvironment({})
+      
+      // Change the type of this.ssl from boolean to object only if relevant environment variables were defined
+      if (!_.isEmpty(sslCandidate)) {
+        this.ssl = sslCandidate
+      }
+    } else if (typeof this.ssl === 'object') {
+      var sslCandidate = fillInSSLVariablesFromEnvironment({})
+      this.ssl = Object.assign(this.ssl, sslCandidate)
```
charmander, line 126: Potentially modifies user-provided `config.ssl` object in place.

```suggestion
      this.ssl = {...this.ssl, ...sslCandidate}
```

packages/pg/lib/connection-parameters.js
```diff
@@ -33,6 +34,27 @@ var readSSLConfigFromEnvironment = function () {
   return defaults.ssl
 }
 
+var fillInSSLVariablesFromEnvironment = function(existingSsl) {
```
charmander, line 37: `existingSsl` is always `{}`?

charmander (changes requested):

packages/pg/lib/connection-parameters.js
dapeleg-dn: Regarding sync / async, read this: https://github.com/brianc/node-postgres/pull/2994#discussion_r1218162821

charmander:
@adaboese You can pass your SSL configuration in the `ssl` property of the pg configuration object instead of implicitly through environment variables (and that `ssl` property can still be constructed from the same environment variables if you want).

brianc:
@adaboese [here's the documentation](https://node-postgres.com/apis/client#new-client) - I hope its clear!

I'm going to close this PR as it doesn't seem to be being moved forward. There's an open issue (#2723) so we can take another shot at it w/o loadash and soforth.

PR #3004, opened 2023-06-08
Catch errors thrown by connect

xvello:
We had a user submit a PG connstring with `?ssl=off`, while this is invalid, it's not failing the parsing, and connect() throws on:

> Cannot use 'in' operator to search for 'key' in off TypeError: Cannot use 'in' operator to search for 'key' in off at Socket.<anonymous> (/Users/xavier/git/node-postgres/packages/pg/lib/connection.js:90:19)

Because `Connection` extends `EventEmitter`, that exception is not `catch`-able with a `try` / `catch` around the call to `client.connect()`. I am following on the footsteps of https://github.com/brianc/node-postgres/pull/2569, and making sure that exception is properly caught internally.

Ideally, this `?ssl=off` value should trigger a validation error, but:
  - I'm not certain not to break compat by adding a stricter input validation
  - With this PR, we're possibly covering other edge cases

Added a unit test, that fails if `connection.js` is not patched: an unhandled exception bubbles up.

### DIFF
--- a/packages/pg/lib/connection.js
+++ b/packages/pg/lib/connection.js
@@ -84,19 +84,20 @@ class Connection extends EventEmitter {
         socket: self.stream,
       }
 
-      if (self.ssl !== true) {
-        Object.assign(options, self.ssl)
-
-        if ('key' in self.ssl) {
-          options.key = self.ssl.key
-        }
-      }
-
-      var net = require('net')
-      if (net.isIP && net.isIP(host) === 0) {
-        options.servername = host
-      }
       try {
+        if (self.ssl !== true) {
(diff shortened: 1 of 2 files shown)


xvello:
Heya @charmander would you have a moment to review this one? It's a pretty small change.

charmander (changes requested):
> Ideally, this `?ssl=off` value should trigger a validation error, but:
> 
> - I'm not certain not to break compat by adding a stricter input validation
> - With this PR, we're possibly covering other edge cases

This strikes me as the bad kind of defensive programming. An early, strict validation of the value of `ssl` would be better. We can note the compatibility and the edge cases properly then.

PR #3005, opened 2023-06-11
Update pg-connection-string url in connecting.mdx

rikurauhala:
The contents of the repository the link is pointing to have been moved to the monorepo.


brianc:
ty!

PR #2932, opened 2023-03-16
Make `async/await` the primary examples in the docs

rijkvanzanten:
Updated all the examples in the docs to use async/await. Callbacks are mentioned as an availability in the main index, but are omitted from the examples to simplify the overall use.

I also updated the examples to use CJS instead of ESM. This is still a bit of a touchy subject in NodeJS land, so no hard feelings if we want to revert that particular commit (https://github.com/brianc/node-postgres/commit/fcd53f93a9b3a68814d1c1a4d202c0001d1a2b9a)

Closes #2855

PR #3006, opened 2023-06-12
Remove await from client release

filipecorrea:
The `client.release()` does not return a promise.


abenhamdine (approved):

charmander:
probably more to fix from https://github.com/brianc/node-postgres/pull/2932#issuecomment-1570544235 if anyone wants to take a look

shishkin:
And maybe update types in [DefinitelyTyped/types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/pg/index.d.ts#L274) as well.

Issue #2980, opened 2023-05-17
Bug: 0.8.11 - Return outside of function


bryantqo:
Starting in version 8 a line was added to which a return happens outside of a function. I think this was added in 5532ca5 located in [packages/pg/lib/crypto/utils.js](https://github.com/brianc/node-postgres/commit/5532ca51db96f3370faf66d9e13f0ba226844f62#diff-ac2431edd3fdcb4a431a0675fd1c2e26fbe6900eb523f70efa15ec7211577aa9). This recently popped up in our jest tests with an error of

SyntaxError: /node_modules/pg/lib/crypto/utils.js: 'return' outside of function. (7:2)

Reverting to 8.10.0 solves the problem for now

Jolg42:
I also noticed this today in a renovate PR upgrade of pg package.

charmander:
This can be changed in pg for everyone’s convenience, but that sounds like a bug in whatever tool isn’t recognizing a legitimate feature of CommonJS modules.

darrinholst:
Just ran into this on an esm project that we were still using jest for. You have to jump through some hoops to get jest working with esm. Switching to vitest fixed it for us.

lucgagan:
Confirmed that pinning version v8.10.0 fixes the issue.

For what it is worth, Next.js team is refusing to fix this (CC @shuding) calling this `pg` issue – making `pg` unusable with Next.js.

gronka:
@lucgagan
How do you pin versions? I added this to package.json and ran `rm -rf node_modules && npm install`
```
	"overrides": {
		"node-postgres": "8.10.0"
	}
```

orlein:
I just commented out the `if` clause using `patch-package` because I'm using node v18.12.0
1. comment out the problematic code in `node_modules/pg/lib/crypto/utils.js`
2. run `npx patch-package pg`
3. add `"postinstall": "patch-package"` inside `package.json`'s `scripts`
4. run `yarn add -D patch-package`

and your project will automatically comment out the if clause whenever you and your colleague run `yarn install`.

check out for further info: https://www.npmjs.com/package/patch-package

marinamashina:
We ran into the same issue but on 8.5.1 while this was not happening before. Looking for any practical explanation and solution to this

onethread:
> This can be changed in pg for everyone’s convenience, but that sounds like a bug in whatever tool isn’t recognizing a legitimate (and not infrequently used) feature of CommonJS modules.

[...]

PR #3012, opened 2023-06-16
lazy load crypto utils

hloughrey:
Fix [2980](https://github.com/brianc/node-postgres/issues/2980)

### DIFF
--- a/packages/pg/lib/crypto/sasl.js
+++ b/packages/pg/lib/crypto/sasl.js
@@ -1,5 +1,13 @@
 'use strict'
-const crypto = require('./utils')
+
+let crypto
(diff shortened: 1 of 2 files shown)


charmander (approved):

brianc:
Thank you!

brianc:
hmmm tests are failing in CI

mschilde:
Wouldn't the fix be simply removing the return outside the function? Or wrap everything after the return into an else block? Currently pg v.8.11+ is broken for a number of users.

charmander:
> Wouldn't the fix be simply removing the return outside the function?

no

> Or wrap everything after the return into an else block?

sure

> let's

pr?

charmander (changes requested):
crypto/utils is also imported by lib/client.

PR #3033, opened 2023-07-18
Remove early return for non commonjs environments

benjreinhart:
Addresses https://github.com/brianc/node-postgres/issues/2980.

PR #3040, opened 2023-08-03
fix: favor `globalThis.crypto` over `require('crypto')`

styfle:
In runtimes outside of Node.js, `require()` is not defined.

So we should favor `globalThis.crypto` and only fallback to `require('crypto')` for older versions of Node.js

https://nodejs.org/api/globals.html#crypto_1

This is following up on the work from PR #3033.

### DIFF
--- a/packages/pg/lib/crypto/utils-webcrypto.js
+++ b/packages/pg/lib/crypto/utils-webcrypto.js
@@ -1,5 +1,3 @@
-const nodeCrypto = require('crypto')
-
 module.exports = {
   postgresMd5PasswordHash,
   randomBytes,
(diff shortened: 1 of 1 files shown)


charmander:
Considering that pg uses `require` everywhere else (with the recent Cloudflare Workers changes expecting stubs), does this help concretely?

styfle:
Ah yes, I should have clarified that bundlers can handle local `require('./path/to/file.js')` because they are combining all the source files into one. But they typically leave `require('crypto')`, `require('fs')`, etc since those are not source files but built in modules for Node.js runtime.

charmander:
events, fs, and net, I mean. Which build environment does this affect?

styfle:
Good point. I'll close this then 👍

PR #3043, opened 2023-08-11
Option to use pre-shaped result rows; fixes #3042

### LINKED ISSUE #3042: Result rows are dynamically shaped; causing massive performance degradations when used
I have been chasing performance within our applications this week; one of the things I encountered yesterday was that cloning a quite huge amount of data (~500mb), which are actually simple objects returned from node-postgres, blocked the thread for seconds; in itself logical because JSON.parse(JSON.stringify(row)) was being used. I migrated to using the spread operator instead and that should've improved things by a factor of 10-20; it however only worsened the situation.

Upon further examination and reproduction I found the root-cause of the problem; result rows created by node-postgres are dynamically created; one starts with an empty object; and for every result; every field is dynamically created on that object. Turns out, that Javascript engines don't like this because these objects have a non-simple shape. For more info: https://dev.to/fromaline/why-does-dynamically-adding-properties-is-slow-in-javascript-4hm8 / https://youtu.be/5nmpokoRaZI?t=1180

In order to resolve this I have created a patch that, if one passes "usePrebuiltEmptyResultObjects" to the query constructor; will create the row dynamically once; and then use clones of that to create new rows. This result in a 'clean' shape, and massive performance gains when using the rows returned from node-postgres down the line.

I will create a pull-request, including a test-class shortly, which demonstrates that this increases cloning these objects is at least 50x faster; I actually measure more in the range of 150x; but lets leave some room for wiggle.

I would actually prefer to opt to not have this be an option; but standard behaviour; but I would say that is up to the maintainer.

koenfaro90:
Adds option usePrebuiltEmptyResultObjects on the Query class; this results in generating pre-shaped result row objects instead of dynamically generated ones - massively increasing performance.

I had some difficulties running all tests; also not quite sure about style/backward compatibility demands, so let me know what needs adjusting before LGTM!


brianc:
ohhhh that's super interesting & doesn't seem to introduce any unwanted weird behavior or dynamic class generation or memory leaks or anything. Exciting! I'll run this in CI & see what we can see. 😄

koenfaro90:
I managed to run all tests at one point and everything passed, found one
issue with multiple results which had an obvious point to fix.

brianc:
All tests pass on all versions of node! FUN! I'm going to pull down your branch and run my horribly crude benchmark against it and current release & i'll report back.

packages/pg/lib/result.js
```diff
@@ -59,8 +59,7 @@ class Result {
     return row
   }
 
-  parseRow(rowData) {
```
charmander, line 62: I don’t know if this API change is considered important, but how about prebuilding in `Result`’s constructor? I think it’s cleaner anyway (`Result` is already responsible for `rowMode`, for example), and then `parseRow`’s signature would be able to stay exactly the same. Also, no lazy-building branch if it prebuilds unconditionally in the constructor, no potentially unwanted shared state between the results of multi-result queries, and we could even drop `} else { row[field] = null }` below.

charmander (changes requested):

packages/pg/lib/result.js
koenfaro90: Regarding moving logic to Result; Result is initialized from the constructor of Query; I think field data is only available after handleRowDescription is called; so then I would modify/append to Result after it being constructed - in my eyes I would potentially introduce some possibilities of failure of edge-cases by doing this in another place. If you are sure that it won't  I can try, but I would first have to get my test-setup running steadily. 

I had the null omission in my original patch which I didn't pull, if we are not making it optional then that makes sense indeed.

koenfaro90: Oh; I can just make Result responsible for having a template row instead of query; that makes sense, working on it!

charmander:
I’d also be interested in how `Object.fromEntries` (+ `Object.defineProperty` polyfill for Node 10) benchmarks, since it avoids the (niche for pg, but uncomfortable) `__proto__` hazard and is enabled by this change, but that’s for a future PR anyway!

packages/pg/lib/result.js
koenfaro90: Pushed a commit which moves all the logic to Result; which is indeed even more clean! Also removed the null part

charmander: Nice. Right, sorry, no row description yet in the constructor – but how about in `addFields`? In the same loop as the parser assignment, even.

koenfaro90: I read over addFields as being potentially useful, as normally I would name what it does as 'setFields'; but indeed, this seems like the best place to avoid an extra check in the often-called-function. Committed

charmander (approved):

koenfaro90:
> I’d also be interested in how `Object.fromEntries` (+ `Object.defineProperty` polyfill for Node 10) benchmarks, since it avoids the (niche for pg, but uncomfortable) `__proto__` hazard and is enabled by this change, but that’s for a future PR anyway!

You mean for constructing the initial cloneable-object? As that's called only once I would estimate that will be negligible in terms of performance regardless, but might indeed provide protection against a polluted prototype; however, that same risk was present in the current code, so indeed another PR for another day!

koenfaro90:
> All tests pass on all versions of node! FUN! I'm going to pull down your branch and run my horribly crude benchmark against it and current release & I'll report back.

Does your benchmark include manipulating/using the result-rows? That's where the major gain would be, not so much in result construction I think, maybe a little. If you cannot quantify it, then I will write a small benchmark for that next week.

koenfaro90:
Seems the CI had a hickup, or is something actually failing now?

koenfaro90:
I wrote a little bench quickly; https://github.com/koenfaro90/node-postgres-bench/tree/master; there seems to be a tiny performance regression SOMEWHERE else, the actual operations seem faster, but somewhere else we seem to loose a little ms - I will investigate on Monday where the regression is caused.

> 8.11.2
< 8.11.2; 9952ms
> pg-performance-pull
< pg-performance-pull; 10278ms
AccessCase { '8.11.2': 39.66, 'pg-performance-pull': 11.47 }
CloneUsingAssignCase { '8.11.2': 164.18, 'pg-performance-pull': 18.52 }
CloneUsingSpreadCase { '8.11.2': 179.6, 'pg-performance-pull': 3.2 }
KeysCase { '8.11.2': 17.12, 'pg-performance-pull': 0.83 }
ValuesCase { '8.11.2': 76.62, 'pg-performance-pull': 8.97 }

brianc:
Nice - I'm assuming on that output above a lower number is better?  Should I be paying attention to the `AccessCase` and `CloneUsingAssignCase` etc or only the diff in numbers between `8.11.2` and `pg-performance-pull`?

It's always kinda cat and mouse game trying to benchmark the driver since postgres may take slightly longer or shorter to do a few queries, node might do a GC pause here or there, network, disk, etc all come into play.  I'd recommend trying to run the bench like 10 times & seeing if you steadily see more or less performance from your branch, because in my experience it's not extremely steady numbers....but if actually accessing the row values by name is actually faster, and that speed up is somewhat linearly related to the number of rows returned, this is gonna be yuge!

koenfaro90:
So, the times you see are all in ms, but the total numbers include setup
(generating rows, inserting them), so the actual row usage is waaay faster
indeed. Both cloning and simply accessing. But the total time including
setup, inserting, querying and executing the specific is slower, perhaps
cloning of the row accounts for this, or addFields is called more than
once. I will research what accounts for this on Monday.

Cases run in a seperate node instance to prevent any internal
optimalization or GC impacting the results. So for now I would hold back on
the merge until this is resolved.

koenfaro90:
Found the issue; needed to ensure that the clone-base record was properly shaped; now the new version is always considerably faster on all fronts (9600ms vs 10000ms for the suite).

koenfaro90:
I just messaged a related idea to one of my colleagues, focusing on being
able to perform all JS side preparation/serialization while a previous
query is being executed, most gains would be in bulk inserts with a certain
chunk size. But will point him to your issue too.

brianc:
Weirdly tests failed on a few versions of node...rerunning them. Looking forward to merging this!!

Issue #3055, opened 2023-09-01
Parsing single result with large amount of columns takes more time in 8.11.3


wannabehero:
Most likely is the direct consequence of #3043 

After upgrading to 8.11.3 we've started to experience the increased latency for one of our queries:
<img width="426" alt="image" src="https://github.com/brianc/node-postgres/assets/1236794/9204eeea-fa78-4238-9c98-404cee7df21a">
The graph shows latency at 8.11.2, then 8.11.3 (increased), and then after we rolled back to 8.11.2. No other dependencies were changed. We also use typeorm on top of pg.

This query fetched 1 row but with ~600 columns. All other queries were not affected this much.

We isolated the issue to the query call (pganalyze didn't show any difference in the actual query time, also nothing else was changed in the code).

What we believe is happening: because of the extra loop for pre-building an empty object it adds some execution time.

We've prepared some sample code to showcase the issue: https://gist.github.com/wannabehero/a6531df714041e559a38c97d49f2f444 

```sh
➜  pg-test node test.js
8.11.3: 164.089ms
8.11.2: 80.921ms
```

Do you think it's valid? If so, can we get back to having an option to enable this behavior? Like it was in https://github.com/brianc/node-postgres/pull/3043/commits/11955a0ad2b2f2687988248c4c995be4cc2bb622

Thanks.

brianc:
@wannabehero oh yup I could see an additional loop slowing it down.  You wanna test out the PR I put up on #3056 ? It removes the additional loop (we were already looping on `rowDescription` anyway, so I just use the same loop now).

charmander:
A difference of tens of milliseconds (in the graph), for making an object with just 600 `null` properties? Is that an aggregate time for multiple runs of this one-row query?

wannabehero:
@charmander we use datadog to collect pg metrics that I included in the issue.

Datadog wraps the query method
https://github.com/DataDog/dd-trace-js/blob/master/packages/datadog-instrumentations/src/pg.js

And the only difference between 8.11.2 and 8.11.3 is the extra loop and the pre-built object.

wannabehero:
@brianc thanks for the swift response❤️

We should be able to validate early next week. I'll get back as soon as we do

brianc:
Awesome! Thanks!

wannabehero:
Hi @brianc 

I run the same small experiment over the updated code from #3056: https://gist.github.com/wannabehero/0ec140e764413624a89004af13de7fc2

[...]

PR #3056, opened 2023-09-01
Remove 1 loop on rowDescription event

brianc:
Addresses #3055

This is functionally the same but avoids the 1 extra loop being added by building the prebuilt results within the same loop already used to build other cached pieces of the result parser when it receives the `rowDescription` event.

### DIFF
--- a/packages/pg/lib/result.js
+++ b/packages/pg/lib/result.js
@@ -98,7 +98,7 @@ class Result {
         this._parsers[i] = types.getTypeParser(desc.dataTypeID, desc.format || 'text')
       }
     }
-    this._prebuiltEmptyResultObject = row
+    this._prebuiltEmptyResultObject = { ...row }
   }
 }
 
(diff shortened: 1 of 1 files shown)


packages/pg/lib/result.js
```diff
@@ -85,22 +85,20 @@ class Result {
...
     }
-    this._createPrebuiltEmptyResultObject()
-  }
-  _createPrebuiltEmptyResultObject() {
-    var row = {}
-    for (var i = 0; i < this.fields.length; i++) {
-      row[this.fields[i].name] = null
-    }
-    this._prebuiltEmptyResultObject = { ... row }
+    this._prebuiltEmptyResultObject = row
```
charmander, line 101: Note: IIRC

```suggestion
    this._prebuiltEmptyResultObject = {...row}
```

was important in #3043, though I haven’t actually run the benchmarks.

huntersgordon: nice

Issue #3032, opened 2023-07-14
nwjs or chromium's recent URL class only decodes known protocols, makes connection string postgres://... useless 


chhu:
Probably a chromium problem, pure node works. Tried with recent 0.77.0 

Don't even know if that should be rather reported to nwjs or chromium... Just to let folks know.

![image](https://github.com/brianc/node-postgres/assets/208672/550f88d6-e9f0-4641-af56-0983f48266dd)

brianc:
Thanks for the heads up. Not sure what nwjs is - but yeah that's a bit annoying it doesn't post the url w/o the protocol being set to http/https

hjr3:
Is this still an issue? I see it working in Chrome `Version 135.0.7049.85 (Official Build) (arm64)`

<img width="715" alt="Image" src="https://github.com/user-attachments/assets/8aca1051-9150-4d30-b771-d1b83cdc53f2" />

PR #3061, opened 2023-09-06
fix: conflict between browser URL object and Node URL object

billybonks:
I am running this package using electron, what i noticed was that due to the fact that the lines between node and browser environments become a bit blurred, the URL class that was being used was the one defined by the browser and not node. By making an explicit require it ensures the correct Class is used.

While creating a test for this would be difficuilt i think adding an eslint rule to stop using globally defined objects and require imports instead would resolve issues like this in the future

### DIFF
--- a/packages/pg-connection-string/index.js
+++ b/packages/pg-connection-string/index.js
@@ -1,9 +1,9 @@
 'use strict'
-
 //Parse method copied from https://github.com/brianc/node-postgres
 //Copyright (c) 2010-2014 Brian Carlson (brian.m.carlson@gmail.com)
 //MIT License
 
+const { URL } = require('url')
 //parses a connection string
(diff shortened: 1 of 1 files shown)


charmander:
What impact does the choice of `URL` implementation have in Electron?

billybonks:
@charmander if it uses the node version of URL it works great

image

and if you run the same in the browser it does not get parsed in the same way

image


which causes the connection string code not to work.

boromisp:
Probably related: https://github.com/brianc/node-postgres/issues/3032

brianc:
cool this seems good to me!

zacharyrankin:
What would be the best way for me to get these changes right now? I am running into this same issue right now.

Thanks so much

PR #3063, opened 2023-09-07
Fix get value of last column with same name in result rows

### LINKED ISSUE #3062: With duplicated column names, result row gets value from rightmost non-null column (or default null)
Noticed when upgraded to 8.11.3 that the parsing of duplicated column names has changed, and broke some of our queries.
I made a minimal example of situation:
```
const res1 = await client.query('SELECT NULL AS test, 10 AS test, NULL AS test')
const res2 = await client.query('SELECT NULL AS test, 10 AS test, 12 AS test')
console.log(res1.rows[0].test)
console.log(res2.rows[0].test)
```

With 8.11.3, code above prints
```
10
12
```
And with 8.11.2, it prints
```
null
12
```

For me, it should be consistent, either leftmost or rightmost column, but never some kind of "last non-null"

I also checked code, and this is caused by #3043 dropping else-branch https://github.com/brianc/node-postgres/pull/3043/files#diff-e7208163de109ecdddff33a5473cb7c255bd7693a1ec06da8300fe88cabf7a4fL70 here

sehrope:
First commit adds a failing test for the issue and the second one fixes the issue by explicitly setting a null value in the else block (like before). I think we'll still get most of the gains from prebuilt object patch as the shape is still not dynamic and I'd bet V8 is smart enough to optimize setting a value to itself (i.e. null to null).

Here's the failing test without the patch: https://github.com/sehrope/node-postgres/actions/runs/6112124966/job/16588908962#step:7:677

Fixes #3062

### DIFF
--- a/packages/pg/lib/result.js
+++ b/packages/pg/lib/result.js
@@ -67,6 +67,8 @@ class Result {
       var field = this.fields[i].name
       if (rawValue !== null) {
         row[field] = this._parsers[i](rawValue)
+      } else {
+        row[field] = null
       }
     }
     return row
--- a/packages/pg/test/integration/gh-issues/3062-tests.js
+++ b/packages/pg/test/integration/gh-issues/3062-tests.js
@@ -1,0 +1,21 @@
+'use strict'
+const helper = require('../test-helper')
+var assert = require('assert')
+const suite = new helper.Suite()
+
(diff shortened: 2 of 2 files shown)


charmander (approved):

brianc:
Thank you!

PR #3073, opened 2023-10-05
suggest to use URL constructor in pg-pool readme

komachi:
improve example with parsing connection string for pg-pool, use URL constructor, available since node 10

https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

### DIFF
--- a/packages/pg-pool/README.md
+++ b/packages/pg-pool/README.md
@@ -52,17 +52,16 @@ The Pool constructor does not support passing a Database URL as the parameter. T
 
 ```js
 const Pool = require('pg-pool');
-const url = require('url')
 
-const params = url.parse(process.env.DATABASE_URL);
-const auth = params.auth.split(':');
+
+const databaseUrl = new URL(process.env.DATABASE_URL);
 
 const config = {
-  user: auth[0],
-  password: auth[1],
-  host: params.hostname,
-  port: params.port,
-  database: params.pathname.split('/')[1],
+  user: databaseUrl.username,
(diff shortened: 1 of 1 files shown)


charmander:
I’m not sure why this is documented this way in the first place instead of using the `connectionString` option…

usahai:
Is there a way to extend the client for pooled connections? I need to use a schema, but the existing type definitions does not accept a schema, unless I pass it in as a part of the URL in the client. However, it's not extensible to pooled connections.

So this is fine for client - 
```
const client: ClientConfig = new Client(
  process.env.DB_URL,
  //   {
  //   host: process.env.DB_HOST,
  //   port: process.env.DB_PORT as unknown as number,
  //   database: process.env.DB_DATABASE,
  //   user: process.env.DB_USER,
  //   password: process.env.DB_PASSWORD,
  // }
)
```

And I was thinking maybe if there was a provision to do something like this - 
```
const pool: Pool = new Pool({
  ...client,
  max: 20,
  idleTimeoutMillis: 60000,
  connectionTimeoutMillis: 20000,
})
```
wherein the URL will get deconstructed on it's own (overly hopeful, I know)
OR
```
const pool: Pool = new Pool(client, {
  max: 20,
  idleTimeoutMillis: 60000,
  connectionTimeoutMillis: 20000,
})
```
where the additional parameters get attached to the Pool object.

charmander:
@usahai As I mentioned in the previous comment, use the `connectionString` pool option to pass a connection string.

PR #3074, opened 2023-10-06
feat: allow specifying a timeout on a per query base

srieding:
It is the same as https://github.com/brianc/node-postgres/pull/2655 but with the requested integration test added in.


charmander:
(As the test shows, needs a corresponding native client change.)

charmander (approved):

PR #3077, opened 2023-10-11
eslintrc: fix prettier base config

alxndrsn:
Potential fix for https://github.com/brianc/node-postgres/issues/3076

Ref https://github.com/prettier/eslint-config-prettier/blob/main/CHANGELOG.md#version-800-2021-02-21

PR #3078, opened 2023-10-11
Fix eslint violations

alxndrsn:
Follow-up to https://github.com/brianc/node-postgres/pull/3077, using `eslint ... --fix`.


brianc (approved):
Thanks for doing this!

PR #3093, opened 2023-11-06
fix: Precision loss on extremely large floating-point numbers

### LINKED ISSUE #3092: Precision loss on numbers close to MAX_SAFE_INTEGER on Postgres < 12
To reproduce:

```js
const assert = require("assert");
const { Client } = require("pg");
const pg = new Client("postgres://prisma:prisma@localhost:5432/tests");

async function main() {
  await pg.connect();

  try {
    const result = await pg.query({
      text: "SELECT $1::FLOAT8",
      values: [Number.MAX_SAFE_INTEGER],
    });
    assert.strictEqual(result.rows[0].float8, Number.MAX_SAFE_INTEGER);
  } finally {
    await pg.end();
  }
}

main();
```

If used with Postgres < 12, this fails with:

```
AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
+ actual - expected

+ 9007199254740990
- 9007199254740991
```

Postgres 12+ is fine.

This value can be correctly stored in both JS and Postgress without precision loss. I believe the problem is default `extra_float_digits` settings: pre-12, postgres rounded floats to 15 digits when outputting them as text, starting from 12 they switched default to outputting as many digits as necessary to preserve the precision. Setting `extra_float_digits` to `1` fixes the problem.

There was a similar issue, #730 that is marked as fixed. Might just not be the case for extremely large numbers.

SevInf:
Sets `extra_float_digits` setting to 1 when connecting to DB. For older
(<12) Postgres, this should be enough to cover all values JS can store
preciesely. For 12+ postgres this setting should have no effect.

Fix #3092

### DIFF
--- a/packages/pg-protocol/src/serializer.ts
+++ b/packages/pg-protocol/src/serializer.ts
@@ -26,6 +26,7 @@ const startup = (opts: Record<string, string>): Buffer => {
   }
 
   writer.addCString('client_encoding').addCString('UTF8')
+  writer.addCString('extra_float_digits').addCString('1')
 
   var bodyBuffer = writer.addCString('').flush()
   // this message is sent without a code
--- a/packages/pg/test/integration/client/type-coercion-tests.js
+++ b/packages/pg/test/integration/client/type-coercion-tests.js
@@ -108,7 +108,7 @@ var types = [
   },
   {
     name: 'double precision',
-    values: [-101.3, -1.2, 0, 1.2, 101.1, null],
+    values: [-101.3, -1.2, 0, 1.2, 101.1, Number.MAX_SAFE_INTEGER, null],
   },
   {
     name: 'timestamptz',


packages/pg-protocol/src/serializer.ts
```diff
@@ -26,6 +26,7 @@ const startup = (opts: Record<string, string>): Buffer => {
   }
 
   writer.addCString('client_encoding').addCString('UTF8')
+  writer.addCString('extra_float_digits').addCString('1')
```
charmander, line 29: I think this belongs in pg’s use of `startup` instead of in pg-protocol. (Side benefit: fewer/no other tests to update, since that’s at least part of why CI is failing right now.)

It also needs to be higher than 1 (take `0.42467013940651843` as a random example), with an accordingly stricter test case. I don’t have time right now to figure out exactly how much precision is enough, but setting it to 3 (https://www.postgresql.org/docs/11/runtime-config-client.html#GUC-EXTRA-FLOAT-DIGITS) is probably fine.

charmander (changes requested):

PR #3102, opened 2023-11-27
Fix typo

reztib:
Pronoun 'its' was written in uppercase ('Its'); fixed that


shugyoza:
It is `lts` that stands for `Long Term Support` version. Updating it into `LTS` would be more appropriate as it is an abbreviation.

PR #3110, opened 2023-12-13
chore: fix the CI by bumping eslint and its dependencies

ghusse:
The current linting script is broken in the CI because of the recent bump of `prettier` that does not seem compatible with the installed version of eslint.

This PR upgrades eslint and its plugins as well, and fixes introduced linting issues on 2 files.

It fixes the current linting script and allows the CI to pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,14 +20,14 @@
     "lint": "eslint '*/**/*.{js,ts,tsx}'"
   },
   "devDependencies": {
-    "@typescript-eslint/eslint-plugin": "^4.4.0",
-    "@typescript-eslint/parser": "^4.4.0",
-    "eslint": "^7.11.0",
-    "eslint-config-prettier": "^8.5.0",
+    "@typescript-eslint/eslint-plugin": "^6.14.0",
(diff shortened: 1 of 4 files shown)


brianc:
ironically the CI run failed here 🙃  Regardless...I appreciate your effort & help here. I'll get this running. Maybe there's a race condition in the test or something? Re-running now. ❤️

ghusse:
Thanks!

abenhamdine:
@brianc we need to merge this, because currenty every PR is failing in CI because of the said issue

alxndrsn:
Closes https://github.com/brianc/node-postgres/issues/3147

brianc:
thanks! I merged another fix for this today - sorry for missing this! I will be more on top of stuff going forward _for sure_. Been dealing w/ a lot of work stuff as I spin my current role down.

brianc:
I think this might be okay to close now (though the effort is much appreciate) as CI is now green?

Its unfortunate how some version bumps caused such a catastrophe. Espehs while I was in the midst of a bit of a life change situation. Will stay more on top of that in the future 🔨

PR #3114, opened 2023-12-26
Clarify usage and update readme

irvinebroque:
refs https://github.com/Ethan-Arrowood/socket/pull/17

@petebacondarwin @brianc — publishing `pg-cloudflare` as a separate package seems to be really helpful to people. Ex:

https://github.com/sidorares/node-mysql2/pull/2289/files#diff-e56fabfb5e90fd8f6265cfbe84f3701a85261d884e198bf61de34958cee4864aR12

Added some docs to clarify usage, and cross link to the Node.js implementation of the Socket API.


petebacondarwin (approved):
LGTM.

brianc:
i dig it! Lint failing for some reason? On nothing but a markdown change? what in the hay hay?

abenhamdine:
> i dig it! Lint failing for some reason? On nothing but a markdown change? what in the hay hay?

@brianc looks like Lint is failing in CI in every PR with this error :

```
TypeError: prettier.resolveConfig.sync is not a function
````

brianc:
> looks like Lint is failing in CI in every PR with this error :

yeah annoying transitive dep break. Fixed on `master` if you wanna rebase. If you don't, I'll try rebasing for ya in a few days

brianc:
> in a few days

like probably not for a week or two...but eventually I will! 😅

brianc:
> like probably not for a week or two...but eventually I will! 😅

better late than never! thanks again!

PR #3128, opened 2024-01-17
feat(pg-connection-string): ClientConfig helper functions

### LINKED ISSUE #2280: pg-connection-string ConnectionOptions interface is not compatible with PoolConfig

hjr3:
Two new functions are introduced to make it easy for TypeScript users to use a PostgresSQL connection string with pg Client.

Fixes #2280

### DIFF
--- a/packages/pg-connection-string/README.md
+++ b/packages/pg-connection-string/README.md
@@ -35,6 +35,27 @@ The resulting config contains a subset of the following properties:
   * `ca`
 * any other query parameters (for example, `application_name`) are preserved intact.
 
+### ClientConfig Compatibility for TypeScript
+
+The pg-connection-string `ConnectionOptions` interface is not compatible with the `ClientConfig` interface that [pg.Client](https://node-postgres.com/apis/client) expects. To remedy this, use the `parseIntoClientConfig` function instead of `parse`:
+
+```ts
+import { ClientConfig } from 'pg';
+import { parseIntoClientConfig } from 'pg-connection-string';
+
+const config: ClientConfig = parseIntoClientConfig('postgres://someuser:somepassword@somehost:381/somedatabase')
+```
+
+You can also use `toClientConfig` to convert an existing `ConnectionOptions` interface into a `ClientConfig` interface:
+
+```ts
+import { ClientConfig } from 'pg';
+import { parse, toClientConfig } from 'pg-connection-string';
+
+const config = parse('postgres://someuser:somepassword@somehost:381/somedatabase')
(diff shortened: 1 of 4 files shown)


hjr3:
I will rebase once https://github.com/brianc/node-postgres/pull/3122 lands

packages/pg-connection-string/test/toClientConfig.js
```diff
@@ -0,0 +1,94 @@
...
+      cert: 'example cert\n',
+      key: 'example key\n',
+    })
+  })
+
+  it('converts unix domain sockets', function () {
+    const config = parse('socket:/some path/?db=my[db]&encoding=utf8&client_encoding=bogus')
+    const clientConfig = toClientConfig(config)
+    clientConfig.host.should.equal('/some path/')
+    clientConfig.database.should.equal('my[db]', 'must to be escaped and unescaped through "my%5Bdb%5D"')
```
ThisIsMissEm, line 91: This feels like it should've thrown an error?

hjr3: this is current behavior according to: https://github.com/brianc/node-postgres/blob/df0f4d19fbf0349c6b41e8982e058f7377ee3bea/packages/pg-connection-string/test/parse.js#L43

hjr3:
@charmander please review this when you have a moment. This is related to https://github.com/brianc/node-postgres/pull/2709

hjr3:
@brianc what can we do to land `pg-connection-string` PRs?

packages/pg-connection-string/test/toClientConfig.js
ThisIsMissEm: huh, okay!

PR #2105, opened 2020-02-18
fix: update pg-connection-string

revolunet:
Hi,

Use a recent [pg-connection-string](https://github.com/iceddev/pg-connection-string) release that handle much more cases (ex: ssl)

Thanks

PR #3129, opened 2024-01-17
chore: add codeowners file

hjr3:
This makes it clear who is maintaining which package. This will also allow commit privileges for a specific package instead of the entire repo.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+/packages/pg-connection-string @hjr3


hjr3:
From https://github.com/brianc/node-postgres/pull/2105#issuecomment-589179652, I was maintaining `pg-connection-string` but my commit access has been removed.

@brianc what do you think about this approach?

brianc:
it makes sense to me - sorry the build was broken. might wanna rebase on it!

hjr3:
@brianc all good now

PR #3133, opened 2024-01-23
Upgrade pg-types to 4.0.2 for performance improvements

cesco69:
pg-types 4.0.2 include performance improvements from https://github.com/brianc/node-pg-types/pull/148


sisou:
This will also fix a Node 20 depreciation warning: `DeprecationWarning: Buffer() is deprecated due to security and usability issues.`

`new Buffer()` is used in `postgres-bytea`, which is depended on by `pg-types`. The buffer usage was updated in https://github.com/bendrucker/postgres-bytea/commit/2317d6d6be2f962d74c378672cea6e65a7c006a3, which went into release `2.0.0`, which `pg-types` was updated to only in version `3.0.0` (https://github.com/brianc/node-pg-types/commit/c605c7cce1bf65d3deeaba07a62ad7a6842a1b21).

brianc:
I want to do this, but pg-types upgrading to 4.x is a breaking change for node-postgres (date handling is different). I want to make some other changes & bundle it all up w/ a 9.0 release for pg so its gonna take some time...but its on my list.

PR #3149, opened 2024-02-19
chore (ci): upgrade github actions to use node 20

alxndrsn:
Fixes:

> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3, actions/setup-node@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.


brianc:
Heeey sorry about that - this should be fixed now. Some crazy dependency update stuff happened which broke the build for a bit. Thankfully we got it solved! if you rebase on master we should be good to merge!

brianc:
Thanks so much for your help!

PR #3157, opened 2024-02-26
Fix build - package version upgrade for build/lint related packages.

brianc:
No functional changes in here. CI broke due to some minor or patch version incomaptibility change between eslint, its army of plugins and configs, prettier, typescript, etc. 🔥


charmander:
Related: #3122, #3110

brianc:
closed by #3122  YAY

PR #3170, opened 2024-03-13
refactor: tighten up cloudflare detection

petebacondarwin:
The previous approach to detecting whether to use Cloudflare's sockets was to check for missing polyfills.
But as we improve the polyfills that Wrangler can provide these checks are no longer valid.

Now we check if the `navigator.userAgent` is `'Cloudflare-Workers'` and fallback to Node.js if not.

**This is a refactoring - there is already a test in place to check that this works.**

### DIFF
--- a/packages/pg/lib/stream.js
+++ b/packages/pg/lib/stream.js
@@ -1,37 +1,54 @@
-let isCloudflareRuntime
+const { getStream, getSecureStream } = getStreamFuncs()
+
+module.exports = {
+  /**
+   * Get a socket stream compatible with the current runtime environment.
+   * @returns {Duplex}
+   */
+  getStream,
+  /**
+   * Get a TLS secured socket, compatible with the current environment,
+   * using the socket and other settings given in `options`.
+   * @returns {Duplex}
+   */
+  getSecureStream,
+}
 
 /**
- * Get a socket stream compatible with the current runtime environment.
- * @returns {Duplex}
+ * The stream functions that work in Node.js
  */
-module.exports.getStream = function getStream(ssl) {
-  if (isCloudflareRuntime === undefined) {
-    isCloudflareRuntime = computeIsCloudflareRuntime()
-  }
-  if (isCloudflareRuntime) {
-    const { CloudflareSocket } = require('pg-cloudflare')
-    return new CloudflareSocket(ssl)
-  } else {
+function getNodejsStreamFuncs() {
+  function getStream(ssl) {
     const net = require('net')
     return new net.Socket()
+  }
+
+  function getSecureStream(options) {
+    var tls = require('tls')
+    return tls.connect(options)
+  }
+  return {
+    getStream,
+    getSecureStream,
   }
 }
 
 /**
- * Get a TLS secured socket, compatible with the current environment,
- * using the socket and other settings given in `options`.
- * @returns {Duplex}
+ * The stream functions that work in Cloudflare Workers
  */
-module.exports.getSecureStream = function getSecureStream(options) {
-  if (isCloudflareRuntime === undefined) {
-    isCloudflareRuntime = computeIsCloudflareRuntime()
+function getCloudflareStreamFuncs() {
+  function getStream(ssl) {
+    const { CloudflareSocket } = require('pg-cloudflare')
+    return new CloudflareSocket(ssl)
   }
-  if (isCloudflareRuntime) {
+
+  function getSecureStream(options) {
     options.socket.startTls(options)
     return options.socket
-  } else {
-    var tls = require('tls')
-    return tls.connect(options)
+  }
+  return {
+    getStream,
+    getSecureStream,
   }
 }
 
@@ -40,7 +57,7 @@ module.exports.getSecureStream = function getSecureStream(options) {
  *
  * @returns true if the code is currently running inside a Cloudflare Worker.
  */
-function computeIsCloudflareRuntime() {
+function isCloudflareRuntime() {
   // Since 2022-03-21 the `global_navigator` compatibility flag is on for Cloudflare Workers
   // which means that `navigator.userAgent` will be defined.
   if (typeof navigator === 'object' && navigator !== null && typeof navigator.userAgent === 'string') {
@@ -54,4 +71,11 @@ function computeIsCloudflareRuntime() {
     }
   }
   return false
+}
+
+function getStreamFuncs() {
+  if (isCloudflareRuntime()) {
+    return getCloudflareStreamFuncs()
+  }
+  return getNodejsStreamFuncs()
 }
--- a/packages/pg/lib/stream.js
+++ b/packages/pg/lib/stream.js
@@ -1,14 +1,19 @@
+let isCloudflareRuntime
+
 /**
  * Get a socket stream compatible with the current runtime environment.
  * @returns {Duplex}
  */
 module.exports.getStream = function getStream(ssl) {
-  const net = require('net')
-  if (typeof net.Socket === 'function') {
-    return new net.Socket()
-  } else {
+  if (isCloudflareRuntime === undefined) {
+    isCloudflareRuntime = computeIsCloudflareRuntime()
+  }
+  if (isCloudflareRuntime) {
     const { CloudflareSocket } = require('pg-cloudflare')
     return new CloudflareSocket(ssl)
+  } else {
+    const net = require('net')
+    return new net.Socket()
   }
 }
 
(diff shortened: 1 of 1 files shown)


packages/pg/lib/stream.js
```diff
@@ -3,12 +3,12 @@
  * @returns {Duplex}
  */
 module.exports.getStream = function getStream(ssl) {
-  const net = require('net')
```
sehrope, line 6: The default / first attempt really should be the real socket in node. Trying for `pg-cloudflare` first is weird in that the other 99.99% of usage of this module don't want that code path.

I'd imagine this would do weird things if they happen to have it in their node_modules for other reasons (e.g. they have the dep but are running locally and not on cloudflare). A project could have that module installed or other "cloud node" shim modules for testing against them without actually intending to use any one of them.

Ideally what we need is to have the CF versions of those libraries actually match the node versions (e.g. if CF provides a `net.Socket(...)` it should be usable as a `net.Socket`. Ditto for the TLS stuff. If it just works then none of these shims are required.

Alternatively, come up with a well defined way to signal to the app that it is running on CF. Like checking for an env var.

petebacondarwin: Yeah good call. I think we can use something like:

```ts
process?.release?.name === 'node'
```

packages/pg/lib/stream.js
```diff
@@ -18,8 +18,8 @@ module.exports.getStream = function getStream(ssl) {
  * @returns {Duplex}
  */
 module.exports.getSecureStream = function getSecureStream(options) {
-  var tls = require('tls')
-  if (tls.connect) {
+  if (process.release && process.release.name === 'node') {
+    var tls = require('tls')
```
brianc, line 22: nit: if you want extreme pedantry (I mean...who doesn't, right? 😉 ) putting the conditional (`process.release && process.release.name === 'node') into a predicate function `isNode()` or something would DRY this up a bit.   I'm also slightly concerned how this impacts bun/deno? Is there any way to invert the check and see if you're inside cloudflare instead?  Then if you're not in cloudflare but you're in a fully node compatible environment (be that node, deno, bun) we're using the node side of things.  Obviously this is subject to change as the node/deno/bun/cf/browser/wasm environments continue to converge...just my thoughts here.

sehrope: > Is there any way to invert the check and see if you're inside cloudflare instead?

Yes, that's what I meant by checking if you're in CF (as opposed to checking if you're not in node).

I'm not sure if they expose anything like that. If not, they definitely should add a `process.env.CLOUDFLARE_WEB_WORKER` env similar to how GitHub does for actions.

> Then if you're not in cloudflare but you're in a fully node compatible environment (be that node, deno, bun) we're using the node side of things.

Reminds me of the io.js days!

petebacondarwin: 😢

petebacondarwin: Of course trying to import from `cloudflare:sockets` has exactly this behaviour. It is checking whether we are in a Worker or not.

petebacondarwin: The import of `pg-cloudflare` outside of a worker is an empty package.

petebacondarwin: I also considered a "Cloudflare specific global" such as `HtmlRewriter` which doesn't exist in node or deno... but it does exist in bun!

I am asking internally to see if there is a more reliable way to determine the runtime.

packages/pg/lib/stream.js
```diff
@@ -18,11 +18,20 @@ module.exports.getStream = function getStream(ssl) {
...
   }
 }
+
+/**
+ * Are we running in a Cloudflare Worker?
+ * 
+ * @returns true if the code is currently running inside a Cloudflare Worker.
+ */
+function isCloudflareRuntime() {
+  return Boolean(process.release && process.release.name !== 'node' && Response && new Response(null, { cf: { thing: true } }).cf.thing === true);
```
mrbbot, line 36: I think `&& Response &&` would `ReferenceError` if `Response` was undefined. Using `typeof` should be safe here. Similarly, `cf.thing` will throw a `TypeError` if `cf` is `undefined`.
```suggestion
  return Boolean(process.release && process.release.name !== 'node' && typeof Response === "function" && new Response(null, { cf: { thing: true } }).cf?.thing);
```

petebacondarwin: Thanks. Incorporated this and also tried a new approach.

charmander:
I haven’t taken a good look at pg-cloudflare yet, but I think it’s always nicer to detect the APIs that are actually needed than to detect a particular runtime, when possible.

charmander:
Trying to detect a specific platform based on API presence isn’t an example of that; trying to import `cloudflare:sockets` is, yeah. Cache the result?

packages/pg/lib/stream.js
```diff
@@ -18,11 +18,32 @@ module.exports.getStream = function getStream(ssl) {
...
+    return tls.connect(options)
+  }
+}
+
+/**
+ * Are we running in a Cloudflare Worker?
+ *
+ * @returns true if the code is currently running inside a Cloudflare Worker.
+ */
+function isCloudflareRuntime() {
```
IgorMinar, line 35: can you memoize this function please? otherwise it would run many times unnecessarily.``
```suggestion
const isCloudflareRuntime = (function isCloudflareRuntimeInit() {
```

packages/pg/lib/stream.js
```diff
@@ -18,11 +18,32 @@ module.exports.getStream = function getStream(ssl) {
...
+  }
+  // In case `navigator` or `navigator.userAgent` is not defined then try a more sneaky approach
+  if (typeof Response === 'function') {
+    const resp = new Response(null, { cf: { thing: true } })
+    if (typeof resp.cf === 'object' && resp.cf !== null && resp.cf.thing) {
+      return true
+    }
   }
+  return false
 }
```
IgorMinar, line 49:
```suggestion
})();
```

packages/pg/lib/stream.js
```diff
@@ -3,12 +3,12 @@
  * @returns {Duplex}
  */
 module.exports.getStream = function getStream(ssl) {
-  const net = require('net')
-  if (typeof net.Socket === 'function') {
-    return new net.Socket()
-  } else {
+  if (isCloudflareRuntime) {
```
IgorMinar, line 6: I think you forgot to call the fn here, but that's ok if you accept my memoization changes below

IgorMinar:
I kicked off a discussion at Cloudflare about possibly taking a different approach by beefing up our `node:net#Socket` support: https://github.com/cloudflare/workerd/issues/2129

I do think that this PR improves the situation and should be merged, otherwise we'll most likely break pg-cloudflare as we improve our node compatibility which will result in misdetection in the current version of pg-cloudflare. But ideally, we should transition completely away from a need to provide a custom socket within `pg` as our `node:net#Socket` should be good enough, but that will take a bit of time.

petebacondarwin:
Thanks to everyone who has looked at this PR over the weeks it has been evolving. I think it is now in a good enough place to land. Could you take a last look? @brianc , @IgorMinar , @sehrope and @charmander ?

sehrope:
Having multiple parts of the code handle the caching is odd. Also, that structure makes it more difficult to add other non-native stream implementations.

How about something like this:

```js
function getNodejsStreamFuncs() {
  function getStream(ssl) {
    const net = require('net')
    return new net.Socket()
  }

  function getSecureStream(options) {
    var tls = require('tls')
    return tls.connect(options)
  }
  return {
    getStream,
    getSecureStream,
  }
}

function getCloudflareStreamFuncs() {
  function getStream(ssl) {
    const { CloudflareSocket } = require('pg-cloudflare')
    return new CloudflareSocket(ssl)
  }

  function getSecureStream(options) {
    options.socket.startTls(options)
    return options.socket
  }
  return {
    getStream,
    getSecureStream,
  }
}

/**
 * Are we running in a Cloudflare Worker?
 *
 * @returns true if the code is currently running inside a Cloudflare Worker.
 */
 function isCloudflareRuntime() {
  // Since 2022-03-21 the `global_navigator` compatibility flag is on for Cloudflare Workers
  // which means that `navigator.userAgent` will be defined.
  if (typeof navigator === 'object' && navigator !== null && typeof navigator.userAgent === 'string') {
    return navigator.userAgent === 'Cloudflare-Workers'
  }
  // In case `navigator` or `navigator.userAgent` is not defined then try a more sneaky approach
  if (typeof Response === 'function') {
    const resp = new Response(null, { cf: { thing: true } })
    if (typeof resp.cf === 'object' && resp.cf !== null && resp.cf.thing) {
      return true
    }
  }
  return false
}

function getStreamFuncs() {
  if (isCloudflareRuntime()) {
    return getCloudflareStreamFuncs()
  }
  return getNodejsStreamFuncs()
}

module.exports = getStreamFuncs()
```

When (if?) we ever change this repo to embed types, that lends itself to a better structure as as each of the `getXyzStreamFuncs()` function would have the same signature. It also makes it easier to add something in the future as there's a clear pattern and a single place to do so.

--- 

I haven't tested the above so if you do go with that approach, obviously test it out and make sure it makes sense.

Regarding the check in `isCloudflareRuntime()`, why is there two different checks? In what situation would the first part not trigger but the second part would? Do they still not have a single env variable to indicate you're running in a Cloudflare worker? (e.g., `process.env.CLOUDFLARE_WORKER`)

petebacondarwin:
Thanks for the input @sehrope.

> Regarding the check in isCloudflareRuntime(), why is there two different checks? In what situation would the first part not trigger but the second part would? Do they still not have a single env variable to indicate you're running in a Cloudflare worker? (e.g., process.env.CLOUDFLARE_WORKER)

This is because the `navigator.userAgent` approach does not work if the user has their Cloudflare Workers compatibility date set to older than `2022-03-21`.

I did consider a solution similar to the one you propose above. The only downside is that the check for Cloudflare will not happen lazily. But I guess that it is so fundamental to get hold of such a socket that there are very few (if any) scenarios where the library will be used and not ask for a socket. So there is not much value in doing this check lazily.

brianc:
Hey I just got back from out of town - will take a look at this tomorrow! thanks!

brianc:
Code looks good to me! I'll get this merged up - thanks. @petebacondarwin are there other things y'all are blocked on wrt cf workers & pg integration? I was talking with @jrf0110 and he said there were some other issues...I'm happy to look at them.

petebacondarwin:
Hi @brianc - can we run a release of pg to get this change out there? It is needed for the new Cloudflare nodejs compatibility to work properly with this library.

PR #3148, opened 2024-02-19
chore (ci): add node v20 to build matrix

alxndrsn:
Per https://nodejs.org/en/about/previous-releases, node v20 is "active".

`pg-native` tests are skipped on Node 20 __TODO because pg-native doesn't support node 20?__

PR #3173, opened 2024-03-15
ci: don't fail fast

alxndrsn:
This allows all matrix builds to run, making it easier to understand if a given failure is specific to nodejs version(s), or more general.


alxndrsn:
I've personally found this useful trying to debug https://github.com/brianc/node-postgres/pull/3148

brianc:
> This allows all matrix builds to run, making it easier to understand if a given failure is specific to nodejs version(s), or more general.

well yeah that's super useful! thanks 😁

PR #3178, opened 2024-03-26
query: remove unused prop: _promise

alxndrsn:
Use of this prop was probably removed in 2017 (a0eb36d81938e488b3bc5369faee74fe22f36949).


charmander (approved):

brianc:
💃 ❤️ ty! Tomorrow is my last day at work! Then some vacation trips, then its node-postgres times for me!

PR #3181, opened 2024-03-29
Remove unused protocol-related dependencies from pg

charmander:
buffer-writer was replaced with pg-protocol in 3ff91eaa3222657fd51ea463b8086d134a505404, and packet-reader in 520bd3531990f32c3e00b20020c67f6ac6c70261.


brianc:
Ahhh yes good old buffer writer.  Back when “make every line of code its own npm module” was trendy. 🤓 thanks for the maintenance.

Issue #3187, opened 2024-04-02
Regression in pg@8.11.4/pg-connection-string@2.6.3 impacting older node URL polyfills


IgorMinar:
We (Cloudflare) received several reports from our developers that a recent release (pg@8.11.4/pg-connection-string@2.6.3) broke applications using node-postgress with Cloudflare.

Our investigation led us to https://github.com/brianc/node-postgres/commit/d21cc09556899b8038ec23613a801c19228637ca which aimed to disambiguate between Node and Browser `URL` object. Unfortunately there are two issues with this change:

1. The change breaks applications that rely on node polyfills that have not been updated to export the `URL` class from the `url` module. For example the commonly used: [rollup-plugin-node-polyfills](https://github.com/ionic-team/rollup-plugin-node-polyfills/blob/master/polyfills/url.js#L39) While these polyfills are not up to date with the latest Node.js API surface, there are many people that depend on them.

2. The change shouldn't actually make a material difference to applications because the `globalThis.URL` and `require('url').URL` are the same in recent Node.js versions (`globalThis.URL === require('url').URL`). Also the implementation of `URL` in browser and node.js are compatible, following the whatwg spec (https://nodejs.org/api/url.html#the-whatwg-url-api) which modern node.js polyfills implement via [a simple reexport of globalThis.URL as the implementation](https://github.com/unjs/unenv/blob/b2e8645a52cfb227d71ddfdf42d8de5e836b849b/src/runtime/npm/whatwg-url.ts#L4).

While this issue is resolvable by updating all outdated polyfills and applications using them, for many users that will require too much churn. Since, as far as we can tell, the original change that caused this problem doesn't seem to make a difference for anyone, we recommend that https://github.com/brianc/node-postgres/commit/d21cc09556899b8038ec23613a801c19228637ca is reverted and v8.11.4 is deprecated on npm.

brianc:
Thanks for the report. Agree w/ you. Will do that now.

brianc:
Thanks for the report - should be fixed in `pg@8.11.5` and `pg-connection-string@2.6.4`

IgorMinar:
Awesome! Thank you so much, @brianc! 🧡

janpio:
Thanks as well for the quick fix from @prisma, 8.11.4 made our CI pretty red 😆

brianc:
@janpio sorry about that! Curious about how it failed for you? Would love to write a test to prevent it from happening in the future!

[...]

PR #3188, opened 2024-04-02
Revert "fix: conflict between browser URL object and Node URL object"

brianc:
Reverts brianc/node-postgres#3061

This addresses a backwards incompatible regression described by #3187


brianc:
For future reference if you run into an incompatibility of the URL module due to old polyfills or other environments, its recommended to use the [config object based connection](https://node-postgres.com/apis/client#new-client) instead of a connection string to avoid the issue.

PR #3192, opened 2024-04-05
docs: update queries.mdx pg-format link to markdown

iaurg:
If we use `<a>` here, the template does not visually identify it as a link, so I have switched to the markdown link format.

With `a`:
image

With markdown:
image


charmander (approved):
Thank you!

PR #3207, opened 2024-05-10
Reorder the user, password, host, port and database appearances for uniformity.

juangpc:
Hi,
Trying to add uniformity to the use of the keys (user, password, host, port and database) in the code and docs.
I see the use of fields scattered in code and in documentation, in many different orders. Figured it could help for clarity to try to have a similar as possible order. Makes it easier to see if you're forgetting one of them, identify changes, etc...

New to pg. Figured could be helpful. Thanks for a great package. 

Fun fact (Had to check it out).
When you have 3 terms -> ternary.
when you have 4  terms -> quaternary. (ref.[stack exchange discussion](https://english.stackexchange.com/questions/25116/what-follows-next-in-the-sequence-unary-binary-ternary))

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,8 +29,8 @@ jobs:
         env:
           POSTGRES_USER: postgres
           POSTGRES_PASSWORD: postgres
-          POSTGRES_DB: ci_db_test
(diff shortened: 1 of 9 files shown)


brianc:
Hi! Thanks for the house keeping PR! Things like this add up overtime to a generally better code base & I appreciate it! 😄

Issue #2774, opened 2022-07-18
Support passing AbortSignal in query()


mattbishop:
Node offers an AbortSignal API that gives async operations a way to cancel their progress if aborted by the caller. Common use cases are web apps that see their request closed before completion, or in a local app, the user hits Ctrl-C to kill a running program. node-postgres's query() command could accept an AbortSignal instance and listen to it for abort events. When an abort occurs, it would close the Postgres query immediately.

A decent writeup can be found in this logrocket blog: https://blog.logrocket.com/complete-guide-abortcontroller-node-js/

Node Stream offers an option to pass an AbortSignal instance to close streams when an abort occurs. https://nodejs.org/dist/latest-v18.x/docs/api/all.html#all_stream_streamaddabortsignalsignal-stream

mattbishop:
Wow I had no idea it was so complicated!

I guess you can’t just kill the tcp connection on abort?

—
Matt Bishop

nicholaswmin:
@brianc Hi Brian,

I assume what you're talking is `pg_terminate_backend` or something of the sort? 
If that's the case it is a pile of garbage to reason with; but I think it should be documented somewhere.This is a recurring problem with people using streams; some I would be are not even aware they have runaway queries that need to be cancelled.

brianc:
> pg_terminate_backend

Yeah the `cancel` command in the protocol works the same as `pg_cancel_backend(pid)` actually - cancels the active query. `pg_terminate_backend` actually (attempts) to terminate the process (postgres uses a 1 process per connection model) which ends up closing the client entirely.   The problem here is if node-postgres  were to have a cancel function its hard to figure out the right API for it while maintaining principal of least surprise. The easiest way honestly would be to export a top level `cancel` function, since canceling a query _requires_ a client and _requires_ a client that's not currently executing an existing query.  This actually works:

```ts
async function cancel(client: pg.Client): void {
  // note: need to use the same exact connection config to connect to the same backend
  const newClient = new pg.Client({
    host: client.host,
    user: client.user,
    // copy every single optional connection config for ssl and so-forth
  })
  await newClient.connect()
  // note: this may or may-not cancel your query...you don't really have total control over that
  // https://www.postgresql.org/docs/current/functions-admin.html#FUNCTIONS-ADMIN-SIGNAL
  await newClient.query('SELECT PG_CANCEL_BACKEND($1)', [client.processId])
}
```

[...]

PR #3212, opened 2024-05-12
Cancellable Query with AbortSignal option

boromisp:
A suggestion for a new cancellation API (based on the original request in https://github.com/brianc/node-postgres/issues/2774)

This query-specific cancellation API only makes sense because the library waits for each query to complete before it submits the next. With query pipelining it would be a lot messier - or simpler, because the API could offer less guarantees.

While the tests are turned off if `AbortController` is not available, the library code itself should still be compatible with older node versions.

**Example**

```js
const r1 = await client.query('SELECT now()')

try {
  await client.query({
    text: 'SELECT pg_sleep(10)',
    signal: AbortSignal.timeout(1_000),
  })
} catch (err) {
  console.log(err.code, err.message) // 57014 canceling statement due to user request
}

const r2 = await client.query(
  'SELECT now() - $1 AS delay',
  [r1.rows[0].now],
)

console.log(r2.rows[0].delay) // PostgresInterval { seconds: 1, milliseconds: 9.592 }
```

---

First, in a separate commit (https://github.com/brianc/node-postgres/commit/913967f795b1a3a72865d94e0020671fba68f0ae) is a small change that moves the SSL initialization logic to the `Connection`.
This means the `Connection` no longer needs a separate an `'sslconnect'` event.

---

Then, the `Connection` got a new `con.cancelWithClone()` method that tries to connect to the exact same server, and send a cancel request.

This required a bit of extra state in the `Connection`, to keep track of the `host`, `port`, `config` and the `backendKeyData`.

I'm not thrilled with duplicating state from the `Client` (or with the name of the method for that matter).

What I like about this approach is that all the logic around the cancellation can be implemented in the `Submittable` and doesn't depend on any special handling by the `Client`.

---

Finally the `Query` got a `signal` option that can be used to cancel it either before, or after it has been submitted to the server.

If the query has been canceled before submit, it will return the abort reason in submit.

If an actual cancel request has been sent, the query callback and the `'error'` and `'end'` events are delayed until the cancel request has finished.

This is done to avoid accidentally canceling a different query on the same connection.

If the cancel request isn't completed cleanly, the original connection is destroyed.

_This might be triggered if you get an ECONNRESET after sending the cancel request? Honestly I'm not sure._
_This should probably treat the errors after the cancel request has been sent differently from errors during connecting._

### DIFF
--- a/packages/pg/lib/client.js
+++ b/packages/pg/lib/client.js
@@ -51,6 +51,7 @@ class Client extends EventEmitter {
         keepAlive: c.keepAlive || false,
         keepAliveInitialDelayMillis: c.keepAliveInitialDelayMillis || 0,
         encoding: this.connectionParameters.client_encoding || 'utf8',
+        Promise: this._Promise,
       })
     this.queryQueue = []
     this.binary = c.binary || defaults.binary
--- a/packages/pg/lib/client.js
+++ b/packages/pg/lib/client.js
@@ -56,15 +56,8 @@ class Client extends EventEmitter {
     this.binary = c.binary || defaults.binary
     this.processID = null
     this.secretKey = null
+    // TODO: remove in next major release?
     this.ssl = this.connectionParameters.ssl || false
-    // As with Password, make SSL->Key (the private key) non-enumerable.
-    // It won't show up in stack traces
-    // or if the client is console.logged
-    if (this.ssl && this.ssl.key) {
-      Object.defineProperty(this.ssl, 'key', {
-        enumerable: false,
-      })
-    }
 
     this._connectionTimeoutMillis = c.connectionTimeoutMillis || 0
   }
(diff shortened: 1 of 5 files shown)


charmander:
> This query-specific cancellation API only makes sense because the library waits for each query to complete before it submits the next.

I’m pretty sure there are still unavoidable race conditions that make this not the right API.

boromisp:
> I’m pretty sure there are still unavoidable race conditions that make this not the right API.

That was my original impression as well.
But I wanted to write up an implementation, to see the actual problems.

One possible issue I see with this is destroying the connection, if the cancellation connection closes with an error.

This is done, so we don't release the connection with a possible cancellation pending.

However, if the cancellation actually failed, we just left possibly long running query alive on the server. [Explicitly sending an RST](https://nodejs.org/api/net.html#socketresetanddestroy) might be necessary in this case, assuming that is handled similarly as cancel requests by the server.

Even if all this works as I think it does, I'm not so sure if this API is a good idea.

---

The intent of the code:

- If you cancel before submitting the query, you skip the query.
- If you cancel after submitting the query,  but before the query is processed, the query should be executed.
- If you cancel while the query is being executed, the query should be canceled. _I'm not so sure about the interaction with the extended query protocol._
- If you cancel after the query has been executed, nothing should happen.
- Additionally, if there is an error during cancellation, the connection is destroyed.

So depending on timing the query might be canceled, prevented from being executed or complete successfully.

But the connection should be left In a clean state, or be destroyed.

PR #3225, opened 2024-05-30
Add pg-native to monorepo

brianc:
Makes fixing things like #3214 easier and generally makes contributing a bit easier.  I didn't do much to "modernize" the pg-native codebase other than running it through the standard `eslint --fix` that is applied to the rest of the code. There's some easy opportunities there to update it to es6 and so on...it still uses some pretty antiquated coding styles in places.


brianc:
Took me a bit to fiddle w/ the cloudflare stuff I didn't write (but now I maintain, i suppose!) - that `wrangler` package was doing some at-runtime npx stuff, so I upgraded it. Anyways...also re-enabled native tests for node 20, and added node 22 to the test matrix - so libpq is building & working on all nodes again. yay.  Will publish this in the am (so I can be around in case there's some weird regression introduced)

PR #3234, opened 2024-06-04
Improve race condition in pg-native cancel test

brianc:
Since the pg-native tests run more now the race condition is more apparent. In CI I think the query doesn't have time to get to the backend sometimes before the cancel is executed.  This should give it more time. Still somewhat of a race condition possible, but much less likely.  Also removed unneeded non-dev dependency in pg-native.

### DIFF
--- a/packages/pg-native/package.json
+++ b/packages/pg-native/package.json
@@ -23,8 +23,7 @@
   "homepage": "https://github.com/brianc/node-pg-native",
   "dependencies": {
     "libpq": "1.8.13",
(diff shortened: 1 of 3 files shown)


brianc:
ohhhh good point about the pg_sleep being in seconds - totally forgot about
that. I'll let the 100ms timeout ride for now to send the cancel (as before
there was NO timeout) and if it bugs again I'll increase it. Appreciate the
eyes on things!! :heart: hope everything is groovy & you can make it down
yonder soon :)

PR #3239, opened 2024-06-05
CHANGELOG: fix typo

alxndrsn:


sehrope:
Maybe combine some of these PRs into logical groups? For example all the docs ones could be together, either as individual commits or one "fix typos..." commit. Would make review and merging easier.

Ditto for the no-op or clean-up-import stuff as it's all kind of related.

PR #3247, opened 2024-06-06
eslint: enable rule: @typescript-eslint/no-unused-vars

alxndrsn:
When enabling this rule, it's recommended to also *disable* the standard `no-unused-vars` rule.  Although `no-unused-vars` is not currently enabled, it seems helpful to explicitly disable it here.

See: https://typescript-eslint.io/rules/no-unused-vars/


packages/pg-query-stream/test/async-iterator.ts
```diff
@@ -122,9 +122,11 @@ if (!process.version.startsWith('v8')) {
       const pool = new pg.Pool({ max: 1 })
       const client = await pool.connect()
 
+      /* eslint-disable @typescript-eslint/no-unused-vars */
       for await (const _ of client.query(new QueryStream('select TRUE', [], { highWaterMark: 1 }))) break
```
alxndrsn, line 126: I'm not sure how this line is different to:

```
await client.query(new QueryStream('select TRUE', [], { highWaterMark: 1 }))
```

.eslintrc
```diff
@@ -11,5 +11,11 @@
     "node": true,
     "es6": true,
     "mocha": true
+  },
+  "rules": {
+    "@typescript-eslint/no-unused-vars": ["error", {
+      "args": "none"
```
alxndrsn, line 17: Unused arguments are declared in a lot of places, and it could be argued they convey information about APIs being overwritten, or opportunities for extension e.g.

https://github.com/brianc/node-postgres/blob/a24a24dea19481fefb84ee42e7e5cbe0f2eb06b4/packages/pg/test/unit/connection/error-tests.js#L67

https://github.com/brianc/node-postgres/blob/a24a24dea19481fefb84ee42e7e5cbe0f2eb06b4/packages/pg/lib/query.js#L122

packages/pg/test/integration/client/error-handling-tests.js
```diff
@@ -113,7 +112,7 @@ var ensureFuture = function (testClient, done) {
 suite.test('when query is parsing', (done) => {
   var client = createErorrClient()
 
-  var q = client.query({ text: 'CREATE TEMP TABLE boom(age integer); INSERT INTO boom (age) VALUES (28);' })
+  client.query({ text: 'CREATE TEMP TABLE boom(age integer); INSERT INTO boom (age) VALUES (28);' })
```
alxndrsn, line 115: Is this line even necessary?

packages/pg-query-stream/test/async-iterator.ts
charmander: The difference is that it actually consumes the async iterable. `client.query` returns the `QueryStream` when it’s passed one, so plain `await` would do nothing.

alxndrsn: Thanks!

> it actually consumes the async iterable

I think "consumes" is a strong word - it looks like `break` means it just takes the first item from the query stream.  It looks analogous to this:

```js
let calls=0;
async function* fakeQueryStream() {
  yield ++calls;
  yield* fakeQueryStream();
}
for await(const _ of fakeQueryStream()) break;
console.log('calls:', calls);
```

which will print:

```
calls: 1
```

-----

This suggests that the current code

```js
for await (const _ of client.query(new QueryStream('select TRUE', [], { highWaterMark: 1 }))) break
```

is the equivalent of:

```js
await client.query(new QueryStream('select TRUE', [], { highWaterMark: 1 }))[Symbol.asyncIterator]().next()
```

-----

To consume the stream, I think `break` should be replaced with `{}` or `continue`.

charmander: There’s only one row, but either asserting that or using an empty loop body seems good, yes.

alxndrsn: Hmm the test title is "supports breaking with low watermark", so maybe it's explicitly testing something to do with `break`...??  Anyway, not directly relevant to this PR.

packages/pg/test/integration/client/error-handling-tests.js
alxndrsn: Separate PR to remove this line at https://github.com/brianc/node-postgres/pull/3256

alxndrsn: Per https://github.com/brianc/node-postgres/pull/3256#issuecomment-2167937757, `boom` table is referenced inside `ensureFuture()`.

brianc:
hey nice! sorry for the delay - was in colorado for some family visiting. back now. merging! thanks for doing this!!

PR #3252, opened 2024-06-10
Add missing _checkPgPass on _handleAuthSASLContinue

cesco69:


charmander:
Why does this one need `_checkPgPass` on top of `_handleAuthSASL`’s?

cesco69:
> Why does this one need `_checkPgPass` on top of `_handleAuthSASL`’s?

`this.password` can be a `function` and `_checkPgPass` handle this case, if `this.password` is a function, call it and set `this.password` as string

charmander:
Does `_handleAuthSASL` not already take care of that? Can you give an example of a use of the library that behaves differently before and after this change?

brianc:
yeah I need tests (preferably one that failed before this code was added) or this can't be merged.  As an aside, I am thinking of removing built in pg-pass support & rather making that something the user does their own self before calling the client rather than reading a file from disk on every client connection.  That would obvs be a breaking change so not until v 9.x

PR #3263, opened 2024-06-19
eslint: enable recommended ruleset

alxndrsn:


alxndrsn:
@brianc is there anything I can do to help get this PR considered for review?

charmander:
I’d merge this if it were just the ESLint rule. Don’t want to combine two unrelated changes.

charmander (changes requested):
I’d merge this if it were just the ESLint rule. Don’t want to combine two unrelated changes.

alxndrsn:
> I’d merge this if it were just the ESLint rule. Don’t want to combine two unrelated changes.

Thanks for the feedback.  I've reverted the unrelated changes.

charmander (approved):

charmander:
Oh, I see there are no code changes associated with this. How about applying the entire ESLint recommended configuration, then?

alxndrsn:
> How about applying the entire ESLint recommended configuration, then?

I've done this, and updated the PR title to reflect the change.

Eslint's ECMAScript version is bumped to 2020 as this was required to support `globalThis` (already in use before this PR), and [2020 seems to be fully supported by node 16](https://compat-table.github.io/compat-table/es2016plus/), which is the oldest version supported in this repo's CI tests:

https://github.com/brianc/node-postgres/blob/a9fd34fb4217a7c16f02a4726efd61ad6a4f4796/.github/workflows/ci.yml#L40-L45

It's picked up some obvious errors, and some stuff which is more debatable.

packages/pg-pool/test/idle-timeout-exit.js
```diff
@@ -11,7 +11,6 @@ if (module === require.main) {
   pool.query('SELECT NOW()', (err, res) => console.log('completed first'))
   pool.on('remove', () => {
     console.log('removed')
-    done()
```
charmander, line 14: I haven’t really looked at this, but there’s probably more to do here than quietly deleting a line. Does none of this code run at all, then?

packages/pg/bench.js
```diff
@@ -25,13 +25,11 @@ const exec = async (client, q) => {
...
   let count = 0
-  while (true) {
+  do {
     await exec(client, q)
     count++
-    if (Date.now() - start > time) {
-      return count
-    }
-  }
+  } while (Date.now() - start > time)
```
charmander, line 31: The logic here has been inverted.

charmander, line 31: The logic here has been inverted.

packages/pg/lib/crypto/cert-signatures.js
```diff
@@ -18,11 +18,11 @@ function readASN1Length(data, index) {
 }
 
 function readASN1OID(data, index) {
-  if (data[index++] !== 0x6) x509Error('non-OID data', data) // 6 = OID
+  if (data[index++] !== 0x6) throw x509Error('non-OID data', data) // 6 = OID
 
   const { length: OIDLength, index: indexAfterOIDLength } = readASN1Length(data, index)
   index = indexAfterOIDLength
-  lastIndex = index + OIDLength
+  let lastIndex = index + OIDLength
```
charmander, line 25: Yikes! Good catch.

packages/pg/test/unit/client/sasl-scram-tests.js
```diff
@@ -225,7 +225,7 @@ suite.test('sasl/scram', function () {
               0x0d, // signature algorithm length
               0x06, // ASN.1 OID
               0x09, // OID length
-              0x2a, // OID: 1.2.840.113549.1.1.11 (RSASSA-PKCS1-v1_5 / SHA-256​)
+              0x2a, // OID: 1.2.840.113549.1.1.11 (RSASSA-PKCS1-v1_5 / SHA-256)
```
charmander, line 228: (note for other readers: the difference is a zero-width space after “SHA-256”)

charmander (changes requested):
Thank you! I see lots of important improvements here.

charmander:
> Eslint's ECMAScript version is bumped to 2020 as this was required to support `globalThis` (already in use before this PR), and [2020 seems to be fully supported by node 16](https://compat-table.github.io/compat-table/es2016plus/), which is the oldest version supported in this repo's CI tests:

You can define `navigator` as a global instead. Despite the lack of CI, pg 8 still supports Node 8(!).

packages/pg/bench.js
alxndrsn: I don't think the logic has been changed?  just changed to a more natural syntax:
```js
do { ... } while(condition) return x;
```
vs previous
```js
while(true) { ...; if(condition) return x; }
```

Code changes could be minimised by using `// eslint-disable-next-line no-constant-condition` if that's preferred.

charmander: Exiting a loop when *P* is the opposite of continuing a loop while *P*.

packages/pg/lib/crypto/cert-signatures.js
alxndrsn: Could implications of this be interesting?  `lastIndex` might be scoped at the file/module level currently, but maybe safe in practice?

charmander: I haven’t run it to check, but I assume it’s just leaking to the global object because the module isn’t in strict mode.

packages/pg/bench.js
alxndrsn: :facepalm: of course it is

alxndrsn: reverted, and added lint exception

alxndrsn:
> You can define `navigator` as a global instead. Despite the lack of CI, pg 8 still supports Node 8(!).

Any thoughts on what to do with the existing `globalThis` in `packages/pg/lib/crypto/utils-webcrypto.js`?  From the filename, maybe it doesn't get used in node, so a specific exception would be ok?

packages/pg-pool/test/idle-timeout-exit.js
charmander: My mistake! This is correct.

charmander (approved):

PR #3267, opened 2024-06-25
Protoype of kerberos support (non parameterized with TODOs)

blackentropy:


packages/pg/lib/client.js
```diff
@@ -198,6 +204,38 @@ class Client extends EventEmitter {
...
+  }
+
+  async _handleGSSContinue(msg) {
+    try {
+      const inToken = msg.inToken
+      const token = await this.client.step(inToken)
+
+      // TODO: probably a better way to handle this.
+      if (token == null) {
+        this.emit('error', 'Received null GSSAPI token on continue')
```
blackentropy, line 230: Just a quick note; the token can be null if the negotiation has already been successfully completed. I suspect the right thing to do is to do something like:

```
if (inToken === 'oRQwEqADCgEAoQsGCSqGSIb3EgECAg==') {
  return; // negotiation successful, no need to do anything further.
}

see: https://github.com/jcmturner/gokrb5/blob/master/spnego/http.go#L198 for where that magic string came from.

albertchang:
@brianc We'd ideally like to get this merged and released through the main package. Could you let us know what the procedure would be? There's unfortunately not a lot of documentation for this flow, and it's not well trodden.

brianc:
> We'd ideally like to get this merged and released through the main package. Could you let us know what the procedure would be?

I can do that! Its very easy on my side to release new versions. Ideally all features should have tests along side them. I realize testing against a kerberos system is.... _unpleasant_ ... do you have any way you could include tests? Even the `pg-protocol` stuff typically ships with unit tests. Its very high likelyhood without tests sometime within the next 10 years this code will change and/or break without them.

brianc:
@albertchang note also: CI is running on this PR now & lint is failing

albertchang:
> > We'd ideally like to get this merged and released through the main package. Could you let us know what the procedure would be?
> 
> I can do that! Its very easy on my side to release new versions. Ideally all features should have tests along side them. I realize testing against a kerberos system is.... _unpleasant_ ... do you have any way you could include tests? Even the `pg-protocol` stuff typically ships with unit tests. Its very high likelyhood without tests sometime within the next 10 years this code will change and/or break without them.

Cool, I'll try to take a stab at writing some tests for this change and circle back afterwards. Thanks!

brianc:
> Also, is CI not configured to run automatically on changes to the branch?

Boo i need to set that up. having to keep re-approving it is no bueno!

albertchang:
@brianc Could you approve the workflow run for CI? Thanks!

albertchang:
@brianc Would also appreciate a review when you get a chance!

Issue #3174, opened 2024-03-20
lib/client.js this.activeQuery sometimes null


rhodey:
Hello All, thanks to all the contributors who keep this project going, I have an issue to report

Before today I ran into this issue in prod approx 6 times total in ~12 months, but today for some reason this issue has turned up six times.

We have approx 24 services with 200 tasks running on AWS ECS and 90% of these services make use of this pg library in one way or another, another maybe important note is that they all use the pg.Pool

So the stack trace we get in prod from multiple services is:

```
TypeError: Cannot read properties of null (reading 'name')
at Client._handleParseComplete (/app/node_modules/pg/lib/client.js:380:26)
at Connection.emit (node:events:517:28)
at Connection.emit (node:domain:489:12)
at /app/node_modules/pg/lib/connection.js:116:12
at Parser.parse (/app/node_modules/pg-protocol/dist/parser.js:40:17)
at Socket.<anonymous> (/app/node_modules/pg-protocol/dist/index.js:11:42)
at Socket.emit (node:events:517:28)
at Socket.emit (node:domain:489:12)
at addChunk (node:internal/streams/readable:335:12)
at readableAddChunk (node:internal/streams/readable:308:9)
```

So most of that stack trace does not say a lot but I think what it is saying is to see this line here:

https://github.com/brianc/node-postgres/blob/91de4b94538ba802456ea52fa830bb5567c7b792/packages/pg/lib/client.js#L389

And it must be saying that `this.activeQuery` is null

Now I could open a PR and modify this function:

```js
  _handleParseComplete(msg) {
    // if a prepared statement has a name and properly parses
    // we track that its already been executed so we don't parse
    // it again on the same client
    if (this.activeQuery.name) {
      this.connection.parsedStatements[this.activeQuery.name] = this.activeQuery.text
    }
  }
```

And simply change it to:

```js
  _handleParseComplete(msg) {
    // if a prepared statement has a name and properly parses
    // we track that its already been executed so we don't parse
    // it again on the same client
    if (this.activeQuery && this.activeQuery.name) {
      this.connection.parsedStatements[this.activeQuery.name] = this.activeQuery.text
    }
  }
```

There is a chance that is all that is needed but someone more knowledgeable about the library would know if this is maybe a symptom of a larger problem. Anyone online care to comment ??

I am concerned about patching the library and testing it out on our prod stack because I don't want my edit to cause the lib to somehow be confused and make a serious error.

[...]

PR #3289, opened 2024-08-01
Handle bad message ordering - make it catchable. Fixes 3174

brianc:
The __biggest__ issue was when messages came in out of order sometimes the client would throw a null ref error in an uncatchable place in the code & crash the node process.  As far as I know, this PR should totally fix that issue, changing the unhandled null ref error in the background into a catchable error emitted by the client (or pool if you use `pool.query`). 

Wrote a fake postgres backend to send messages out of order to test these weird edge cases outlined in #3174 where proxies or pg-bouncer or the like is getting messages mixed up.  Can extend the test suite as needed and improve other edge cases. The native driver only emits a `notice` event when it receives an out of order message.

### DIFF
--- a/packages/pg/lib/client.js
+++ b/packages/pg/lib/client.js
@@ -383,7 +383,8 @@ class Client extends EventEmitter {
 
   _handleParseComplete() {
     if (this.activeQuery == null) {
-      this.emit('error', new Error('Received parseComplete when not in parsing state'))
+      const error =new Error('Received parseComplete when not in parsing state') 
+      this._handleErrorEvent(error)
       return
     }
     // if a prepared statement has a name and properly parses
--- a/packages/pg/lib/client.js
+++ b/packages/pg/lib/client.js
@@ -377,13 +377,18 @@ class Client extends EventEmitter {
   }
 
   _handleCommandComplete(msg) {
+    if (this.activeQuery == null) {
+      const error = new Error('Received unexpected commandComplete message from backend.') 
+      this._handleErrorEvent(error)
+      return
+    }
     // delegate commandComplete to active query
     this.activeQuery.handleCommandComplete(msg, this.connection)
   }
 
   _handleParseComplete() {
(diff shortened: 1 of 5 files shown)


packages/pg/lib/client.js
```diff
@@ -381,7 +381,11 @@ class Client extends EventEmitter {
     this.activeQuery.handleCommandComplete(msg, this.connection)
   }
 
-  _handleParseComplete(msg) {
+  _handleParseComplete() {
+    if (this.activeQuery == null) {
+      this.emit('error', new Error('Received parseComplete when not in parsing state'))
```
charmander, line 386: Should this use `_handleErrorEvent`, since the connection is in an unpredictable state afterwards?

rhodey: I can see how using `_handleErrorEvent` might be a good idea here too, I think what @brianc communicated to me about this PR is that he thinks the issue is always related to an unexpected message arriving at a client which is in the idle state and so if that is the case it might not be required to do the full `_handleErrorEvent` but I can see how maybe doing full `_handleErrorEvent` is more robust.

rhodey:
Hey @brianc, here to add my two cents, first wanna say thanks for your attention to the issue report and thanks to @charmander as well for weighing in! I can see that this PR as is addresses `_handleParseComplete()` being called on a idle client and that makes good sense to me, however if you review [this comment](https://github.com/brianc/node-postgres/issues/3174#issuecomment-2010734480) on 3174 you will see that we also have a history of seeing `_handleCommandComplete` being called on (most likely) an idle client and crashing the lib as well.

I think it makes sense that whatever is decided to be done for `_handleParseComplete` can be done within `_handleCommandComplete` as well, right? And on the subject of there being a second case is there a third or fourth kind of protocol message which could show up unexpected on an idle client and maybe if there is then there could be a place further up in the call stack to check if the client is idle and emit errors from there?

Thanks!

PR #3143, opened 2024-02-07
crypto for edge runtime

artemavrin:

PR #3296, opened 2024-08-14
fix - Removed import and usage of nodeCrypto when not running on nodejs

ETHANTALJAFFE:
Following [this](https://github.com/brianc/node-postgres/pull/3143) discussion I implemented the suggested solution.

Removing the `nodeCrypto` check inside of `md5` function because we anyway only export `utils-webcrypto` when we're in the correct environment.

### DIFF
--- a/packages/pg/lib/crypto/utils-webcrypto.js
+++ b/packages/pg/lib/crypto/utils-webcrypto.js
@@ -1,5 +1,3 @@
-const nodeCrypto = require('crypto')
-
 module.exports = {
   postgresMd5PasswordHash,
   randomBytes,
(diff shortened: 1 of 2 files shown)


brianc:
Looks like the tests are dying with an 'unrecognized algorithm name' in node:

https://github.com/brianc/node-postgres/actions/runs/10389540544/job/28768489773?pr=3296#step:8:580

charmander:
As the comment in the existing code said,

> Note that the MD5 algorithm on WebCrypto is not available in Node.js. This is why we cannot just use WebCrypto in all environments.

charmander (changes requested):
As the comment in the existing code said,

> Note that the MD5 algorithm on WebCrypto is not available in Node.js. This is why we cannot just use WebCrypto in all environments.

ETHANTALJAFFE:
@brianc ok I think the reason is because we are checking for node 15 even though node.js only started support webCrypto starting version [22](https://nodejs.org/api/webcrypto.html#web-crypto-api)

I am gonna try and change the if statement to check node 22 instead of 15.

charmander:
[Node 15](https://nodejs.org/docs/latest-v15.x/api/webcrypto.html#webcrypto_web_crypto_api) is correct. Even the version history on the Node 22 docs page mentions versions back to 16.

brianc:
Is this still required? It's been a while and this might have gone stale. Please feel free to re-open if so (there are merge conflicts now)

PR #3308, opened 2024-09-05
Replace `client.release()` with `await client.end()`

Tam:


charmander (changes requested):
Calling `end` on pooled clients isn’t correct. `client.release(true)` would achieve what I think you’re going for, but would always remove the client from the pool (even on a successful transaction), which is inefficient.

brianc:
yeah...you really shouldn't be calling end on pooled clients yourself ever. Will cause weird issues.

Issue #3309, opened 2024-09-07
Custom type parsers do not run for multi-line statements


mantariksh:
## Minimal reproducible example

See https://github.com/mantariksh/pg-bug-repro/blob/main/pg.mjs, the instructions are in the README.

## Behaviour observed

When I pass a custom type parser via the `types.getTypeParser` function passed to the `Client` constructor, the type parser runs correctly for single-line statements but not for multi-line statements.

For instance, consider the following table (where `enum_Users_roles` has two possible values, `Admin` and `View`)
```
CREATE TABLE IF NOT EXISTS "Users" ("id"   SERIAL , "name" VARCHAR(255), "roles" "public"."enum_Users_roles"[], PRIMARY KEY ("id"));
```

I want `Client` to check for any values wrapped in braces (`{}`) and parse them as an array, as follows:
```javascript
const getTypeParser = () => {
  return (value) => {
    if (
      typeof value === "string" &&
      value.startsWith("{") &&
      value.endsWith("}")
    ) {
      return value.slice(1, -1).split(",");
    }
    return value;
  };
};

const client = new Client({
  connectionString: CONNECTION_STRING,
  types: {
    getTypeParser,
  },
});
```

When I run a single-line statement, I get the correct result:
```JavaScript
const singleStatementResult = await client.query(
  'INSERT INTO "Users" ("id","name","roles") VALUES (DEFAULT,$1,$2) RETURNING "id","name","roles";',
  ["Name", ["Admin", "View"]]
);
console.log(singleStatementResult.rows);
// Console output with array correctly parsed
// [ { id: '1', name: 'Name', roles: [ 'Admin', 'View' ] } ]
```

However, when I run a multi-line statement, the array is no longer parsed (sorry that the SQL statement here is somewhat complex, I got it directly from `sequelize`):
```JavaScript
const multiStatementResult = await client.query(
[...]

PR #3310, opened 2024-09-07
fix: use existing Result types for new Result

mantariksh:
## Context

Closes https://github.com/brianc/node-postgres/issues/3309
Related to https://github.com/sequelize/sequelize/issues/17479

## Approach

The issue seems to be that the `_types` member of `Result` is directly overwritten by `Client.query`:
https://github.com/brianc/node-postgres/blob/54eb0fa216aaccd727765641e7d1cf5da2bc483d/packages/pg/lib/client.js#L567

However, in `_checkForMultiRow`, we then create new `Result`s, but without the overridden types:
https://github.com/brianc/node-postgres/blob/54eb0fa216aaccd727765641e7d1cf5da2bc483d/packages/pg/lib/query.js#L68

The approach is to reuse the types of the original `Result` when creating a new one.

### DIFF
--- a/packages/pg/lib/query.js
+++ b/packages/pg/lib/query.js
@@ -65,7 +65,7 @@ class Query extends EventEmitter {
       if (!Array.isArray(this._results)) {
         this._results = [this._result]
       }
-      this._result = new Result(this._rowMode, this.types)
+      this._result = new Result(this._rowMode, this._result._types)
       this._results.push(this._result)
     }
   }
--- a/packages/pg/test/integration/client/custom-types-tests.js
+++ b/packages/pg/test/integration/client/custom-types-tests.js
@@ -22,6 +22,21 @@ suite.test('custom type parser in client config', (done) => {
   })
 })
 
+suite.test('custom type parser in client config with multiple results', (done) => {
+  const client = new Client({ types: customTypes })
+
+  client.connect().then(() => {
+    client.query(
+      `SELECT 'foo'::text as name; SELECT 'bar'::text as baz`,
+      assert.success(function (res) {
+        assert.equal(res[0].rows[0].name, 'okay!')
+        assert.equal(res[1].rows[0].baz, 'okay!')
+        client.end().then(done)
+      })
+    )
+  })
+})
+
 // Custom type-parsers per query are not supported in native
 if (!helper.args.native) {
   suite.test('custom type parser in query', (done) => {


charmander:
Thanks for proposing a fix.

`Client` probably shouldn’t be reaching into `Query` and `Result` like that in the first place, and should forward its `types` to the query objects it creates instead. That would be a breaking change now, when passing query objects to `Client#query`, but maybe it’s possible to add the types to the options now to fix this bug and defer removing `query._result._types = this._types` until the next major version? @brianc

(Either way, it should also have a test.)

mantariksh:
Got it, I'll add a test. I'll try to find some time to work on it over the next couple of days.

mantariksh:
@charmander I added a test for this case in 3c1a7d1

mantariksh:
Just checking, are there any further changes I should make? No rush, just wondering if there's anything I can do to help!

brianc:
Not a huge fan of reaching into the client like that, but also I think I went a bit overboard with what was "private" in the past (particularly since not a lot is really actually private in JS). I'm okay w/ this as a quick fix for now, and since there are tests covering it it will be resilient to refactoring when that time comes. :)

mantariksh:
Seems like tests are failing for `pg-native` in NodeJS 20 and 22. Do you think this is related to the changes, or might it be that the tests are flaky? Noticed a recent test run for an unrelated change that failed with a similar error code: https://github.com/brianc/node-postgres/actions/runs/10098627832/job/27926116837?pr=3271

Happy to debug further if you think the changes caused the test to fail.

mantariksh:
Hey @brianc @charmander! The tests are passing for me locally on both Node 20 and 22 (see screenshots). Do you think it might be worth rerunning the actions to see if the test failures might be due flaky tests? Alternatively let me know if you think I need to debug further.

Screenshot 2024-10-19 at 11 34 41 PM
Screenshot 2024-10-19 at 11 37 08 PM

brianc:
> Do you think it might be worth rerunning the actions to see if the test failures might be due flaky tests?

I'll re-run! Once tests pass I'll merge this. sorry for the delay, this looks good. :)

brianc:
This has to do w/ randomly segfaulting (sometimes, but not regularly) in the native bindings, which is quite annoying.  I'm unable to re-run the job? I don't see the button anywhere anyway. Mind pushing an empty or just whitespace commit & hopefully that'll trigger a re-run?

mantariksh:
@brianc Done!

brianc:
Thanks! Got it merged! Will do a release tomorrow! :)

PR #3316, opened 2024-09-19
Support binary result format

msand:


packages/pg-protocol/src/parser.ts
```diff
@@ -278,13 +278,15 @@ export class Parser {
...
+    reader.setBuffer(offset, bytes)
+    const fieldCount = reader.int16()
     const fields: any[] = new Array(fieldCount)
+    const textMode = this.mode === 'text'
     for (let i = 0; i < fieldCount; i++) {
-      const len = this.reader.int32()
+      const len = reader.int32()
       // a -1 for length means the value of the field is null
-      fields[i] = len === -1 ? null : this.reader.string(len)
+      fields[i] = len === -1 ? null : textMode ? reader.string(len) : reader.bytes(len)
```
charmander, line 289: Should this use the field’s mode?

The constructor also still has

```typescript
if (opts?.mode === 'binary') {
  throw new Error('Binary mode not supported yet')
}
```

charmander (changes requested):
Thanks for the bug fix in the serializer.

Tests?

packages/pg-protocol/src/serializer.ts
```diff
@@ -160,7 +160,15 @@ const bind = (config: BindOpts = {}): Buffer => {
...
-  writer.addInt16(binary ? ParamType.BINARY : ParamType.STRING)
+  if (binary) {
+    const columnFormatCount = 1
+    writer.addInt16(columnFormatCount)
+    writer.addInt16(ParamType.BINARY)
+  } else {
+    const columnFormatCount = 0
+    writer.addInt16(columnFormatCount)
+  }
+
```
hjr3, line 171: this has been fixed by https://github.com/brianc/node-postgres/pull/3494

packages/pg-protocol/src/parser.ts
hjr3: Agreed that `mode` is incorrect here.

This may be fixed by https://github.com/brianc/node-postgres/issues/3495

packages/pg-protocol/src/serializer.ts
msand: Great, thanks!

PR #3326, opened 2024-10-05
Fix: Add validation for invalid query parameter values

18rameez:
#### **Issue Description**:
Currently, `new Date(undefined)` (an invalid date) is serialized as `"0NaN-NaN-NaNTNaN:NaN:NaN.NaN+NaN:NaN"`. It should fail early with a JS error instead of being passed to the database.


ikonst:
You probably want to fix `dateToString` in utils.js

ikonst:
Also, consider adding tests.

Issue #3318, opened 2024-09-20
Do not serialize new Date(undefined) as "0NaN-NaN-NaNTNaN:NaN:NaN.NaN+NaN:NaN"


ikonst:
Currently `new Date(undefined)` (an invalid date) is serialized as `"0NaN-NaN-NaNTNaN:NaN:NaN.NaN+NaN:NaN"`. It should preferably fail early with a JS error instead.

18rameez:
@charmander  https://github.com/brianc/node-postgres/pull/3326

itisrajeevsingh:
The behavior you're describing—where new Date(undefined) produces an invalid date is quirk. When you pass undefined, it tries to create a date from undefined, resulting in an Invalid Date object. However, when serialized (e.g., converted to a string), it can lead to unexpected results, like the strange "0NaN-NaN-NaNTNaN:NaN:NaN.NaN+NaN:NaN" output you observed.

To address this, the best practice would be to add validation logic that explicitly checks for invalid inputs to new Date() and throws an error early if the input is invalid. If you'd prefer the function to return a fallback value instead of throwing errors, you could use this approach:


function safeCreateDate(input) {
    const date = new Date(input);
    return isNaN(date.getTime()) ? null : date;
}


console.log(safeCreateDate(undefined)); 
console.log(safeCreateDate("2024-11-19"));

charmander:
@brianc Does making this an error sound good? Target pg 9 with deprecation warning in pg 8, or…?

brianc:
Yah i think that makes sense!

charmander:
(After 9 bad AI-generated PRs on this, I figured I could use it as a demonstration of why the intermediary isn’t necessary. But I also did end up writing all of the code except for the deprecation message and the identifier names, because the AI-generated PRs on this are bad, and had to clean up the branch myself too. It was much slower and messier than doing it entirely by hand.)

PR #3330, opened 2024-10-12
Fix: Add validation for invalid date query parameter values

18rameez:
PR for Issue #3318

Issue Description:
Currently, new Date(undefined) (an invalid date) is serialized as "0NaN-NaN-NaNTNaN:NaN:NaN.NaN+NaN:NaN". It should fail early with a JS error instead of being passed to the database.

### DIFF
--- a/packages/pg/test/integration/client/error-handling-tests.js
+++ b/packages/pg/test/integration/client/error-handling-tests.js
@@ -258,24 +258,17 @@ suite.test('cannot pass non-string values to query as text', (done) => {
   })
 })
 
-
 if (!helper.args.native) {
   suite.test('when a query has an invalid date binding', function (done) {
     var client = createErorrClient()
     var calledDone = false
 
-    client.query(
-      new pg.Query({
-        text: 'SELECT $1::timestamp',
-        values: [new Date(undefined)],
-      }),
-      function (err, res) {
-        if (!calledDone) {
-          calledDone = true
-          assert.equal(err.message, 'Query parameter value cannot be an invalid date.')
-          client.end(done)
-        }
+    client.query(new pg.Query({ text: 'SELECT $1::timestamp', values: [new Date(undefined)] }), function (err, res) {
+      if (!calledDone) {
+        calledDone = true
+        assert.equal(err.message, 'Query parameter value cannot be an invalid date.')
+        client.end(done)
       }
-    )
+    })
   })
 }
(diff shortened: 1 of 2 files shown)


packages/pg/test/integration/client/error-handling-tests.js
```diff
@@ -257,3 +257,18 @@ suite.test('cannot pass non-string values to query as text', (done) => {
...
+    var client = createErorrClient()
+    var calledDone = false
+
+    client.query(new pg.Query({ text: 'SELECT $1::timestamp', values: [new Date(undefined)] }), function (err, res) {
+      if (!calledDone) {
+        calledDone = true
+        assert.equal(err.message, 'Query parameter value cannot be an invalid date.')
+        client.end(done)
+      }
+    })
```
charmander, line 272: The `query` callback can only be called once (and if it weren’t, we wouldn’t want to silently ignore that):

```suggestion

    client.query(new pg.Query({ text: 'SELECT $1::timestamp', values: [new Date(undefined)] }), function (err, res) {
      assert.equal(err.message, 'Query parameter value cannot be an invalid date.')
      client.end(done)
    })
```

18rameez: It will cause this error, as the connection also gets terminated, and the callback will be executed again.

```
Message: 'Connection terminated' == 'Query parameter value cannot be an invalid date.'
AssertionError [ERR_ASSERTION]: 'Connection terminated' == 'Query parameter value cannot be an invalid date.
```

charmander: That’s a serious bug, then.

packages/pg/test/integration/client/error-handling-tests.js
```diff
@@ -257,3 +257,18 @@ suite.test('cannot pass non-string values to query as text', (done) => {
     })
   })
 })
+
+if (!helper.args.native) {
```
ikonst, line 261: What happens in the 'native' case?

brianc:
hmmm...I'm not sure we should be in the business of validating the date passed to the library isn't invalid.  The backend already does this w/ an informative error saying the date is bad. If this resulted in insert garbage or doing something unexpected I'd be more concerned, but this _already_ results in an error.

ikonst:
The way it fails is with this cryptic date string. If you're intent on sending it, you can serialise it as "InvalidDate" or something like that, so that the cause would be clearer.

PR #3333, opened 2024-10-22
add cching

louieheaton:


boromisp:
FYI, dnscache seems to be abandoned: https://github.com/yahoo/dnscache/issues/28

louieheaton:
> FYI, dnscache seems to be abandoned: [yahoo/dnscache#28](https://github.com/yahoo/dnscache/issues/28)

Yeah @boromisp  - got any alternatives ? We need a caching strategy and this is really the only one that seemingly _works_. We already have NodeLocal on our K8s environments.

boromisp:
> > FYI, dnscache seems to be abandoned: [yahoo/dnscache#28](https://github.com/yahoo/dnscache/issues/28)
> 
> Yeah @boromisp - got any alternatives ? We need a caching strategy and this is really the only one that seemingly _works_. We already have NodeLocal on our K8s environments.

Sorry, I've never used anything like this before. There are a bunch of options but none of them stands out.

charmander:
I think this should be implemented in the form of an option to provide any user-defined DNS lookup function.

louieheaton:
We've solved this now internally with a custom approach, but may not be a bad thing to think about. I'll close for now as I think there are some workarounds that can be done externally to the package. But the ability to provide a DNS lookup function is not a bad shout

PR #3336, opened 2024-10-30
chore: align pg-types of pg-native to the same version of vanilla pg

cesco69:
This PR align the version of `pg-types` into `pg-native` to the same version of vanilla `pg` ( 2.1.0 see https://github.com/brianc/node-postgres/blob/master/packages/pg/package.json#L26C18-L26C24 )


brianc (approved):
ohhhh yeah nice find, nice fix! TY

PR #3338, opened 2024-10-31
chore: automatically replace var with let or const

DavideViolante:
- add eslint rule no-var
- add eslint rule prefer-const
- run npm run lint --fix to automatically fix the code

Note: I didn't fix any code manually, just ran eslint.


packages/pg-protocol/src/inbound-parser.test.ts
```diff
@@ -39,58 +39,58 @@ var twoRowBuf = buffers.rowDescription([
   },
 ])
 
-var emptyRowFieldBuf = new BufferList().addInt16(0).join(true, 'D')
+let emptyRowFieldBuf = new BufferList().addInt16(0).join(true, 'D')
 
-var emptyRowFieldBuf = buffers.dataRow([])
+emptyRowFieldBuf = buffers.dataRow([])
```
DavideViolante, line 44: The only manual change was here due to: cannot redeclare variable. Same at line 52.
I think the line 42 and 46 can be safely removed because they are declared twice but only the 2nd var is used.

brianc:
I really like this...I'm a bit worried about breaking other PRs so lemme go through & merge a few, then I'll merge this. Any additional outstanding PRs should be able to aut-fix themselves w/ `yarn lint:fix` which is 👍

DavideViolante:
Resolved the conflicts, mergiable again.

DavideViolante:
Resolve conflicts 
cc @brianc

PR #3339, opened 2024-10-31
chore: update README.md, add how to install

### LINKED ISSUE #3314: Clarify that this is not `node-postgres` (on npm)?
Hi - I just created https://github.com/fuxingZhang/postgres/issues/2 because for the Nth time I accidentally went to the wrong npm package.

I wonder if this library should just stop referring to itself as node-postgres, since the way users install is `npm install pg`, and not `npm install node-postgres`. I know the website is node-postgres.com but uhh maybe that could eventually not be the case either.

DavideViolante:
Fix #3314


brianc (approved):
simple, but i like it :p

Issue #3325, opened 2024-10-02
Horrible query performance since 8.2.0+


francois-egner:
I was trying out drizzle, using a PostgreSQL database (v15). I experience a huge performance issue just querying one row of a table with just one row. So I dig a little deeper inside the code and found out that it internally calls the `query` method of the `Client` class. I also tried to call the `query` method of the `Client` class with a plain query text/string which resulted in a way faster performance. I was able to reproduce everything using the plain pg module:

```js
const pg = new Client({
   connectionString: configuration.get('DATABASE_URL')
});

pg.connect();
```
```js
const config = {
  text: 'select count(*) as count from matcher.swipes where id = $1',
  rowMode: 'array'
}
const result = await pg.query(config, ['3a1ec2e5-41b9-4866-a210-f5102f66ca4d'])
```
This `pg.query` call takes 40-50ms!

```js
const result = await pg.query("select count(*) as count from matcher.swipes where id = '3a1ec2e5-41b9-4866-a210-f5102f66ca4d';")
```
This `pg.query` call takes 2-3ms

By downgrading over and over again I found out that version `8.1.0 ` is the last version that produces the same timing results for both `query ` calls. Now I wonder what I am missing here. I wonder what I am doing wrong and why it seems to only be me having this problem...

Dimitris-Tzilopoylos:
Hi there,

I believe this issue might be a duplicate of [#3098](https://github.com/brianc/node-postgres/issues/3098). It seems to cover similar concerns.

charmander:
(8.1.0 → 8.2.0 is when the entire protocol message parsing/serialization was rewritten into pg-protocol.)

This probably isn’t related to the issue, but note that `Client#connect()` is asynchronous; to avoid involving the query queue, make sure to `await pg.connect()`.

francois-egner:
> (8.1.0 → 8.2.0 is when the entire protocol message parsing/serialization was rewritten into pg-protocol.)
> 
> This probably isn’t related to the issue, but note that `Client#connect()` is asynchronous; to avoid involving the query queue, make sure to `await pg.connect()`.

Thanks for the hint. I verified that this isnt the cause for my problem. But indeed I should await this for sure :)

francois-egner:
> Hi there,
> 
> I believe this issue might be a duplicate of [#3098](https://github.com/brianc/node-postgres/issues/3098). It seems to cover similar concerns.

[...]

Issue #3332, opened 2024-10-17
pg-native don't works on Node 23


cesco69:
@brianc 

Install `pg-native` fails on Node 23 because `pg-native` depends on `node-libpq` that depends on `nan`  2.19.0.
`nan` 2.19.0 official support node at version 21 (see https://www.npmjs.com/package/nan/v/2.19.0 ) but also works on node 22, but doesn't works on node 23 due to this issue https://github.com/nodejs/nan/issues/953 fixed on nan 2.22.0 (see https://www.npmjs.com/package/nan/v/2.22.0). 


### How fix it
Fixing `pg-native` on node 23 require to update `nan` dependecies on `node-libpq` (see this PR https://github.com/brianc/node-libpq/pull/103 ), releasing a new version and update `pg-native` dependecies of `node-libpq`.


### Workaround
The workaround is override `nan` peer dependencies into package.json, eg (with `pnpm`):

```json
{
     "pnpm": {
        "overrides": {
            "nan": "2.22.0"
        }
    }
}
```

### Step to riproduce
Create a `Dockerfile` file with node 23
```docker
FROM node:23-slim

ENV DEBIAN_FRONTEND noninteractive
RUN apt-get update && apt-get -y install git python3 libpq-dev g++ make dos2unix && rm -rf /var/lib/apt/lists/* /var/cache/apt/archives/*

# install pg-native
RUN npm i -g pg-native
```

run docker build
```shell
docker build
```

Output
<details>
<summary>show</summary>

```shell
docker build
[+] Building 56.4s (7/7) FINISHED
 => [internal] load build definition from Dockerfile                                                                                                                                                                                                                         0.0s
 => => transferring dockerfile: 341B                                                                                                                                                                                                                                         0.0s 
 => [internal] load .dockerignore                                                                                                                                                                                                                                            0.0s 
 => => transferring context: 34B                                                                                                                                                                                                                                             0.0s 
 => [internal] load metadata for docker.io/library/node:23-slim                                                                                                                                                                                                              0.7s 
 => CACHED [1/4] FROM docker.io/library/node:23-slim@sha256:70092ae76d3b1a66dc51d659fd8882c42b2d816333c45ca4f2367ff0628e20a4                                                                                                                                                 0.0s
[...]

PR #3340, opened 2024-11-01
Batch query packets to avoid additional latency

thedadow451:
This fixes issue described in https://github.com/brianc/node-postgres/issues/3325
It seems like batching packages helps avoiding additional ~40ms latency (even on local network) for queries that require bind params.

I wrote short bench to reproduce the problem locally https://gist.github.com/thedadow451/b1ed04441bfa6fdf70a7b1c56504e988

It takes 60-70ms after this fix and ~42**s** without the fix. 

I imagine query path is already good covered with tests and since it's just optimization and not new feature I'm not adding any more tests. (Let me know if there are any you would like to have, I can add them)

Cheers!

### DIFF
--- a/packages/pg/lib/connection.js
+++ b/packages/pg/lib/connection.js
@@ -152,6 +152,44 @@ class Connection extends EventEmitter {
     this._send(serialize.query(text))
   }
 
+  // query using single batch of packets
+  // https://github.com/brianc/node-postgres/issues/3325
+  queryWithPacketBatching(query, valueMapper) {
+    const packets = []
+
+    if (!query.hasBeenParsed(this)) {
+      packets.push(serialize.parse({
+        text: query.text,
+        name: query.name,
+        types: query.types,
+      }))
+    }
+
+    packets.push(serialize.bind({
+      portal: query.portal,
+      statement: query.name,
+      values: query.values,
+      binary: query.binary,
+      valueMapper: valueMapper,
+    }))
+
+    packets.push(serialize.describe({
+      type: 'P',
+      name: query.portal || '',
+    }))
+
+    packets.push(serialize.execute({
+      portal: query.portal,
+      rows: query.rows,
+    }))
+
+    if (!query.rows) {
+      packets.push(syncBuffer)
+    }
+
+    this._send(Buffer.concat(packets))
+  }
+
   // send parse message
   parse(query) {
     this._send(serialize.parse(query))
--- a/packages/pg/lib/query.js
+++ b/packages/pg/lib/query.js
@@ -161,7 +161,7 @@ class Query extends EventEmitter {
       return new Error('Query values must be an array')
     }
     if (this.requiresPreparation()) {
-      this.prepare(connection)
+      connection.queryWithPacketBatching(this, utils.prepareValue)
     } else {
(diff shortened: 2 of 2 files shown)


brianc:
> I imagine query path is already good covered with tests and since it's just optimization and not new feature I'm not adding any more tests. (Let me know if there are any you would like to have, I can add them)

yah _certainly_ covered in tests already. :p That's really interesting this saves so much time. I'll run a few benches I have locally over the weekend & then ship this because that seems like an actual huge & significant speedup for things. Really sorry for the delay - was lost in some life stuff for a bit.  Thanks so much for this.

charmander:
Side note: Node streams’ [`writable.cork()`](https://nodejs.org/api/stream.html#writablecork) is a nice way to do this without extra copying and possibly without duplicating the implementation, but I don’t know if that introduces compatibility issues with pg-cloudflare now.

packages/pg/lib/connection.js
```diff
@@ -152,6 +152,44 @@ class Connection extends EventEmitter {
...
+      }))
+    }
+
+    packets.push(serialize.bind({
+      portal: query.portal,
+      statement: query.name,
+      values: query.values,
+      binary: query.binary,
+      valueMapper: valueMapper,
+    }))
```
charmander, line 174: This seems to have lost the `try`? I’m not sure if it’s a bug in #1855’s test or if the rewrite to pg-protocol or some other change made the `try` unnecessary, but it should probably be consistent with `prepare`.

packages/pg/lib/connection.js
```diff
@@ -152,6 +152,44 @@ class Connection extends EventEmitter {
...
+
+    packets.push(serialize.execute({
+      portal: query.portal,
+      rows: query.rows,
+    }))
+
+    if (!query.rows) {
+      packets.push(syncBuffer)
+    }
```
charmander, line 188: What about the Flush message when `query.rows` is present?

charmander (changes requested):

Phara0h (approved):

Phara0h:
Any way we can get this this pushed into a pre-release npm version for the mean time if we are not going to get it merged into master right away please @brianc

cesco69:
> Hello everyone, this bug has existed for a very long time now (since 8.1+) and was last officially reported in October 2024. The fix is available since November 2024. Considering the fact that this bug is an absolute no go for productive applications, I wonder why it is not acted upon with absolute priority. What is the reason? Is more help needed to release the patch?

Me too, I notice in this project a lot of slowness in releasing new patches/features. How can we help?

Another example is "pg-native don't works on Node 23" ( https://github.com/brianc/node-postgres/issues/3332 ) opened at Oct 17, 2024 fixed/merged but not released

brianc:
Sorry for the delay. I should have this fixed & release either tomorrow or Monday. Thanks for your patience!

brianc:
Ehh I might try to use the `writable.cork()` approach....

brianc:
looks like the cork method works. I am not _yet_ able to reproduce the slow down. I think someone said it was related to windows or WSL? Anyways i'm installing WSL on a VM and will test on windows too & see.

PR #3345, opened 2024-11-13
Extend test for issue 3174 - add cursor cases

joacoc:
This PR is just an extension for https://github.com/brianc/node-postgres/pull/3289. It aims to solve the message ordering issues described in #3174 for [this single particular case](https://github.com/brianc/node-postgres/issues/3174#issuecomment-2473394510) involving the use of cursors.

### DIFF
--- a/packages/pg/test/integration/gh-issues/3174-tests.js
+++ b/packages/pg/test/integration/gh-issues/3174-tests.js
@@ -3,6 +3,7 @@ const buffers = require('../../test-buffers')
 const helper = require('../test-helper')
 const assert = require('assert')
 const cli = require('../../cli')
+const Cursor = require('../../../../pg-cursor')
 
 const suite = new helper.Suite()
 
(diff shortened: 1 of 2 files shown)


packages/pg/test/integration/gh-issues/3174-tests.js
```diff
@@ -106,6 +107,18 @@ const testErrorBuffer = (bufferName, errorBuffer) => {
       await assert.rejects(() => client.query('SELECTR NOW()'), 'Further queries on the client should reject')
     }
 
+    // Same run but using cursor
+    const cursor = await client.query(new Cursor('SELECT NOW()'))
```
charmander, line 111: The client is still in an error state from the previous test and `errorHit` is still true, so this doesn’t test anything (similar to the `SELECTR NOW()` that always rejects because of the typo, which I apparently didn’t submit the review for before it was merged, oops).

charmander (changes requested):

packages/pg/test/integration/gh-issues/3174-tests.js
joacoc: I added the suite cases for the cursor, and the pool test case for the cursor raised the following issue:

```
Out of order parseComplete on pool using cursors is catchable Message: Cannot read properties of null (reading 'push')
TypeError: Cannot read properties of null (reading 'push')
at Cursor.handleDataRow (/...node-postgres/packages/pg-cursor/index.js:121:16)
at Client._handleDataRow (/.../node-postgres/packages/pg/lib/client.js:376:22)
at Connection.emit (node:events:519:28)
```

I applied a similar fix in the cursor client to handle the null, but I am unsure if this is the correct approach since the test continues to fail at the pool level:
```
  Out of order parseComplete on simple query is catchable ✔
  Out of order parseComplete on extended query is catchable ✔
  Out of order parseComplete on pool is catchable ✔
  Out of order parseComplete on simple query using cursors is catchable ✔
  Out of order parseComplete on extended query using cursors is catchable ✔
  Out of order parseComplete on pool using cursors is catchable FAILED!
  Error: Received unexpected parseComplete message from backend. 
```

By the way, I updated `SELECTR NOW()` to `SELECT NOW()`

PR #3347, opened 2024-12-03
read dataTypeID and tableID as unsigned uint

alexDevBR:
The oids are supposed to be unsigned ints, according to the [docs](https://www.postgresql.org/docs/current/datatype-oid.html#:~:text=The%20oid%20type%20is%20currently,has%20few%20operations%20beyond%20comparison.).
This causes issues with oids larger than 2^31 since they become negative.
This is causing issues in other projects, like https://github.com/sequelize/sequelize/issues/15466#issuecomment-2514805522

### DIFF
--- a/packages/pg-protocol/src/buffer-reader.ts
+++ b/packages/pg-protocol/src/buffer-reader.ts
@@ -31,6 +31,12 @@ export class BufferReader {
     return result
   }
 
+  public uint32(): number {
+    const result = this.buffer.readUInt32BE(this.offset)
+    this.offset += 4
+    return result
+  }
+
   public string(length: number): string {
     const result = this.buffer.toString(this.encoding, this.offset, this.offset + length)
     this.offset += length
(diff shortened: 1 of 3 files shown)


charmander (approved):
I wonder if any other things [documented as Int*N*](https://www.postgresql.org/docs/current/protocol-message-formats.html) are secretly unsigned?

alexDevBR:
@charmander I made a request for the docs to be updated reflecting what fields are signed and which are unsigned. Can we merge this one, and when the request goes through and the docs are updated we can open another issue here to make the changes?

alexDevBR:
@charmander,is this ok to be merged?

charmander:
Only brianc can publish a new version of the package, so most times, at best it doesn’t really help anything for me to merge a PR (and at worst I missed something in review, so may as well wait for the chance to get a second review anyway).

alexDevBR:
Ok. The request I made for documentation doesn't seem like it's going anywhere, and I found someone made the same request in 2020 here: https://www.postgresql.org/message-id/159173854100.661.1819403154632752741%40wrigleys.postgresql.org, so it doesn't seem like anything is going to change.

brianc:
whoah yikes! awesome find - i'm sure that took some digging. Thank you!

PR #3356, opened 2025-01-10
Add support for SCRAM-SHA-256-PLUS i.e. channel binding

jawj:
Hi all. I hope you'll consider this patch, which adds support for SCRAM-SHA-256-PLUS.

SCRAM-SHA-256-PLUS on Postgres enables `tls-server-end-point` channel binding, where the client sends the server a hash of the certificate it received as part of the TLS handshake. This prevents some kinds of MITM attacks where the attacker obtains a certificate that appears valid for the server, but is not actually the server's.

So far I've tested it working against Neon (who support SCRAM-SHA-256-PLUS) and Supabase (who don't), and with the new `client.enableChannelBinding` flag both `true` and `false`.

Feel free to make any changes you think appropriate.

### DIFF
--- a/packages/pg/lib/client.js
+++ b/packages/pg/lib/client.js
@@ -20,7 +20,7 @@ class Client extends EventEmitter {
     this.database = this.connectionParameters.database
     this.port = this.connectionParameters.port
     this.host = this.connectionParameters.host
-    this.enableChannelBinding = false // set true to use SCRAM-SHA-256-PLUS when offered
+    this.enableChannelBinding = Boolean(config.enableChannelBinding) // set true to use SCRAM-SHA-256-PLUS when offered
 
     // "hiding" the password so it doesn't show up in stack traces
     // or if the client is console.logged
--- a/packages/pg/lib/client.js
+++ b/packages/pg/lib/client.js
@@ -20,6 +20,7 @@ class Client extends EventEmitter {
     this.database = this.connectionParameters.database
     this.port = this.connectionParameters.port
     this.host = this.connectionParameters.host
+    this.enableChannelBinding = true
 
     // "hiding" the password so it doesn't show up in stack traces
     // or if the client is console.logged
@@ -258,7 +259,7 @@ class Client extends EventEmitter {
   _handleAuthSASL(msg) {
     this._checkPgPass(() => {
       try {
-        this.saslSession = sasl.startSession(msg.mechanisms)
+        this.saslSession = sasl.startSession(msg.mechanisms, this.enableChannelBinding && this.connection.stream)
         this.connection.sendSASLInitialResponseMessage(this.saslSession.mechanism, this.saslSession.response)
       } catch (err) {
         this.connection.emit('error', err)
@@ -268,7 +269,7 @@ class Client extends EventEmitter {
 
   async _handleAuthSASLContinue(msg) {
     try {
-      await sasl.continueSession(this.saslSession, this.password, msg.data)
+      await sasl.continueSession(this.saslSession, this.password, msg.data, this.enableChannelBinding && this.connection.stream)
       this.connection.sendSCRAMClientFinalMessage(this.saslSession.response)
     } catch (err) {
       this.connection.emit('error', err)
--- a/packages/pg/lib/client.js
+++ b/packages/pg/lib/client.js
@@ -20,7 +20,7 @@ class Client extends EventEmitter {
     this.database = this.connectionParameters.database
     this.port = this.connectionParameters.port
     this.host = this.connectionParameters.host
-    this.enableChannelBinding = false  // set true to use SCRAM-SHA-256-PLUS when offered
+    this.enableChannelBinding = false // set true to use SCRAM-SHA-256-PLUS when offered
 
     // "hiding" the password so it doesn't show up in stack traces
     // or if the client is console.logged
@@ -269,7 +269,12 @@ class Client extends EventEmitter {
 
   async _handleAuthSASLContinue(msg) {
     try {
-      await sasl.continueSession(this.saslSession, this.password, msg.data, this.enableChannelBinding && this.connection.stream)
+      await sasl.continueSession(
+        this.saslSession,
+        this.password,
+        msg.data,
+        this.enableChannelBinding && this.connection.stream
+      )
       this.connection.sendSCRAMClientFinalMessage(this.saslSession.response)
     } catch (err) {
       this.connection.emit('error', err)
--- a/packages/pg/lib/client.js
+++ b/packages/pg/lib/client.js
@@ -20,7 +20,6 @@ class Client extends EventEmitter {
     this.database = this.connectionParameters.database
     this.port = this.connectionParameters.port
     this.host = this.connectionParameters.host
-    this.enableChannelBinding = Boolean(config.enableChannelBinding) // set true to use SCRAM-SHA-256-PLUS when offered
 
     // "hiding" the password so it doesn't show up in stack traces
     // or if the client is console.logged
@@ -44,6 +43,7 @@ class Client extends EventEmitter {
     this._connectionError = false
     this._queryable = true
 
+    this.enableChannelBinding = Boolean(c.enableChannelBinding) // set true to use SCRAM-SHA-256-PLUS when offered
     this.connection =
       c.connection ||
       new Connection({
--- a/packages/pg/lib/client.js
+++ b/packages/pg/lib/client.js
@@ -20,7 +20,7 @@ class Client extends EventEmitter {
     this.database = this.connectionParameters.database
     this.port = this.connectionParameters.port
     this.host = this.connectionParameters.host
-    this.enableChannelBinding = true
+    this.enableChannelBinding = false  // set true to use SCRAM-SHA-256-PLUS when offered
 
     // "hiding" the password so it doesn't show up in stack traces
     // or if the client is console.logged
--- a/packages/pg/lib/crypto/sasl.js
+++ b/packages/pg/lib/crypto/sasl.js
@@ -1,6 +1,7 @@
 'use strict'
 const crypto = require('./utils')
-const tls = require('tls');
+const tls = require('tls')
+const x509 = require('@peculiar/x509')
 
 function startSession(mechanisms, stream) {
   const candidates = ['SCRAM-SHA-256']
@@ -65,7 +66,6 @@ async function continueSession(session, password, serverData, stream) {
   // override if channel binding is in use:
   if (session.mechanism === 'SCRAM-SHA-256-PLUS') {
     const peerCert = stream.getPeerCertificate().raw
-    const x509 = await import('@peculiar/x509')
     const parsedCert = new x509.X509Certificate(peerCert)
     const sigAlgo = parsedCert.signatureAlgorithm
     if (!sigAlgo) {
--- a/packages/pg/lib/crypto/sasl.js
+++ b/packages/pg/lib/crypto/sasl.js
@@ -1,6 +1,6 @@
 'use strict'
 const crypto = require('./utils')
-const x509 = require('@peculiar/x509')
+const { signatureAlgorithmHashFromCertificate } = require('./cert-signatures')
 
 function startSession(mechanisms, stream) {
   const candidates = ['SCRAM-SHA-256']
@@ -59,20 +59,8 @@ async function continueSession(session, password, serverData, stream) {
   // override if channel binding is in use:
   if (session.mechanism === 'SCRAM-SHA-256-PLUS') {
     const peerCert = stream.getPeerCertificate().raw
-    const parsedCert = new x509.X509Certificate(peerCert)
-    const sigAlgo = parsedCert.signatureAlgorithm
-    if (!sigAlgo) {
-      throw new Error('Could not extract signature algorithm from certificate')
-    }
-    const hash = sigAlgo.hash
-    if (!hash) {
-      throw new Error('Could not extract hash from certificate signature algorithm')
-    }
-    let hashName = hash.name
-    if (!hashName) {
-      throw new Error('Could not extract name from certificate signature algorithm hash')
-    }
-    if (/^(md5|sha-?1)$/i.test(hashName)) hashName = 'SHA-256' // for MD5 and SHA-1, we substitute SHA-256
+    let hashName = signatureAlgorithmHashFromCertificate(peerCert)
+    if (hashName === 'MD5' || hashName === 'SHA-1') hashName = 'SHA-256'
     const certHash = await crypto.hashByName(hashName, peerCert)
     const bindingData = Buffer.concat([Buffer.from('p=tls-server-end-point,,'), Buffer.from(certHash)])
     channelBinding = bindingData.toString('base64')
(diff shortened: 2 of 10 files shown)


packages/pg/lib/crypto/sasl.js
```diff
@@ -40,7 +58,34 @@ async function continueSession(session, password, serverData) {
...
   var serverFirstMessage = 'r=' + sv.nonce + ',s=' + sv.salt + ',i=' + sv.iteration
-  var clientFinalMessageWithoutProof = 'c=biws,r=' + sv.nonce
+
+  // without channel binding:
+  let channelBinding = stream ? 'eSws' : 'biws' // 'y,,' or 'n,,', base64-encoded
+
+  // override if channel binding is in use:
+  if (session.mechanism === 'SCRAM-SHA-256-PLUS') {
+    const peerCert = stream.getPeerCertificate().raw
+    const x509 = await import('@peculiar/x509')
```
brianc, line 68: curious: why make this a lazy/dynamic import?

packages/pg/package.json
```diff
@@ -20,6 +20,7 @@
   "author": "Brian Carlson <brian.m.carlson@gmail.com>",
   "main": "./lib",
   "dependencies": {
+    "@peculiar/x509": "^1.12.3",
```
brianc, line 23: should absolutely pin this dependency - i only allow packages within this repo to float patch versions w/ semver. Anything external needs to be vetted & checked on exact versions, and using exact versions in package.json please!

packages/pg/test/integration/client/sasl-scram-tests.js
```diff
@@ -45,12 +45,25 @@ if (!config.user || !config.password) {
...
 }
 
-suite.testAsync('can connect using sasl/scram', async () => {
+suite.testAsync('can connect using sasl/scram (channel binding enabled)', async () => {
   const client = new pg.Client(config)
   let usingSasl = false
   client.connection.once('authenticationSASL', () => {
     usingSasl = true
   })
+  client.enableChannelBinding = true // default
```
brianc, line 54: would suggest making the default false - adding new defaults in a patch version of `pg` is a risk for introducing breaking changes

brianc (changes requested):
thanks for the PR! Put a few comments on here. My main concern is changing the default behavior - as much as its a pain on users to enable the feature, I'd rather it be an opt-in thing vs a default...otherwise it's a slight breaking change, which requires a new semver major.  Would be _nice_ to add a callout in the docs about this as well...though I can handle that if you don't want to

packages/pg/lib/crypto/sasl.js
jawj: I suppose only that it may often go unused. But that's not very compelling, so I've shifted it to a `require` at the top.

packages/pg/package.json
jawj: Sure.

packages/pg/test/integration/client/sasl-scram-tests.js
jawj: Fair point. In fact, even if you wanted to make it the default, it would probably make sense for it to get some testing in the wild by people who opt into it before doing that.

jawj:
Thanks @brianc. I've made those changes and added a short section to the docs.

However, while writing the docs it struck me that I have only provided the opt-in option directly on `Client`. How should this feature be exposed for those using `pool.query`?

packages/pg/lib/crypto/sasl.js
```diff
@@ -40,7 +59,33 @@ async function continueSession(session, password, serverData) {
...
   var serverFirstMessage = 'r=' + sv.nonce + ',s=' + sv.salt + ',i=' + sv.iteration
-  var clientFinalMessageWithoutProof = 'c=biws,r=' + sv.nonce
+
+  // without channel binding:
+  let channelBinding = stream ? 'eSws' : 'biws' // 'y,,' or 'n,,', base64-encoded
+
+  // override if channel binding is in use:
+  if (session.mechanism === 'SCRAM-SHA-256-PLUS') {
+    const peerCert = stream.getPeerCertificate().raw
+    const parsedCert = new x509.X509Certificate(peerCert)
```
charmander, line 69: Instead of a separate package, could `stream.getPeerCertificate().fingerprint256` be used with a fixed selection of SHA-256? Or is that not the same hash?

packages/pg/lib/client.js
```diff
@@ -20,6 +20,7 @@ class Client extends EventEmitter {
     this.database = this.connectionParameters.database
     this.port = this.connectionParameters.port
     this.host = this.connectionParameters.host
+    this.enableChannelBinding = false // set true to use SCRAM-SHA-256-PLUS when offered
```
charmander, line 23: Should read this from `config` (which is also the answer to the question of how to configure this when using a `Pool`).

```suggestion
    this.enableChannelBinding = Boolean(config.enableChannelBinding) // set true to use SCRAM-SHA-256-PLUS when offered
```

packages/pg/test/integration/client/sasl-scram-tests.js
```diff
@@ -45,12 +45,25 @@ if (!config.user || !config.password) {
   return
 }
 
-suite.testAsync('can connect using sasl/scram', async () => {
+suite.testAsync('can connect using sasl/scram (channel binding enabled)', async () => {
```
charmander, line 48: These tests would work just as well if there were no implementation of channel binding at all, so this PR probably needs a targeted test from someone.

charmander (changes requested):

packages/pg/lib/crypto/sasl.js
charmander: Never mind – [that wouldn’t be spec-compliant](https://datatracker.ietf.org/doc/html/rfc5929#section-4.1) and I missed that the `hash` wasn’t being used for anything more than its name anyway.

packages/pg/lib/crypto/sasl.js
```diff
@@ -1,22 +1,41 @@
...
+  const candidates = ['SCRAM-SHA-256']
+  if (stream) candidates.unshift('SCRAM-SHA-256-PLUS') // higher-priority, so placed first
+
+  let mechanism
+  for (const candidate of candidates) {
+    if (mechanisms.indexOf(candidate) !== -1) {
+      mechanism = candidate
+      break
+    }
+  }
```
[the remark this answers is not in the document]
jawj: I wasn't sure what node-postgres's policy was on compatibility, but `find` and `includes` look to have been widely available since 2016 and Node 6, which I'm guessing is good enough.

packages/pg/lib/crypto/sasl.js
```diff
@@ -40,7 +59,33 @@ async function continueSession(session, password, serverData) {
...
+    }
+    const hash = sigAlgo.hash
+    if (!hash) {
+      throw new Error('Could not extract hash from certificate signature algorithm')
+    }
+    let hashName = hash.name
+    if (!hashName) {
+      throw new Error('Could not extract name from certificate signature algorithm hash')
+    }
+    if (/^(md5)|(sha-?1)$/i.test(hashName)) hashName = 'SHA-256' // for MD5 and SHA-1, we substitute SHA-256
```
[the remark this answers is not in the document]
jawj: Yes!

packages/pg/lib/crypto/sasl.js
```diff
@@ -1,22 +1,41 @@
...
+      mechanism = candidate
+      break
+    }
+  }
+
+  if (!mechanism) {
+    throw new Error('SASL: Only mechanisms ' + candidates.join(' and ') + ' are supported')
+  }
+
+  if (mechanism === 'SCRAM-SHA-256-PLUS' && !(stream instanceof tls.TLSSocket)) {
```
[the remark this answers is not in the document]
jawj: Probably not, so let's change to this, which only checks that the `stream` object has the method we need:

```js
  if (mechanism === 'SCRAM-SHA-256-PLUS' && typeof stream.getPeerCertificate !== 'function') {
    // this should never happen if we are really talking to a Postgres server
    throw new Error('SASL: Mechanism SCRAM-SHA-256-PLUS requires a certificate')
  }
```

packages/pg/lib/client.js
jawj: Sounds good.

packages/pg/lib/crypto/sasl.js
jawj: I do appreciate that bringing in a whole X509 parsing library (and then parsing the whole certificate, when all we actually need is the signature algorithm) feels like overkill.

I did actually have a go at doing the minimum necessary parsing manually: see https://gist.github.com/jawj/04a90e51196ac054d6741c8d079d9cff

The reason I didn't go with that in the PR is that I haven't been able to find a list of either (a) what signature algorithms could theoretically be used or even (b) what signature algorithms would cover 99% of cases.

But I strongly suspect that the cases covered by this code would be almost all of them, and any missing ones might be plugged if people filed issues about them. So this could be another option?

packages/pg/package.json
jawj: See also comments below.

packages/pg/lib/crypto/sasl.js
brianc: > But I strongly suspect that the cases covered by this code would be almost all of them, and any missing ones might be plugged if people filed issues about them. So this could be another option?

I am absolutely totally down w/ the "make it mostly work and patch if different algorithms show up later" mode if it removes the requirement to pull in an entire dependency! I wouldn't say it's a mandatory change but certainly would be welcome. 😄 I have tried _very hard_ over the years to keep as many 3rd party dependencies out of the code as possible just because....well...left-pad and all that stuff, ya know?

brianc:
> How should this feature be exposed for those using `pool.query`?

hmm good question - would need to make it something that's thread through the pool to its create method as well.  The pool [passes its own constructor options to the client](https://github.com/brianc/node-postgres/blob/master/packages/pg-pool/index.js#L218) which probably if i were designing the API today from scratch I wouldn't do but..that ship sailed like 10 years ago 🙃  - so it should just be able to be set on the pool as well?

brianc:
@jawj ruh roh! looks like a test is failing (should be an easy fix).  There are a __lot__ of tests - some are duplicates, etc. If you just wanna run a subset locally for faster turn around you can run `make test-unit` and/or `make test-integration` from the `./packages/pg` subfolder.  In this case it was a unit test that broke (meaning you don't need an instance of postgres running locally)

packages/pg/lib/crypto/sasl.js
jawj: @brianc Good to hear: I'm all for that approach. I've done a bit more work on this code (including putting a base64 hash of the public cert in the error message, to make it easy to report failures in a way we can investigate) and removed the dependency.

packages/pg/test/integration/client/sasl-scram-tests.js
jawj: I think in an ideal world we would test using SSL certs with a variety of different signature algorithms. But I'm not sure how easily that fits with the current test setup.

Locally, I have been using:

* The `server-*-cert.pem` files from the OpenSSL tests: https://github.com/openssl/openssl/tree/master/test/certs

* This Docker file:

```bash
#!/bin/zsh
docker run --rm --name postgres-tls-cert \
  -p 5432:5432 \
  -e POSTGRES_USER=frodo \
  -e POSTGRES_PASSWORD=friend \
  -v ./certs:/etc/pgssl \
  postgres:17.2-bookworm \
  -c ssl=on \
  -c ssl_cert_file=/etc/pgssl/server-$1-cert.pem \
  -c ssl_key_file=/etc/pgssl/server-$1-key.pem
```

* And this test:

```bash
SCRAM_TEST_PGDATABASE=frodo \                                          
  SCRAM_TEST_PGUSER=frodo \
  SCRAM_TEST_PGPASSWORD=friend \
  PGSSLMODE=no-verify \
  node --tls-keylog=/path/to/keylog.txt packages/pg/test/integration/client/sasl-scram-tests.js
```

(The `--tls-keylog` enables the use of Wireshark if needed).

jawj:
I think the tests should be passing again now.

jawj:
BTW, you may find this interesting: https://bytebybyte.dev/?postgres

(Writing this was how I figured out how channel binding works).

jawj:
OK, this seems to have stalled a little in relation to testing.

To summarise the previous position:

* I had already added an integration test to ensure that SCRAM-SHA-256-PLUS is used (and the connection succeeds) when it's offered and channel binding is enabled. I think this may be failing in CI because an SSL connection is not used (`PGTESTNOSSL: true`). I guess I can add a check for SSL in the test, but of course that means it won't be testing what it's meant to in CI.

* What's not currently tested automatically (via integration tests) is compatibility with various certificate signature algorithms. However, I have tested successfully DSA, ECDSA, CECDSA, PSS and Ed25519 certificates locally (Ed448 and ecdsa-brainpoolP256r1 fail at the TLS-handshake stage, so the question of channel binding doesn't come up).

I think that last point would be nice to have, but involves larger changes to the testing/CI infrastructure than I'm comfortable making.

I've just added some more unit tests:

* A few unit tests to ensure that `startSession` returns the expected data when channel binding is and isn't enabled and SCRAM-SHA-256-PLUS is and isn't offered.

* A similar unit test for `continueSession` with `mechanism: "SCRAM-SHA-256-PLUS"`, which includes a minimal ASN.1 certificate structure that can be parsed for a signature hash algorithm.

As part of the testing, I realised I have not provided a way to enforce (rather than just enable) channel binding, but I think I'm minded to delay that to a future PR.

@brianc and @charmander, how do you think we can move this forward?

jawj:
Thanks @brianc. Tests are now passing! I just modified the SCRAM-SHA-256-PLUS test to only expect SCRAM-SHA-256-PLUS if we're on an SSL connection (because otherwise there's no certificate to check).

So maybe this is ready to merge?

As I said before, in an ideal world this would probably have CI tests (1) over an SSL connection and, even better, (2) with a variety of certificate signature algorithms. But I'm probablty not the right person to make those larger changes to the testing setup, and since this is an opt-in feature I think it's probably OK to go ahead without those for now.

jawj:
@brianc I think this is ready to merge (see above). Any remaining blockers?

brianc:
Wow this is really fabulous. Thanks so much for going out of your way to take the time to add this.  Just got back from a relatively long trip to south america. Merging this now! Will do a release tomorrow AM (I like to release in morning my time just in case something blows up I have more of the day to react to problems) 😄 ❤️

brianc:
@jawj published!

jawj:
Hi @brianc. I see [some changes are proposed](https://github.com/brianc/node-postgres/pull/3473#issuecomment-4476941121) for connection security in v9

As noted there, I'm not sure I like these specific changes very much! But while we're trying to be more like libpq, perhaps this would this be a good time to recognise libpq's `channel_binding=disable|prefer|require` option, and make `channel_binding=prefer` the default? At present (based on this PR of mine), users have to specifically opt in via the non-standard `enableChannelBinding` option, and I suspect that means that the vast majority do not get the security bump this ought to be giving them.

Issue #2012, opened 2019-11-26
JSONB arguments not treated consistently


jasperblues:
According to [docs](https://node-postgres.com/features/types), there is no special treatment required for JSONB bind arguments. In practice there is: 

* When the argument is a single object, it is bound correctly
* When the argument is an Array, it is necessary to call JSON.stringify on the args. 

Here's a test case that reproduces the issue: 

```javascript
const { Client } = require('pg');
const { inspect } = require('util');
async function doIt() {
    const client = new Client();
    console.log('Connecting...');
    await client.connect();
    console.log('Connected');
    const a = [
        {
            the: 'quick',
            brown: 'fox',
            jumps: 'over the',
            lazy: 'dog',
            some: { other: 'object', dilroy: ['hello', 'world'], numpty: new Date() }
        }
    ];
    console.log('Inserting...');
    const result = await client.query(`insert into sb_json(c1) values ($1) returning *`, [a]);
    console.log(`result = ${inspect(result, { depth: null })}`);
    console.log('Disconnecting...');
    await client.end();
}
doIt()
    .then(() => console.log('All done'))
    .catch(e => console.log(inspect(e)));
```

Current documented contract is that no special treatment is required. 

*Possible solution: 

* Document current behavior. This is the less ideal solution, as arguably it is better to be consistent. Never call JSON.stringify on args, or always do so. Not one case for arrays, and on case for single items. 
* Inspect the contents of the array. If it contains all primitive types insert for that type. Otherwise call JSON.stringify internally. This operation would run in O(n) time - for a very large array it would be slow.

charmander:
> Likewise sending **an object** to the PostgreSQL server via a query from node-postgres, node-posgres will call JSON.stringify on your outbound value, automatically converting it to json for the server.

With the existence of `jsonb[]` and without knowing the desired types ahead of time, no solution by inspecting JavaScript types can work.

See also #442

jasperblues:
> Yeah, I see what you're saying. It's a bit funky that node-pg assumes to stringify Objects but not Arrays. I think making it more consistent, or rather, no incoming magic (I like the magic coming out though!) would be for the best. It seems a little arbitrary, especially now that JSON has been added to the data-type mix, that JS arrays are assumed to be PG Arrays, when they could be type json or even hstore.
[...]

PR #3360, opened 2025-01-22
Support .toPostgres() on Array-derived objects

mordae:
Related to #2012.

Since we cannot disambiguate between Arrays that should be converted to PostgreSQL array types and Arrays that should be treated as JSON, make it at least possible to define custom Array-derived types.

This also makes it possible to properly serialize Array-derived multirange collections where elements must not be independently escaped.

### DIFF
--- a/packages/pg/lib/utils.js
+++ b/packages/pg/lib/utils.js
@@ -66,7 +66,7 @@ var prepareValue = function (val, seen) {
       return dateToString(val)
     }
   }
-  if (Array.isArray(val)) {
+  if (Array.isArray(val) && val.toPostgres !== 'function') {
     return arrayString(val)
   }
   if (typeof val === 'object') {


charmander (changes requested):
Code is not correct or tested. Also, inheriting from `Array` is pretty niche, creating `Array.prototype.toPostgres` is a really bad idea, and the existing alternatives to achieve the same goal (global or one-off custom serialization functions) seem pretty good to me.

mordae:
@charmander I don't follow. How is the condition incorrect? In the sole case of Array instance with `toPostgres` method it falls through and gets prepared as other objects with `toPostgres` method.

Regardless of your judgement on goodness or badness of extending built-in prototypes, users are going to, when they need to. This change just makes `toPostgres` handling uniform for all "objects" like one would expect it to be.

You were right about Array subclassing being... niche... Or rather utterly broken. TIL.

charmander:
> How is the condition incorrect?

It’s missing `typeof`.

> Regardless of your judgement on goodness or badness of extending built-in prototypes, users are going to, when they need to.

Well, users can’t do that in this case because we don’t currently support it, and I’m making an argument against adding support for it.

mordae:
> > How is the condition incorrect?
> 
> It’s missing `typeof`.

🤦‍♂️ thanks. I must have been asleep.

> > Regardless of your judgement on goodness or badness of extending built-in prototypes, users are going to, when they need to.
> 
> Well, users can’t do that in this case because we don’t currently support it, and I’m making an argument against adding support for it.

If you see the linked issue, users have been doing exactly that by modifying the method on the fly.

But OK, if you disagree, feel free to close the issue. This was preferable solution for me, but I can probably solve my use case in other ways or apply the patch locally.

PR #3370, opened 2025-02-06
perf(utils): fast prepareValue

cesco69:
This PR add a performance improvements at prepare Value for non-object by skipping useless condition

### DIFF
--- a/packages/pg/lib/utils.js
+++ b/packages/pg/lib/utils.js
@@ -45,31 +45,32 @@ function arrayString(val) {
 // note: you can override this function to provide your own conversion mechanism
 // for complex types, etc...
 var prepareValue = function (val, seen) {
-  // null and undefined are both null for postgres
-  if (val == null) {
-    return null
-  }
-  if (val instanceof Buffer) {
-    return val
-  }
-  if (ArrayBuffer.isView(val)) {
-    var buf = Buffer.from(val.buffer, val.byteOffset, val.byteLength)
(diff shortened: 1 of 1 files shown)


packages/pg/lib/utils.js
```diff
@@ -45,31 +45,32 @@ function arrayString(val) {
 // note: you can override this function to provide your own conversion mechanism
 // for complex types, etc...
 var prepareValue = function (val, seen) {
-  // null and undefined are both null for postgres
-  if (val == null) {
-    return null
-  }
```
charmander, line 51: I would prefer preserving this for readability and to avoid having to check `typeof val === 'undefined'`.

charmander (changes requested):
I can’t reproduce those benchmark numbers with that link. Microbenchmarks can be misleading and that website seems unreliable. Like #3372, it’s fine as just cleanup anyway, though.

packages/pg/lib/utils.js
cesco69: There is a test `prepareValues: undefined prepared properly` https://github.com/brianc/node-postgres/blob/master/packages/pg/test/unit/utils-tests.js#L93 that check if `undefined` value return `null`. Whitout `typeof val === 'undefined'` my PR fails this test!

cesco69: done

brianc:
Looks good to me! thanks for this. 😄

Issue #3373, opened 2025-02-10
`TypeError: Cannot set property message of #<_ErrorEvent> which has only a getter` on connection timeout with `@neondatabase/serverless` library


andyjy:
Using [`@neondatabase/serverless`](https://github.com/neondatabase/serverless) (which extends `pg` `Client` and `Pool`) I encountered the following fatal error thrown within `pg-pool`:

```TypeError: Cannot set property message of #<_ErrorEvent> which has only a getter```

This occurs since the exception passed to the connection callback for Pool is of type `ErrorEvent` (from the Cloudflare Workers websocket implementation) instead of `Error`, and `ErrorEvent` [has a readonly message property that can't be overwritten](https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent#instance_properties).

While a fix could be written into @neondatabase/serverless to only pass an `Error` as `pg-pool` currently expects, it seemed more useful to address this here in `pg-pool` to be tolerant of different types for the exception in general, while preserving the original error message instead of overwriting it as per the current implementation. 

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/pg-pool/index.js b/node_modules/pg-pool/index.js
index 94004e0..52c595a 100644
--- a/node_modules/pg-pool/index.js
+++ b/node_modules/pg-pool/index.js
@@ -244,7 +244,7 @@ class Pool extends EventEmitter {
         // remove the dead client from our list of clients
         this._clients = this._clients.filter((c) => c !== client)
         if (timeoutHit) {
-          err.message = 'Connection terminated due to connection timeout'
+          err = new Error('Connection terminated due to connection timeout', { cause: err });
         }
 
         // this client won’t be released, so move on immediately
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

PR #3374, opened 2025-02-10
fix(pg-pool): don't throw on readonly .message e.g. ErrorEvent, preserve original exception

andyjy:
Fixes https://github.com/brianc/node-postgres/issues/3373

### DIFF
--- a/packages/pg-pool/index.js
+++ b/packages/pg-pool/index.js
@@ -244,7 +244,7 @@ class Pool extends EventEmitter {
         // remove the dead client from our list of clients
         this._clients = this._clients.filter((c) => c !== client)
         if (timeoutHit) {
-          err = new Error('Connection terminated due to connection timeout', { cause: err } )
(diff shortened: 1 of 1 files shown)


packages/pg-pool/index.js
```diff
@@ -244,7 +244,7 @@ class Pool extends EventEmitter {
         // remove the dead client from our list of clients
         this._clients = this._clients.filter((c) => c !== client)
         if (timeoutHit) {
-          err.message = 'Connection terminated due to connection timeout'
+          err = new Error('Connection terminated due to connection timeout', { cause: err } )
```
andyjy, line 247: Node.js support for `Error.cause` added in 16.9.0, prior versions will silently ignore - has the effect of losing the stack trace of the original exception under these older Node.js versions

brianc:
Thanks! I'll push a new patch version out tomorrow morning. 😄

PR #3382, opened 2025-02-12
fix: revert PR #3379 (avoid useless spread)

cesco69:
see https://github.com/brianc/node-postgres/pull/3379#issuecomment-2653679598


brianc:
sorry about that, folks. releasing new patch version now.

cesco69:
> sorry about that, folks. releasing new patch version now.

Thanks and sorry for my mistake!

PR #3387, opened 2025-02-15
Fix connection timeout test for proper error handling

armaan4658:
This PR fixes the expired connection timeout test by:

- Ensuring ConnectionTimeoutError is properly handled.
- Clearing the timeout in both success and failure cases to prevent unexpected errors.
- Avoiding calling client.end() when the connection is not established.

These improvements make the test more reliable and prevent false positives. Let me know if any additional changes are needed! 🚀


charmander:
What is this, some kind of AI-generated pull request description? It doesn’t line up with the actual content of the pull request. Please write a meaningful and original justification.

Issue #3397, opened 2025-03-04
feat(pg-pool): Optimize client retrieval from pool by prioritizing free client with cached prepared statements


cesco69:
Hi, 

Instead of retrieving the first available client from the pool, would it be possible to first check if there is a free client that already has cached prepared statements? This could improve performance by reducing the need to re-prepare statements on different clients.

I have made a PR https://github.com/brianc/node-postgres/pull/3398

PR #3398, opened 2025-03-04
perf(pg-pool): optimize client retrieval from pool by prioritizing prepared client

cesco69:
see https://github.com/brianc/node-postgres/issues/3397

Instead of retrieving the first available client from the pool, check if there is a free client that already has cached prepared statements. This could improve performance by reducing the need to re-prepare statements on different clients.

### DIFF
--- a/packages/pg-pool/index.js
+++ b/packages/pg-pool/index.js
@@ -111,7 +111,7 @@ class Pool extends EventEmitter {
     return this._clients.length >= this.options.max
   }
 
-  _pulseQueue() {
+  _pulseQueue(name) {
     this.log('pulse queue')
     if (this.ended) {
       this.log('pulse queue ended')
@@ -142,7 +142,10 @@ class Pool extends EventEmitter {
     }
     const pendingItem = this._pendingQueue.shift()
     if (this._idle.length) {
-      const idleItem = this._idle.pop()
+      let idleItem = name ? this._idle.find((item) => item.client.connection.parsedStatements[name]) : null
+      if (!idleItem) {
+        idleItem = this._idle.pop()
+      }
       clearTimeout(idleItem.timeoutId)
       const client = idleItem.client
       client.ref && client.ref()
@@ -168,7 +171,7 @@ class Pool extends EventEmitter {
     this.emit('remove', client)
   }
 
-  connect(cb) {
+  connect(cb, name) {
     if (this.ending) {
       const err = new Error('Cannot use a pool after calling end on the pool')
       return cb ? cb(err) : this.Promise.reject(err)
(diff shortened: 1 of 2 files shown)


packages/pg-pool/index.js
```diff
@@ -142,7 +142,10 @@ class Pool extends EventEmitter {
     }
     const pendingItem = this._pendingQueue.shift()
     if (this._idle.length) {
-      const idleItem = this._idle.pop()
+      let idleItem = name ? this._idle.find((item) => item.client.connection.parsedStatements[name]) : null
```
charmander, line 145: `find` doesn’t remove the element it finds.

charmander (changes requested):
Needs tests.

packages/pg-pool/index.js
cesco69: Oh! you are absolutely right! made it with https://github.com/brianc/node-postgres/pull/3398/commits/498d7f47e6877a941fed99ab04e346d3d3bb3e45

cesco69:
> Needs tests.

see
https://github.com/brianc/node-postgres/pull/3398/files#diff-939dc0e2d2a5de32423ce5295bf1adbfb7c63bf67637efe3fc5fe334cbfa21a9

I'm not convinced this is a fair test. In fact, I should check that the query was executed with the same client, but how can I tell if it's the same client?

brianc:
> I'm not convinced this is a fair test. In fact, I should check that the query was executed with the same client, but how can I tell if it's the same client?

You can call `pg_backend_pid()` in a query - that should return the connection ID which will certainly be the same across queries on the same client.

cesco69:
> > I'm not convinced this is a fair test. In fact, I should check that the query was executed with the same client, but how can I tell if it's the same client?
> 
> You can call `pg_backend_pid()` in a query - that should return the connection ID which will certainly be the same across queries on the same client.

Thanks! done with https://github.com/brianc/node-postgres/pull/3398/commits/20e7fa2fcc3b9c48b03a8661dc71f8284d50f0e1

brianc:
While I appreciate the thought here....I'm thinking this might be a bit over-optimization and prioritizing a somewhat lesser used use-case over the general purpose.  For example, applications that don't use named queries will incur an additional for-loop over the connection pool every time issuing a query via `pool.query.` It also complicates the pool's api surface slightly by including the name of the query more prominently....which could potentially impact future enhancments like "auto-prepare named statements" where we use md5s of query texts to manage the preparation or something fancier.  Finally, and IMO most importantly: in any application under sufficient load, in which this feature would expect to potentially cause performance improvements, your pool will likely be saturated and since every client will be "hot," eventually every client will end up w/ the named queries stored & cached anyway....making this code not really needed.  I think the largest measurable performance impact above and beyond what using named prepared statements already makes in applications underload is likely in micro-benchmarks....which isn't worth complicating the API surface for.

PR #3402, opened 2025-03-15
Batch network packets in prepared statements

### LINKED ISSUE #3098: nodejs v18.17.1 / v20.9.0 - pg ^8.11.0 - slow select query
I am using the following query to retrieve users from database, and before this one i just perform a count(*) to retrieve the total number of users.
```
SELECT coalesce(json_agg(_0_users),'[]') as users 
        FROM (
          SELECT row_to_json((
            SELECT _0_users
            FROM ( SELECT _0_users.id,_0_users.email,_0_users.password,_0_users.name,_0_users.last_name,_0_users.created_at,_0_users.verified,_0_users.blacklist,_0_users.profile_image,_0_users.role_id,_0_users.country_id,_0_users.city,_0_users.phone,_0_users.address,_0_users.postal_code,_0_users.is_business,_0_users.business_name,_0_users.business_address,_0_users.business_country_id,_0_users.vat,_0_users.parent_id) _0_users )) _0_users
            FROM (
              SELECT  _0_users.id,_0_users.email,_0_users.password,_0_users.name,_0_users.last_name,_0_users.created_at,_0_users.verified,_0_users.blacklist,_0_users.profile_image,_0_users.role_id,_0_users.country_id,_0_users.city,_0_users.phone,_0_users.address,_0_users.postal_code,_0_users.is_business,_0_users.business_name,_0_users.business_address,_0_users.business_country_id,_0_users.vat,_0_users.parent_id FROM public.users _0_users
       ) _0_users
       ) _0_users
```
The execution with nodejs v18.17.1 / v20.9.0 - pg ^8.11.0  results to something between 60 and 92 ms for both count and the query mentioned above, but the same process with nodejs v16.20.2  results to 12ms for both queries (average).
Are there any issues with node-postgres and nodejs versions > 16? Am I missing something?

### LINKED ISSUE #3325: Horrible query performance since 8.2.0+

brianc:
Fixes #3340
Fixes #3325
Fixes #3098

### DIFF
--- a/packages/pg/bench.js
+++ b/packages/pg/bench.js
@@ -47,21 +47,27 @@ const run = async () => {
   for (let i = 0; i < 4; i++) {
     let queries = await bench(client, params, seconds * 1000)
     console.log('')
-    console.log('little queries:', queries)
+    console.log('param queries:', queries)
     console.log('qps', queries / seconds)
-    console.log('on my laptop best so far seen 733 qps')
+    console.log('on my laptop best so far seen 987 qps')
+
+    queries = await bench(client, { ...params, name: 'params' }, seconds * 1000)
+    console.log('')
+    console.log('named queries:', queries)
+    console.log('qps', queries / seconds)
+    console.log('on my laptop best so far seen 937 qps')
 
     console.log('')
     queries = await bench(client, seq, seconds * 1000)
     console.log('sequence queries:', queries)
     console.log('qps', queries / seconds)
-    console.log('on my laptop best so far seen 1309 qps')
+    console.log('on my laptop best so far seen 2725 qps')
 
     console.log('')
     queries = await bench(client, insert, seconds * 1000)
     console.log('insert queries:', queries)
     console.log('qps', queries / seconds)
-    console.log('on my laptop best so far seen 6445 qps')
+    console.log('on my laptop best so far seen 27383 qps')
 
     console.log('')
     console.log('Warming up bytea test')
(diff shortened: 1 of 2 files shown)


brianc:
This actually did result in a small speedup on my version of node and osx...though maybe 5-10% (there's so much variance its hard to tell what amount is code change related and what is just "less operating system interrupts got in the way during the run") but there did seem to be a small _consistent_ speedup which is cool!  The bigger issue is on some operating system / node combos (wasn't able to identify any myself) there is a much more significant speedup this should offer.

brianc:
Special thanks, as always, to @charmander for the tip about `stream.cork()`! :) I'll leave this up for a day waiting for feedback then push a new semver patch.

brianc:
published in `pg@8.14.1`

Issue #2336, opened 2020-09-07
Feature: Memory limits


pauldraper:
I'd like a large result to produce an Error (which can be handled), rather than crashing the entire process with OutOfMemory.

This is similar in nature to limit I/O for HTTP requests:

```
const express = require('express')
const bodyParser = require('body-parser')

const app = express()
app.use(bodyParser.json({ limit: '5mb' }))
```

I imagine an option to Connection/Pool like `maxResultSize`. When the result exceeds the configured number of bytes, an Error is thrown. (I understand this limit would be subject to the choice of binary and text formats...this isn't super precise, just a way to gracefully handle unexpectedly large results.)

This would significantly improve the stability of multi-operation processes using this library.

sehrope:
Are you concerned with a single row / field being too large or an entire result set (returned as an array) being too large?

If it's the total result set, have you considered using a cursor? That would let you operate on the row in chunks without needing to allocate memory for the entire result set. If you have commands that could return an arbitrary / huge number of rows, it's the way to go to ensure you do not blow up the node process memory usage.

pauldraper:
Both. Yes, I could use a cursor.

But not I'm interested in the complexity of streaming the results, *or even getting large results set to work.* I'm interested in erroring in a sane (catchable, non-crash) way.

---

It's the same rationale as limiting HTTP message size. Yes, you could stream the HTTP body. However that is complex. Libraries can be configure to just cancel/error the operation.

In my specific case, I have a server making PostgreSQL queries. If the response is ever too large, the entire server crashes. Instead, I'd like to cancel/error the specific offending operation.

brianc:
> If it's the total result set, have you considered using a cursor? That would let you operate on the row in chunks without needing to allocate memory for the entire result set. If you have commands that could return an arbitrary / huge number of rows, it's the way to go to ensure you do not blow up the node process memory usage.

Yup, 100% this is the way to handle this!  The postgres protocol supports this, and node is obvs build w/ streaming in mind.  This library [already exists](https://github.com/brianc/node-postgres/tree/master/packages/pg-cursor) as well.  We use pg and regularly stream 500-600 megabytes of results down to clients w/o issue using cursors.

As far a _single row_ taking up more memory than your server...you should fix your data in that case.  🤷 

[...]

PR #3406, opened 2025-03-19
Feat/add max result size

avallete:


brianc:
I appreciate the work here & the thought...I'm just not sure this is something we should support? It seems very much like a "don't do that query" kinda thing. It's loosely similar to "i wrote an infinite loop in my program. it should automatically abort instead." If your table has rows or columns that are gigabytes in size and you attempt to read them....then it's not gonna work well.  I would be surprised if this works well in _any_ database tool...though node is exceptionally limited with ram compared to some desktop tools.  What about truncating the column in the query so you only read the [first number of characters from the column](https://stackoverflow.com/questions/74417376/how-can-i-select-just-the-first-x-characters-from-a-text-column) or something else?  This feels to me like code to prevent a problem that is more of a use-case error than an actual bug or feature.

sehrope:
Haven't looked through the PR itself to see if it makes sense, but I can see a valid use for something like this when you're not actually sure what SQL the user is executing in advance (e.g., a building a database GUI or report generator that's using this driver to execute the commands). Having it fail and possibly close the connection is okay as long as it's handled gracefully (v.s. the alternative of an OOM killing the node process).

avallete:
Hey @brianc  !

Thank's for the comment. Indeed, as stated in the issue, the use case here is actually one where the actor "performing the query" and the "database" are two separate entity.

Basically, an api serving a web client, and proxying the "queries" in and out of the user database. The web-client allowing the user to run arbitrary queries and see the results for it in the browser. 

In such case, when a user query can OOM the "api server" accidentally forces the whole machine to die and reboot a new instance which is not ideal. Because then, the user doesn't get it's response but other connections that could be on the same instance from other users will also dies from the OOM.

I could understand if that's not something you're willing to pursue tough as it's indeed corner-casy. I opened the PR because  of this comment on an older issue mentioning similar problem: https://github.com/brianc/node-postgres/issues/2336#issuecomment-689188126

Let me know how you want to proceed.

Issue #2653, opened 2021-11-14
pg-protocol can throw an uncatchable error "Cannot create a string longer than 0x1fffffe8 characters" when reading large data


laurent22:
Using the "pg" package 8.5.1

To replicate the issue, save some large data as a blob into the database. In my case 340 MB was enough to trigger the bug. Then try to read back the data.

It will throw this uncatchable error that crashes the application:

```
Error: Cannot create a string longer than 0x1fffffe8 characters
    at Buffer.utf8Slice (<anonymous>)
    at Object.slice (node:buffer:593:37)
    at Buffer.toString (node:buffer:811:14)
    at BufferReader.string (/home/joplin/packages/server/node_modules/pg-protocol/src/buffer-reader.ts:35:32)
    at Parser.parseDataRowMessage (/home/joplin/packages/server/node_modules/pg-protocol/src/parser.ts:274:51)
    at Parser.handlePacket (/home/joplin/packages/server/node_modules/pg-protocol/src/parser.ts:172:21)
    at Parser.parse (/home/joplin/packages/server/node_modules/pg-protocol/src/parser.ts:101:30)
    at Socket.<anonymous> (/home/joplin/packages/server/node_modules/pg-protocol/src/index.ts:7:48)
    at Socket.emit (node:events:390:28)
    at addChunk (node:internal/streams/readable:315:12)
```

As a test I've tried to make it reject the promise when `parser.parse(buffer, callback)` throws an error in this function. At that time I can indeed catch the error, but rejecting the promise properly doesn't help for some reason and the error is still uncatchable:

https://github.com/brianc/node-postgres/blob/947ccee346f0d598e135548e1e4936a9a008fc6f/packages/pg-protocol/src/index.ts#L5

Any idea what might be the issue and how to fix it?

laurent22:
I think it may be related to this TODO here, as it's try to read as UTF-8 a binary blob:

https://github.com/brianc/node-postgres/blob/947ccee346f0d598e135548e1e4936a9a008fc6f/packages/pg-protocol/src/buffer-reader.ts#L6

And somewhat related is why does it try to read all fields as strings, even when it's a binary blob:

https://github.com/brianc/node-postgres/blob/947ccee346f0d598e135548e1e4936a9a008fc6f/packages/pg-protocol/src/parser.ts#L288

sehrope:
> And somewhat related is why does it try to read all fields as strings, even when it's a binary blob:

[...]

PR #3409, opened 2025-03-27
fix(parser): handle exceptions within handlePacket

avallete:
Hey there !

This is a proposal, in order to address #2653 it introduce another error type in addition to `DatabaseError` named `ParserError`, in charge of bubbling up any unexpected error that might occurs not on the Database itself, but between the database and the javascript parsing code (such as string too long for javascript to handle).

This is not the perfect solution since the parser will still throw an error. The benefit of this is that the `pg` package will no longer throw uncatchable exceptions when faced with such cases.
Instead the error will bubble up into the same `.on('error')` channel and can be caught by the caller program.

One of the parsing error (string too long) can be reproduced with this database:

```sql
CREATE TABLE public.large_data(data text);
INSERT INTO public.large_data(data) VALUES (repeat('A', 750 * 1024 * 1024)); -- create a string larger than what JS can handle
-- within pg:  client.query(`SELECT * FROM public.large_data`)
```

### DIFF
--- a/packages/pg-protocol/src/parser.ts
+++ b/packages/pg-protocol/src/parser.ts
@@ -25,6 +25,7 @@ import {
   MessageName,
   AuthenticationMD5Password,
   NoticeMessage,
+  ParserError,
 } from './messages'
 import { BufferReader } from './buffer-reader'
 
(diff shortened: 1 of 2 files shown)


brianc:
thanks for this PR! is it possilbe to write a test for this?

avallete:
@brianc I did my best but had to forge a specific invalid packet to force the parse error (since we can't reproduce the "max array error" in JS really).

I would like to make a more e2e test (ensuring within pg that the pooler/client now emit an error rather than an uncaught exception) but I can't find a way to do this without bumping the `pg-protocol` dependency. Do you have recommendation for how to proceed in that case ? I'm thinking I could do a commit by binding the "pg-protocol" lib to the local one so you can checkout and ensure it run. Keep the test skipped for now, deploy the `pg-protocol` next version and unskip the test.

packages/pg-protocol/src/parser.ts
```diff
@@ -152,53 +153,57 @@ export class Parser {
...
-        return this.parseParameterDescriptionMessage(offset, length, bytes)
-      case MessageCodes.CopyIn:
-        return this.parseCopyInMessage(offset, length, bytes)
-      case MessageCodes.CopyOut:
-        return this.parseCopyOutMessage(offset, length, bytes)
-      case MessageCodes.CopyData:
-        return this.parseCopyData(offset, length, bytes)
-      default:
-        return new DatabaseError('received invalid response: ' + code.toString(16), length, 'error')
+    try {
```
charmander, line 156: I think this `try` is too broad, and that every individual message handler should be able to handle arbitrary data gracefully, with `try` only being introduced at the most granular level possible when arbitrary exceptions can occur (usually because of user-provided functions).

It also erases important information about the original error since it only preserves its string representation instead of the complete object, but that’s more easily solved.

charmander (changes requested):

charmander:
Also, #2653’s case can be a query error, same as if a type parser failed. It doesn’t need to be treated like a protocol error.

packages/pg-protocol/src/parser.ts
avallete: Thanks for the comment!

Yeah, I totally agree with [[this comment](https://github.com/brianc/node-postgres/issues/2653#issuecomment-968317903)](https://github.com/brianc/node-postgres/issues/2653#issuecomment-968317903)
> “uncatchable errors are always no bueno” sums it up well.

So to me, it makes sense to have a catch-all at the top level, just in case something slips through in a handler. That way, worst case, it's still catchable and doesn't crash the app.

Then, we can still work toward handling better each possible exception within each of the packet handler. But this seems like a related but different concern that can be handled on it's own.

About this part:

> Also, [https://github.com/brianc/node-postgres/issues/2653’s](https://github.com/brianc/node-postgres/issues/2653%E2%80%99s) case can be a query error, same as if a type parser failed. It doesn’t need to be treated like a protocol error.

Not totally sure I follow — by "query error", do you mean something like a SQL error (e.g. missing table)? If so, that should already be returned as a `DatabaseError`, no? This won't throw an exception but return the error to the caller. But I guess you’re referring to something else, maybe you can clarify?

charmander: A generic protocol error is fatal for the connection, since it’s in an unknown state after the failure. A row field that can’t be converted into a JavaScript string because it’s too big doesn’t have to be this type of error. The ideal fix to #2653 would handle this condition by exposing it as a query error instead of a protocol error.

PR #3414, opened 2025-04-01
docs: fix bug in transaction example

lucaesposto:
Throwing error will not allow "finally" execution, so client.release() must be invoked before it.

Here a very simple example of JS behaviour:
```
try {
  console.log('init');
  throw new Error('TestError');
} catch (e) {
  console.log('catch'); / prints "Error: TestError"
  throw e
} finally {
  console.log('finally'); // not printed
}
```

I know this is an easy catch for experienced js developers, nevertheless i'm afraid providing this bugged example will cause (and caused) a lot of subtle bugs into beginners code.

### DIFF
--- a/docs/pages/features/transactions.mdx
+++ b/docs/pages/features/transactions.mdx
@@ -31,10 +31,10 @@ try {
   const insertPhotoValues = [res.rows[0].id, 's3.bucket.foo']
   await client.query(insertPhotoText, insertPhotoValues)
   await client.query('COMMIT')
+  client.release()
 } catch (e) {
   await client.query('ROLLBACK')
-  throw e
-} finally {
   client.release()
+  throw e
 }
 ```


brianc:
ah you're totally right - sorry about that!

brianc:
rerunning the failed job - sometimes tests timeout depending on infra issues...sucks but since some of the tests _actually test timeouts_ its a bit of a catch-22 to try & fix (unless it happens regularly)

brianc:
badda boom! ty!

charmander:
> Throwing error will not allow "finally" execution, so client.release() must be invoked before it.
> 
> Here a very simple example of JS behaviour:
> ```
> try {
>   console.log('init');
>   throw new Error('TestError');
> } catch (e) {
>   console.log('catch'); / prints "Error: TestError"
>   throw e
> } finally {
>   console.log('finally'); // not printed
> }
> ```

```
> try {
  console.log('init');
  throw new Error('TestError');
} catch (e) {
  console.log('catch');
  throw e
} finally {
  console.log('finally');
}

init
catch
finally
Uncaught Error: TestError
    at <anonymous>:3:9
```

PR #3415, opened 2025-04-01
Improved example in transactions.mdx

adriancuadrado:
This is the kind of thing the `finally` block was inveted for, to release resources and things like that.


charmander:
#3414

charmander:
Er, sorry, #3414 is wrong. But the original code also has a different bug.

charmander:
Fixed with a revert in a9fd34fb4217a7c16f02a4726efd61ad6a4f4796, thank you for pointing this out.

PR #3407, opened 2025-03-19
build: add esm exports

mesqueeb:
closes #3060

The implementation should work for enabling ESM compatibility without changing the core library code. Each package has:

- A properly structured ESM wrapper that:
  - Imports the CommonJS module
  - Re-exports both named exports and the default export
- An updated `package.json` with an appropriate `exports` field providing both `import` and `require` paths
- The `esm` directory added to the `files` array so it's included when published
- Proper handling of TypeScript type definitions where applicable

No further changes are needed. The solution is non-invasive and should work for all the packages in the repository.

PR #3423, opened 2025-04-21
Add esm exports

brianc:
This builds off of #3407 and hopefully helps address the issue [here](https://github.com/cloudflare/workers-sdk/issues/8840).


brianc:
Still adding more tests, but wanted to open the PR to see how it runs in CI.

brianc:
todo:

- [ ] test [if this gets fixed](https://github.com/awkweb/pg-cloudflare-issue)
- [ ] update documentation to ESM imports & put note about also supporting commonJS

B4nan:
This feels breaking, since now you are disallowing the imports for `pg/lib/type-overrides`. Or is there some other way to do this import?

```ts
import TypeOverrides from 'pg/lib/type-overrides';
```

mesqueeb:
@B4nan importing from undocumented paths is debatable wether it's breaking or not.

To my understanding, in an ESM environment this (your snippet) will get tricky since the source code is mostly CJS, and ESM only has single wrapper index files.

B4nan:
Fair enough, but how can we do this properly if this was the wrong way? i believe that class is supposed to be used in downstream.

mesqueeb:
@B4nan I guess the best way is to make sure that class is also exported as part of the esm-facing index file.
So that you can do:
`import { TypeOverrides } from 'pg'`
That should be a simple addition is

B4nan:
Btw I learned about this trick ages ago from this issue, maybe there is another way to override type mapping per connection? 

https://github.com/brianc/node-postgres/issues/1838

There is an old PR that was exposing the class in root exports but it never got merged. 

https://github.com/brianc/node-postgres/pull/2369

PR #3425, opened 2025-04-22
fix(pg-connection-string): export default from esm wrapper

### LINKED ISSUE #3424: pg-connection-string v2.8.0 broke default export
> @hjr3 just a heads up, 2.8.0 was technically a breaking change. Even though `module.exports = parse` still existed, it broke the default export:
> 
> ```
> import pgConnectionString from 'pg-connection-string';
>        ^^^^^^^^^^^^^^^^^^
> SyntaxError: The requested module 'pg-connection-string' does not provide an export named 'default'
>     at ModuleJob._instantiate (node:internal/modules/esm/module_job:180:21)
>     at async ModuleJob.run (node:internal/modules/esm/module_job:263:5)
>     at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:578:26)
>     at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:116:5)
> 
> Node.js v22.14.0
> ```
> 
> I'm guessing it's an ESM / bundler or whatever type error, but it's easy enough to fix and switch to `import { parse } from 'pg-connection-string';` 

 _Originally posted by @ThisIsMissEm in [#3421](https://github.com/brianc/node-postgres/issues/3421#issuecomment-2822279994)_

hjr3:
Prior to v2.8.0, the parse function was the default when using import. When esm compatibility was introduced in v2.8.0, there was not default specified. This broke existing code that relied on that default.

Fixes #3424

### DIFF
--- a/packages/pg-connection-string/esm/index.mjs
+++ b/packages/pg-connection-string/esm/index.mjs
@@ -2,6 +2,7 @@
 import connectionString from '../index.js'
 
 // Re-export the parse function
+export default connectionString.parse
 export const parse = connectionString.parse
 export const toClientConfig = connectionString.toClientConfig
 export const parseIntoClientConfig = connectionString.parseIntoClientConfig


brianc:
oh my bad! Sorry about that! The whole esm/commonjs thing is quite an easy thing to trip over. I'll publish this right away.

PR #3433, opened 2025-04-23
Expose TypeOverrides in esm & cjs from root of package

brianc:
This [replaces this old PR here](https://github.com/brianc/node-postgres/pull/2369). Added some more tests to the exports from pg, too.

### DIFF
--- a/packages/pg-esm-test/pg.test.js
+++ b/packages/pg-esm-test/pg.test.js
@@ -1,6 +1,17 @@
 import assert from 'node:assert'
 import { describe, it } from 'node:test'
-import pg, { Client, Pool } from 'pg'
+import pg, {
+  Client,
+  Pool,
+  Connection,
+  defaults,
+  types,
+  DatabaseError,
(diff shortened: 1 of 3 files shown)


B4nan:
Thanks for this, but please note that the types package is still outdated.

brianc:
Yeah the types package unfortunately has some breaking changes i am _never_
going to incorporate into pg due to the amount of disruption it would
cause. I am still figuring out how I want to get onto a more recent version
but without the breaking changes. It will happen though, its high on my
priority list.

brianc:
> just to be clear, i am talking about the `@types/pg` package. would be great to move the type definitions here, its a shame such a popular and maintained package is not maintaining the types together with the code.

ohhhhhhhh i see. Yeah I would like to move the type definitions to this library. I also want to convert the rest of this library _to typescript_. I was gonna do typedefs first, so then its easier to align the typescript to them. 😄 definitely on my todo list!

Issue #3009, opened 2023-06-13
Support Min connection pool parameter


Smriti-OSS:
Hey @brianc,

**Problem Statement:** I have an application which is currently using the default configuration of node-postgres pool to connect to postgres. I have noticed that whenever we query the application - node-postgres connection pool always creates a new connection and so the total time taken for query execution is always between 3-5sec. If we re-use an idle connection for query execution - it just takes 2 millisecond to 300 millisecond.

**Proposed Solution:**I am thinking of keeping some idle connections in the pool - so that they could be used whenever the database is queried.  Something like this -
```
let pool = new Pool({
        database: connectionConfig.database,
        port: connectionConfig.port,
        user: connectionConfig.user,
        password: connectionConfig.password,
        host: connectionConfig.host,
        min: 1,
        max: 10
    });
```

I am open to submitting a PR for the above feature but thought of first discussing it with the community whether this would be a feature worth exploring.

khoirulamri:
I experienced the same problem.  each initial set of connections takes more than 200 ms, so the api response time becomes > 1s

capJavert:
@brianc we are working with `pg-pool` and would really benefit from this option. My question is, is there any issue on driver level that would block implementing something like that?

I took a brief look at the code and it seems like it should be possible as an addition to pool logic. Would offer to contribute as well 🙌

htrajan:
I took a crack at addressing this. Would love to get your thoughts @brianc @charmander.

PR #3438, opened 2025-04-23
Support Min connection pool parameter #3009

htrajan:
My attempt to address #3009 

Hopefully this is sensible with the tests I've included

I suspect that lack of this functionality is causing my company's app to frequently hit bottlenecks with allocating new db pool connections

### DIFF
--- a/packages/pg-pool/index.js
+++ b/packages/pg-pool/index.js
@@ -367,17 +367,15 @@ class Pool extends EventEmitter {
 
     // idle timeout
     let tid
-    if (this.options.idleTimeoutMillis) {
-      if (this._isAboveMin()) {
-        tid = setTimeout(() => {
-          this.log('remove idle client')
-          this._remove(client)
-        }, this.options.idleTimeoutMillis)
+    if (this.options.idleTimeoutMillis && this._isAboveMin()) {
+      tid = setTimeout(() => {
+        this.log('remove idle client')
+        this._remove(client)
+      }, this.options.idleTimeoutMillis)
 
-        if (this.options.allowExitOnIdle) {
-          // allow Node to exit if this is all that's left
-          tid.unref()
-        }
+      if (this.options.allowExitOnIdle) {
+        // allow Node to exit if this is all that's left
+        tid.unref()
       }
     }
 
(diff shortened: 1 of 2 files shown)


chasdevs (approved):
LGTM; I'm not a maintainer of this library, though 😅

brianc:
nice tests. this looks good to me, tbh! I prefer my pools to drain when my app is idle...feels better to me than leaving a bunch of sockets open doing nothing for > 5 minutes, but I can definitely see in some "needs to be fast, but gets spikey traffic" this could come in handy. That initial connection to the backend is no joke. I've seen it take upwards of a second on GCS in the past.

htrajan:
> nice tests. this looks good to me, tbh! I prefer my pools to drain when my app is idle...feels better to me than leaving a bunch of sockets open doing nothing for > 5 minutes, but I can definitely see in some "needs to be fast, but gets spikey traffic" this could come in handy. That initial connection to the backend is no joke. I've seen it take upwards of a second on GCS in the past.

Great! What would be the next steps to get this into the next release?

charmander:
Thoughts on implementing or documenting:

- prefilling the pool to the minimum
- maintaining the minimum on connection errors

?

Keepalives are also sort of related to this, but I think any documentation and work needed there stands on its own.

brianc:
> Thoughts on implementing or documenting:
> 
> * prefilling the pool to the minimum
> * maintaining the minimum on connection errors
> 
> ?

I can document that this is _not_ implemented at the moment, and merge this as a first step. Those other things yeah could come in the future.

brianc:
I will release a new `minor` version of `pg-pool` on Monday morning. i don't like doing afternoon/evening releases because I always like to be around and online throughout the day to respond to any unintentional breaking changes or bugs

htrajan:
Hi @brianc, I'm not seeing the update in the latest 8.15.6 release. Do you have an ETA on this?

zyf0330:
When create connections with number more than `min` simultaneously, they all will be `remove`d after `idleTimeoutMillis`.

brianc:
@htrajan @zyf0330 I'm seeing it [right here](https://github.com/brianc/node-postgres/blob/pg-pool%403.10.0/packages/pg-pool/index.js#L90) am I missing something?

htrajan:
@brianc nope, it's all good now. @zyf0330 might have spotted an edge case where my implementation doesn't work. Unfortunately, I'm pretty tied up now and won't have time to look into it in the near future. I was able to deploy the change to my company's application and get the desired outcome of no more `pg-pool.connect` spikes.

PR #3440, opened 2025-04-24
Make pg-cursor compatible with older versions of pg

### LINKED ISSUE #3430: TypeError: Cannot read properties of undefined (reading 'prepareValue')
Hi, 
Latest PR: https://github.com/brianc/node-postgres/commit/940479bc4b601a89a6fc915be959ca96e83dd4d0 broke our runtime completely.
This occurred even though we did not upgrade the version of pg so it was a complete surprise.

Issue is coming from: https://github.com/brianc/node-postgres/blob/940479bc4b601a89a6fc915be959ca96e83dd4d0/packages/pg-cursor/index.js#L4

```
TypeError: Cannot read properties of undefined (reading 'prepareValue')
    at Object.<anonymous> (/<path>/node_modules/pg-cursor/index.js:4:23)
    at Module._compile (node:internal/modules/cjs/loader:1565:14)
    at node:internal/modules/cjs/loader:1708:10
    at Object.require.extensions.<computed> [as .js] (/<path>/node_modules/ts-node/src/index.ts:1608:43)
    at Module.load (node:internal/modules/cjs/loader:1318:32)
    at Function._load (node:internal/modules/cjs/loader:1128:12)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:219:24)
    at Module.require (node:internal/modules/cjs/loader:1340:12)
    at require (node:internal/modules/helpers:138:16)
```

Please look into it as soon as possible.
pg version: 8.14.1
node version: 22.12.0
npm version: 10.9.0

Let me know if you need any other details.

brianc:
closes #3430

### DIFF
--- a/packages/pg-cursor/index.js
+++ b/packages/pg-cursor/index.js
@@ -1,7 +1,7 @@
 'use strict'
-const pg = require('pg')
-const { Result, utils } = pg
-const prepare = utils.prepareValue
+// note: can remove these deep requires when we bump min version of pg to 9.x
+const Result = require('pg/lib/result.js')
+const prepare = require('pg/lib/utils.js').prepareValue
 const EventEmitter = require('events').EventEmitter
 const util = require('util')
 


charmander:
Hmm, I haven’t been completely following the situation here, but isn’t the bug that the version of this code before this PR doesn’t work? It should, right?

charmander:
Oh, I didn’t realize those weren’t exported, got it. Never mind!

brianc:
ah yeah it was related to me changing things because i didn't think the old
export style worked, but then fixed the old export style and never went
back and changed things back because I kinda took over the PR from someone
else. Once I wrote tests with that separate esm & cjs module import thing
it made things much clearer, thankfully.

PR #3461, opened 2025-05-10
pool.end() resolves before the last pool.query()

hjr3:
dupe of https://github.com/brianc/node-postgres/pull/3254 with 289273c to fix the failing test

### DIFF
--- a/packages/pg-pool/index.js
+++ b/packages/pg-pool/index.js
@@ -373,7 +373,7 @@ class Pool extends EventEmitter {
     if (this.options.idleTimeoutMillis && this._isAboveMin()) {
       tid = setTimeout(() => {
         this.log('remove idle client')
-        this._remove(client, this._pulseQueue.bind(this))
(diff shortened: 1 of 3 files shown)


hjr3:
note: if we want to merge this, then i should update the docs for https://node-postgres.com/apis/pool#remove to document the change in behavior

brianc:
> note: if we want to merge this, then i should update the docs for https://node-postgres.com/apis/pool#remove to document the change in behavior

Would you consider this a bug-fix or a breaking change? Seems to me like its a semantic alignment w/ expectations and more of a fix than a break, and no exising tests required changing so its more of a patch than a semver major.  But I'd love to hear your thoughts!

hjr3:
I agree this is a fix. No one expects `pool.end` to resolve before the last `pool.query`. Also, the documentation did not state that `remove` event is emitted prior to client end finishing.

PR #3472, opened 2025-05-26
refactor: use for...of for better readability

noritaka1166:
use for...of for better readability


brianc:
I appreciate the thought here but I'm going to stick with `for i` loops in the code where they exist. Last I checked they still have a performance improvement/advantage over a `for of` loop.  While none of these code paths are seriously white-hot execution frequency I still don't mind using `for i` loops and so I'm just going to pass on this one. Thank you though!

PR #3473, opened 2025-05-26
feat(pg-connection-string): warn if non-standard ssl options are used

hjr3:
In preparation for v3.0.0, we start warning users to be explicit about the sslmode they want.

### DIFF
--- a/packages/pg-connection-string/index.js
+++ b/packages/pg-connection-string/index.js
@@ -210,7 +210,7 @@ function deprecatedSslModeWarning(sslmode) {
   if (!deprecatedSslModeWarning.warned) {
     deprecatedSslModeWarning.warned = true
     emitWarning(`SECURITY WARNING: The SSL modes 'prefer', 'require', and 'verify-ca' are treated as aliases for 'verify-full'.
-In the next major version (v3.0.0), these modes will adopt standard libpq semantics, which have weaker security guarantees.
+In the next major version (pg-connection-string v3.0.0 and pg v9.0.0), these modes will adopt standard libpq semantics, which have weaker security guarantees.
 
 To prepare for this change:
 - If you want the current behavior, explicitly use 'sslmode=verify-full'
--- a/packages/pg-connection-string/index.js
+++ b/packages/pg-connection-string/index.js
@@ -1,5 +1,7 @@
 'use strict'
 
+const { emitWarning } = require('node:process')
+
 //Parse method copied from https://github.com/brianc/node-postgres
 //Copyright (c) 2010-2014 Brian Carlson (brian.m.carlson@gmail.com)
 //MIT License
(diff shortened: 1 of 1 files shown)


charmander (approved):
Could warn this way for repeat callers: https://nodejs.org/api/process.html#avoiding-duplicate-warnings

hjr3:
> Could warn this way for repeat callers: https://nodejs.org/api/process.html#avoiding-duplicate-warnings

TIL - implemented

packages/pg-connection-string/index.js
```diff
@@ -201,6 +206,20 @@ function parseIntoClientConfig(str) {
   return toClientConfig(parse(str))
 }
 
+function deprecatedSslModeWarning(sslmode) {
+  if (!deprecatedSslModeWarning.warned) {
+    deprecatedSslModeWarning.warned = true
+    emitWarning(`SECURITY WARNING: The SSL modes 'prefer', 'require', and 'verify-ca' are treated as aliases for 'verify-full'.
+In the next major version (v3.0.0), these modes will adopt standard libpq semantics, which have weaker security guarantees.
+
```
brianc, line 214: nit: I think this would probably be considered a breaking change to `pg` itself so might want to call out we'd be bumping pg semver to 9.x at this point as well?

hjr3: called out both packages

brianc: awesome! ty! will merge now. 😄

jawj:
Has the change that drives this warning been discussed elsewhere?

Usually I would be all for adopting the standard or upstream semantics — but I actually think this is an example where libpq gets it badly wrong and current node-postgres has historically done much better. libpq's`sslmode=require` really _sounds_ secure but [it absolutely is not](https://neon.com/blog/postgres-needs-better-connection-security-defaults): this is a serious footgun that people are shooting themselves in the foot with all the time.

The change here means that, if and when developers upgrade node-postgres to 9.0, connections that used to be secure will suddenly become insecure. I mean, _hopefully_ they'll see the warning and do something about it ... but surely many will not.

IMHO it would be much better to have users actively opt in to this downgrade in TLS protection. That would mean keeping the `uselibpqcompat` flag, and keeping it `false` by default.

As I understand it, node-postgres historically treats every `sslmode` from `require` upwards as not just `verify-full`, but as `sslrootcert=system`. IMHO this is also behaviour that libpq got wrong and node-postgres got right. If the developer asks for `verify-full` and doesn't provide a root CA cert, libpq throws. In the same situation, node-postgres validates against the system root CA certs. That matches the behaviour of https and is the logic I think most current developers would expect.

Is it proposed to match libpq here too, and throw an error if the user specifies `sslmode=verify-full` without also providing a certificate? I don't think it should be — but that's also where the logic of matching libpq takes you.

hjr3:
@jawj there was a long discussion on https://github.com/brianc/node-postgres/pull/2709

PR #3482, opened 2025-06-03
fix(pg-cloudflare): use conditional export to support bundlers that d…

### LINKED ISSUE #3452: [v8.15.0] Regression: UnhandledSchemeError "cloudflare:sockets" when bundling with Webpack
After upgrading to 8.15.0, bundling with Webpack results in the following error:

```
Webpack compilation failed:
in cloudflare:sockets
  Module build failed: UnhandledSchemeError: Reading from "cloudflare:sockets" is not handled by plugins (Unhandled scheme).
  Webpack supports "data:" and "file:" URIs by default.
  You may need an additional plugin to handle "cloudflare:" URIs.
```

This previously occurred in version 8.11.0: https://github.com/brianc/node-postgres/issues/2975

And was fixed in this PR: https://github.com/brianc/node-postgres/pull/2978

CHC383:
close #3452

Conditional export was chosen as the fix to align with the previous solution #2978, however this does require the consumers to explicitly resolve the condition to `cloudflare` in their bundler's config, which might be considered as a breaking change.

A private package `webpack-test` is also created to test the bundling case. It will be run through `yarn test` and calls webpack to bundle the code that imports `pg-cloudflare`, and it does reproduce the error if the fix is not in place.

I didn't update `docs` as I searched for `cloudflare` on [node-postgres.com](https://node-postgres.com/) and didn't find anything, so I am not sure where is the right place. Hopefully the README of pg-cloudflare is good enough, and feel free to add more docs if necessary.


brianc:
This is absolutely, super helpful. Thank you __so much__ 🎉

Issue #3298, opened 2024-08-22
errors in docs and poorly chosen default for PoolConfig.maxLifetimeSeconds


mutsys:
the docs state:

```
  // number of milliseconds a client must sit idle in the pool and not be checked out
  // before it is disconnected from the backend and discarded
  // default is 10000 (10 seconds) - set to 0 to disable auto-disconnection of idle clients
  idleTimeoutMillis?: number
```

however, the actual default value is 0:

line 92 in `pg-pool/index.js`
```
    this.options.maxLifetimeSeconds = this.options.maxLifetimeSeconds || 0
```

very frustrating. the default configuration for `pg-pool` results in no pooling whatsoever. acquiring a connection always returns a new, freshly created connection that is immediately discarded once it is released. 0 is not a sensible default. 10 seconds doesn't seem like a sensible default either.

PR #3484, opened 2025-06-10
docs(pg.Pool): describe `maxLifetimeSeconds` config option

birjj:
The official [`pg.Pool`](https://node-postgres.com/apis/pool) page doesn't describe the `maxLifetimeSeconds` option, which is nonetheless [supported by the code](https://github.com/brianc/node-postgres/blob/03642abec1c09337fa9cfb4b7ce5543b086c6437/packages/pg-pool/index.js#L93), and present in the `@types/pg` package (it was added in https://github.com/brianc/node-postgres/pull/2698 all the way back in 2022). The lack of documentation makes it slightly difficult to discover the option.

The wording I've used was taken from https://github.com/brianc/node-postgres/issues/3298#issuecomment-2305207256. Feel free to change it if you're not happy with it.

### DIFF
--- a/docs/pages/apis/pool.mdx
+++ b/docs/pages/apis/pool.mdx
@@ -50,6 +50,12 @@ type Config = {
   // to the postgres server.  This can be handy in scripts & tests
   // where you don't want to wait for your clients to go idle before your process exits.
   allowExitOnIdle?: boolean
+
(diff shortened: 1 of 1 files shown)


brianc:
thanks for updating the docs! always super useful!

PR #3489, opened 2025-06-14
Do not concatenate an array if passed to escapeLiteral

brianc:


packages/pg/lib/utils.js
```diff
@@ -176,6 +176,14 @@ const escapeLiteral = function (str) {
   let hasBackslash = false
   let escaped = "'"
 
+  if (str == null) {
+    return "''"
+  }
+
+  if (typeof str !== 'string') {
+    return "''"
```
charmander, line 184: How about throwing a `TypeError`? Seems about the same level of breaking change, but potentially less confusing.

brianc: that's what i was going to do at first, but after writing a bunch of tests for previously indeterminate behavior, almost everything returned `"''"` when passed to this function - numbers, boolean, date, object. The only problem was array. The only problem is with things with a `.length` property which aren't strings.   It actually kinda feels more "escapey" to me to just turn your "not a string" into an empty string in the query if you're concatenating in there. But I was definitely only the fence.

charmander: Now that you mention those examples… how about casting everything to `String`? pg 9?

brianc: yeah that isn't a good idea - def breakage, though subtle. tbh whenever I, very rarely, use these functions to concatenate some form of user-input into a query directly I do a _ton_ of external sanitization first. Check types, if its a number make sure its in an expected range, etc. It's scary & should be a last resort in most cases.

PR #3490, opened 2025-06-16
Bump engines since we do not test lower than 16 anymore

brianc:


brianc:
cc @bitifet 😄  thanks for your nudge on this

bitifet:
> cc @bitifet 😄  thanks for your nudge on this

Don't mention it!

Thanks to you for your invaluable work leading node-postgres!!

PR #3491, opened 2025-06-17
Implement retry on timeout when creating new client

MathieuGuillet:
Add new options and instructions to implement retries on timeout when the pool tries to create new client in the pool.

This will avoid to throw an error on the first timeout error if it can be recover in a given time.

### DIFF
--- a/packages/pg-pool/index.js
+++ b/packages/pg-pool/index.js
@@ -91,6 +91,9 @@ class Pool extends EventEmitter {
     this.options.maxUses = this.options.maxUses || Infinity
     this.options.allowExitOnIdle = this.options.allowExitOnIdle || false
     this.options.maxLifetimeSeconds = this.options.maxLifetimeSeconds || 0
+    this.options.retryOnTimeout = this.options.retryOnTimeout || false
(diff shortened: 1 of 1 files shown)


charmander:
I think library users should be able to achieve this by calling the relevant pool function repeatedly as long as it returns a specific type of error (which might be different types of errors according to different needs). You can write a module for that.

MathieuGuillet:
> I think library users should be able to achieve this by calling the relevant pool function repeatedly as long as it returns a specific type of error (which might be different types of errors according to different needs). You can write a module for that.

I understand your suggestion, but we are using an ORM (TypeORM) and it does not allow to trick that point easily (we have to change the driver to achieve what you suggest).
I believe it would be reasonable for the connection pool to handle retrying failed connections, within a configurable limit, for a predefined set of error types.
I agree that limiting the retries to only timeout errors is too specific/restrictive. I can improve that by exposing a function in the configuration options to check if an error should trigger the retries (like in TypeORM config: `toRetry: (err) => boolean`)

What do you think @charmander ? Can I re-open this PR and improve the behavior as explained?

Issue #3487, opened 2025-06-13
Bind does not generate the result column format encoding in pg-protocol


fqazi:
While running some tests for CockroachDB we noticed that the binary encoding flag passed during the bind operation does not follow the Postgres specifications for the result-column format. This problem reproduces easily on Postgres as well, since the binary flag is not being encoded correctly at:
https://github.com/brianc/node-postgres/blob/6b016b37d434cbe88b139743b1bc9a1861382750/packages/pg-protocol/src/serializer.ts#L161

This violates the specification (https://www.postgresql.org/docs/current/protocol-message-formats.html#PROTOCOL-MESSAGE-FORMATS-BIND), which says for the result column formats:
```
Int16
The number of result-column format codes that follow (denoted R below). This can be zero to indicate that there are no result columns or that the result columns should all use the default format (text); or one, in which case the specified format code is applied to all result columns (if any); or it can equal the actual number of result columns of the query.

Int16[R]
The result-column format codes. Each must presently be zero (text) or one (binary).
```

The pg-protocol code incorrectly only puts a single Int16 for the binary encoding case, which leads to a protocol violation error on Postgres.

```
const pg = require('pg');
const fs = require('fs');

const config = {
  user: 'blah2',
  database: 'postgres',
  host: process.env.PGHOST || 'localhost',
  port: process.env.PGPORT || '5432',
};
const assert             = require('assert');

const client = new pg.Client(config);
client.connect()

describe('regression tests', () => {
  it('allows you to switch between format modes for arrays', () => {
// pg-node has a regression where the binary encoding
    // is sending incorrect data. See:
   return  client.query({
            text: 'SELECT CAST($1 AS INT[]) as b',
            values: [[1, 2, 8]],
            binary: false,
          }).then(r => {
            return client.query({
              text: 'SELECT CAST($1 AS INT[]) as a',
              values: [[4, 5, 6]],
              binary: true,
            });
          }).then(results => {
            assert.deepEqual([4, 5, 6], results.rows[0].a);
          });
  })
```

Running this test we get the following output when the binary encoding is sued:
```
  1) regression tests
       allows you to switch between format modes for arrays:
     error: insufficient data left in message
      at node_modules/pg/lib/client.js:526:17
      at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
```

hjr3:
@fqazi we landed a fix for this. we will release soon :tm:. please re-open this issue if you are still seeing problems.

[...]

PR #3494, opened 2025-06-19
Fixes for binary protocol array handling

hjr3:


brianc:
ohh nice is this a fix #3487?

hjr3:
The first commit fixes the reported error in #3487. I annotated that in the first commit, but it seems like Github did not pick it up.

While fixing the reported issue, I discovered that the binary array handling was not working. So, the second commit fixes that. I can split this into two commits if that is easier as well.

brianc:
> So, the second commit fixes that. I can split this into two commits if that is easier as well.

one is totally fine!

brianc:
@hjr3 you're the best! 🫶

Issue #3493, opened 2025-06-18
CloudflareSocket is not a constructor in pg-cloudflare@1.2.6


astanciu:
I'm using `pg` in a cloudflare worker and this worker is using vite, using the cloudflare vite plugin.

The new `pg-cloudflare` package on version 1.2.6 is causing this error:

```
Unhandled error: TypeError: CloudflareSocket is not a constructor
    at getStream2 (/Users/user/a/apps/a.com/node_modules/.vite/deps_a_test/pg.js:3090:16)
    at new Connection2 (/Users/user/a/apps/a.com/node_modules/.vite/deps_a_test/pg.js:3136:40)
    at new Client2 (/Users/user/a/apps/a.com/node_modules/.vite/deps_a_test/pg.js:3669:43)
    at /Users/user/a/packages/worker/src/lib/service.ts:9:18
    at dispatch (/Users/user/a/apps/a.com/node_modules/.vite/deps_a_test/chunk-7KAI6PC7.js:36:23)
    at /Users/user/a/apps/a.com/node_modules/.vite/deps_a_test/chunk-7KAI6PC7.js:36:46
    at AsyncContext.runWithContext (/Users/user/a/packages/utils/logger/context.ts:19:35)
    at /Users/user/a/packages/worker/src/lib/requestPrep.ts:29:24
    at dispatch (/Users/user/a/apps/a.com/node_modules/.vite/deps_a_test/chunk-7KAI6PC7.js:36:23)
    at /Users/user/a/apps/a.com/node_modules/.vite/deps_a_test/chunk-7KAI6PC7.js:19:12
```

I dug around trying to figure it out, but was unable to. I saw in the pg-cloudflare readme references to things related to this.. I updated my worker's vite.config file with 

```
  build: {
    rollupOptions: {
      external: ["cloudflare:sockets"],
    },
  },
resolve: {
    conditions: ["cloudflare"],
}
```

but this didn't help. Reverting back to 1.2.5 fixed me. 

Am I missing some new required config or is there a bug somewhere?

TrySound:
This is because vite-plugin-cloudflare hardcodes export conditions and is using "workerd" instead of "cloudflare"
https://github.com/cloudflare/workers-sdk/blob/main/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts#L128

TrySound:
Reported https://github.com/cloudflare/workers-sdk/issues/9668

korywka:
[...]

PR #3498, opened 2025-06-26
Replace `cloudflare` export condition in `pg-cloudflare` with `workerd`

jamesopstad:
Fixes https://github.com/brianc/node-postgres/issues/3493
Fixes https://github.com/cloudflare/workers-sdk/issues/9668

This changes the export conditions for `pg-cloudflare` to use `workerd` instead of `cloudflare`. `workerd` is the standard export condition for Cloudflare Workers, as specified by WinterCG - https://runtime-keys.proposal.wintercg.org/#workerd. This will ensure compatibility with Wrangler and the Cloudflare Vite plugin.

### DIFF
--- a/packages/pg-cloudflare/package.json
+++ b/packages/pg-cloudflare/package.json
@@ -11,7 +11,7 @@
   },
   "exports": {
     ".": {
-      "cloudflare": {
+      "workerd": {
         "import": "./esm/index.mjs",
         "require": "./dist/index.js"
       },
--- a/packages/pg-bundler-test/esbuild-cloudflare.config.mjs
+++ b/packages/pg-bundler-test/esbuild-cloudflare.config.mjs
@@ -4,5 +4,5 @@ await esbuild.build({
   entryPoints: ['./src/index.mjs'],
   bundle: true,
   outfile: './dist/esbuild-cloudflare.js',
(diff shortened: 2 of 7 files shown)


packages/pg-cloudflare/package.json
```diff
@@ -11,7 +11,7 @@
   },
   "exports": {
     ".": {
-      "cloudflare": {
+      "workerd": {
         "import": "./esm/index.mjs",
         "require": "./dist/index.js"
       },
```
TrySound, line 17: I would keep both for backward compatibility

jamesopstad: Given that the `cloudflare` condition was only added very recently and seems broken for most people using it, I would be in favour of removing it. We don't want to encourage others to use this as an export condition for Cloudflare Workers.

thomasgauvin: From what we can tell, this was merged in https://github.com/brianc/node-postgres/pull/3482 by @brianc, but in so doing broke the majority of Cloudflare projects given that `workerd` is the export condition expected

brianc:
aye! sorry for breaking this - trying to migrate to esm while also maintaining commonjs compat has been an absolute nightmare, surprisingly! I'll get this pushed up ASAP! ❤️

PR #3499, opened 2025-06-26
fix: typos

nigrosimone:
just some docs typos


brianc:
nice! Always appreciated!

PR #3509, opened 2025-07-10
fix #3508 - recheck min client count during idle callback

### LINKED ISSUE #3508: pg-pool min parameter is not obeyed with bursty traffic
The `min` parameter for pg.Pool was implemented [in this PR](https://github.com/brianc/node-postgres/pull/3438).

The implementation has a problem that will cause the number of clients to drop below the min if multiple clients release during the same idle window. The `isAboveMin()` check occurs during the release of a client, but the actual removal occurs in a`setTimeout` callback.

Example 
```
 const pool = new Pool({ max: 10, min: 5})
```

If you start with an empty pool, and you fire 10 concurrent requests to the database, you will fill up the pool to 10 clients.
Now, if you make no other requests, all of those 10 clients will release at roughly the same time.  Each of those will evaluate `isAboveMin()` to true, so each will create the callback to remove the client after the idleTimeout.  This will drop the client count to 0.

The `isAboveMin()` check should also be made in the callback, so you can get atomic removals and prevent all of the clients from being removed if they release in the same idle window and all think they can be removed and stay above the `min` settings.

As it stands, if more than (max-min) clients release in a time window less than the idle timeout, they will cause the client count to drop below min.   This is mostly a problem for a bursty traffic profile.

barryhagan:


brianc:
wow! amazing turn around on this! thank you! ❤️

brianc:
unrelated side note: I should port the test to use async/await now that it exists instead of using `co`. I used `co` because _way way back_ async/await wasn't a thing, but writing callbacks was always icky.

sehrope:
>  I used co because way way back async/await wasn't a thing, but writing callbacks was always icky.

Async / await?! Ha, we didn't even have native Promises!

PR #3510, opened 2025-07-16
Deprecations

brianc:
- deprecate `client.activeQuery` - the concept needs to be overhauled a bit to support pipeline mode in which multiple queries can be "in flight" at once
- deprecate `client.queryQueue` - this should never been manipulated from outside the client, and again will be changing with a pipeline mode implementation. The entire concept of "queueing" queries is very old, and honestly was a bad idea when I implemented it 15 years ago or however long ago it was.  Lets just all pretend this doesn't exist and make it "private" so its more of an implementation detail. 
- deprecate built-in `pgpass` support - I'd rather this be done in userspace since the connection parameters can now take any old async function as the value to password, if someone's using `pgpass` they can install it & write a 1 liner in their connection config to use their file.  I'm not 100% sold on this deprecation & removal being the best...but I also don't love having 3rd party dependencies for stuff like this.  Interested in feedback here.


packages/pg/lib/client.js
```diff
@@ -540,7 +583,7 @@ class Client extends EventEmitter {
         result = new this._Promise((resolve, reject) => {
           query.callback = (err, res) => (err ? reject(err) : resolve(res))
         }).catch((err) => {
-          // replace the stack trace that leads to `TCP.onStreamRead` with one that leads back to the
+          // replace the stack trace that leads to \`TCP.onStreamRead\` with one that leads back to the
```
charmander, line 586:
```suggestion
          // replace the stack trace that leads to `TCP.onStreamRead` with one that leads back to the
```

packages/pg/lib/client.js
```diff
@@ -70,20 +96,35 @@ class Client extends EventEmitter {
...
+    activeQueryDeprecationNotice()
+    return this._activeQuery
+  }
+
+  set activeQuery(val) {
+    activeQueryDeprecationNotice()
+    this._activeQuery = val
+  }
+
+  _getActiveQuery() {
```
charmander, line 109: Why not `._activeQuery` directly, like `._queryQueue`?

charmander (commented):
The deprecations seem reasonable!

packages/pg/lib/client.js
brianc: uhh so I have been experimenting a _lot_ with AI assisted coding and I asked all the major foundational models to port this to support pipeline mode and they all did a pretty crap job in various ways. So, I thought hey lemme do a more incremental approach. But _then_ I thought, since I have to deprecate `this.activeQuery` I might as well do a few more.  To answer your question, the query queue isn't used w/ pipelining and the active query is actually going to be a lookup into an array of inflight queries. Hence the function. That being said, if it ends up eventually not needing to be a func, I'll redo it back to an instance variable.

brianc:
Thank you, as always, for your tireless & thoughtless contributions @charmander. If you ever make it to Ausitn TX, lemme know I'd love to take you out to dinner or getcha a gift card to H.E.B. :)

brianc:
(i pressed wrong button)

PR #3514, opened 2025-07-17
Fix #3513 - throw the right error

### LINKED ISSUE #3513: Invalid URL does not throw correct error
This commit introduced redaction of URL details in this file packages/pg-connection-string/index.js:
https://github.com/brianc/node-postgres/commit/7ab5923fad85994d74b24e7c12bb79c9441da86b#diff-08a5e82487ebd9b43751630019753901fae0a111f8d009ad2e9d194445e96922

This is good, but the problem is that the error is not thrown anymore. It's just caught and modified, but the modified error is not thrown again. The actual error that is thrown is `TypeError: Cannot read properties of undefined (reading 'searchParams')`, which is not helpful at all.

mathiasbockwoldt:
Due to a recent change, the thrown error for invalid URLs was not the excepted error, but another error, which made the problem hard to understand.
This PR re-throws the original error (with redacted password).
The existing test was updated to make sure the right error is thrown.

### DIFF
--- a/packages/pg-connection-string/index.js
+++ b/packages/pg-connection-string/index.js
@@ -35,6 +35,7 @@ function parse(str, options = {}) {
   } catch (err) {
     // Remove the input from the error message to avoid leaking sensitive information
     err.input && (err.input = '*****REDACTED*****')
+    throw err
   }
 
   // We'd like to use Object.fromEntries() here but Node.js 10 does not support it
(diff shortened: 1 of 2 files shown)


dunghuynh-kdi:
@hjr3 Is there any chance we could get your approval for this PR, pretty please?

dunghuynh-kdi:
@hjr3 thank you!!

PR #3518, opened 2025-07-18
Add pg-transaction module

brianc:
I'm taking over the old [pg-transaction](https://www.npmjs.com/package/pg-transaction) module, and turning it into a simpler async/await affair.  I basically write this same utility in every project I do, so thought it'd be helpful.  Pass the `transaction` function a connected client __or__ an instance of a pool and it will call back with the client (or one checked out from the pool) already within a `BEGIN` block. If your async callback doesn't throw, it will `COMMIT` after, and if it does throw, it will `ROLLBACK`.  It also releases the client back to the pool regardless of if the your async callback function throws or not.


brianc:
~todo: add docs~ done

packages/pg-transaction/src/index.ts
```diff
@@ -0,0 +1,35 @@
+import type { Client, Pool, PoolClient } from 'pg'
+
+function isPoolClient(clientOrPool: Client | PoolClient): clientOrPool is PoolClient {
+  return 'release' in clientOrPool
+}
+
+function isPool(clientOrPool: Client | Pool): clientOrPool is Pool {
+  return 'idleCount' in clientOrPool
```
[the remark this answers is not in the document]
brianc: `client.connect` is also a function, so this would not work

docs/pages/features/transactions.mdx
```diff
@@ -37,3 +244,49 @@ try {
   client.release()
 }
 ```
+
+### When to Use Manual Transactions
```
charmander, line 248: (Is the README starting from “Best Practices” AI-generated? 😛 It kind of has that feel.)

I don’t see any reason these cases aren’t covered by pg-transactions, except maybe isolation levels (`SET TRANSACTION ISOLATION LEVEL …` requiring one extra query compared to `BEGIN ISOLATION LEVEL …`).

packages/pg-transaction/src/index.test.ts
```diff
@@ -0,0 +1,150 @@
...
+        // row should be visible within transaction
+        const { rows } = await client.query('SELECT * FROM test_table')
+        assert.equal(rows.length, 1, 'Row should be inserted within transaction')
+
+        // while inside this transaction, the changes are not visible outside
+        await withClient(async (innerClient) => {
+          const { rowCount } = await innerClient.query('SELECT * FROM test_table')
+          assert.equal(rowCount, 0, 'Temp table should still be empty inside transaction')
+        })
+      })
```
charmander, line 57: The pool version tests this, and the test description makes it sound like it was intended.

```suggestion

        // now that the transaction is committed, the changes are visible outside
        await withClient(async (innerClient) => {
          const { rowCount } = await innerClient.query('SELECT * FROM test_table')
          assert.equal(rowCount, 1, 'Row should be inserted after transaction commits')
        })
      })
```

(Used `rowCount` to match the preceding call; other places use `rows.length` 🤷)

packages/pg-transaction/src/index.ts
```diff
@@ -0,0 +1,38 @@
...
+  try {
+    const result = await cb(client)
+    await client.query('COMMIT')
+    return result
+  } catch (error) {
+    await client.query('ROLLBACK')
+    throw error
+  } finally {
+    if (isPoolClient(client)) {
+      client.release()
```
charmander, line 33: If the rollback failed, `client` could be returned to the pool while still in a transaction or in a broken-connection state. I don’t know if the former is actually possible in practice (`ROLLBACK` failing without breaking the connection), but the latter definitely is (when `error` doesn’t crash the process, e.g. because the user is attaching `error` listeners to all clients using the pool’s `connect` event). Might want to add another `catch` for the `ROLLBACK` and `.release(rollbackError)`.

packages/pg-transaction/src/index.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+async function transaction<T>(
+  clientOrPool: Client | Pool,
+  cb: (client: Client | PoolClient) => Promise<T>
+): Promise<T> {
+  let client: Client | PoolClient
+  if (isPool(clientOrPool)) {
+    // It's a Pool
+    client = await clientOrPool.connect()
```
charmander, line 18: This doesn’t do the `error` listener dance, so connection errors can crash a Node process here. It might be surprising that the behavior is different from `pool.query`. But maybe pg should add a default handler for the `error` event in a minor version update anyway?

brianc: > I think a lot of people probably use pg without doing this part strictly right…)

yeah, honestly myself included....for yeras, in multiple apps. 😬  

the whole async/evented error stuff is so so tricky to get "really just right" - would love to see an example if you have one. I'm totally cool working on some kind of "buffer the error(s) if no default `.on('error')` listener is added and throw it on the next call to the library" feature instead of what it is now which is effectively "if its not an in-flight query, then you're probably just gonna crash, which doesn't happen much, but async race conditions are their own kind of personal hell." I do think that's a minor version bump fix too because it could probably be backwards-comp where if there _is_ an error listener things proceede as normal but if there isn't we don't just explode the universe w/ an `unhandledError` and instead buffer it.

Honestly it could potentially be a lot to take on and needs some thinking about where the right place to handle these errors is (because legacy reasons we do pool -> client -> connection -> pg-protocol) callstacks, but its probably worth spending more time there myself.

charmander: I already did the error replay part in #1503; it’d just be a matter of adding `this.on('error', () => {})` to `Client`’s constructor (and `Pool`’s too).

packages/pg-transaction/src/index.ts
brianc: oh that's a good point. I'll take stab at this...I think I can repro by `pg_terminate_backend` the active in transaction client backed from within the transaction itself. That should disconnect it and make it fail to be able to send `rollback` - anyways....I'll test a few more failure modes for both this comment and the other one about the error listener dance and see if I can make it throw an uncaught anywhere else.

brianc:
this is not something I need to ship. need to focus on what already exists.

PR #3521, opened 2025-07-23
Handle SASL SCRAM server error responses

v0idpwn:
Add proper error handling for SCRAM-SERVER-FINAL-MESSAGE error attribute. The SCRAM specification allows servers to return error messages via the 'e' attribute in the server final message. Currently, these errors are ignored and authentication fails later during signature verification.

Postgres typically doesn't return this error (see [here](https://github.com/postgres/postgres/blob/2047ad068139f0b8c6da73d0b845ca9ba30fb33d/src/backend/libpq/auth-scram.c#L423) on why), but poolers, or other applications using the postgres protocol might, and it's part of the SCRAM spec, so it probably makes sense for node-postgres to handle it.

Aligns behaviour with psql, postgrex, and somewhat with pgJDBC (pgJDBC is stricter with scram errors).

For reference:

- libpq handling it: https://github.com/postgres/postgres/blob/2047ad068139f0b8c6da73d0b845ca9ba30fb33d/src/interfaces/libpq/fe-auth-scram.c#L708

### DIFF
--- a/packages/pg/lib/crypto/sasl.js
+++ b/packages/pg/lib/crypto/sasl.js
@@ -178,7 +178,13 @@ function parseServerFirstMessage(data) {
 
 function parseServerFinalMessage(serverData) {
   const attrPairs = parseAttributePairs(serverData)
+  const error = attrPairs.get('e')
   const serverSignature = attrPairs.get('v')
+
+  if (error) {
+    throw new Error(`SASL: SCRAM-SERVER-FINAL-MESSAGE: server returned error: "${error}"`)
+  }
+
   if (!serverSignature) {
     throw new Error('SASL: SCRAM-SERVER-FINAL-MESSAGE: server signature is missing')
   } else if (!isBase64(serverSignature)) {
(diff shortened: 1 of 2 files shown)


charmander (approved):

brianc (changes requested):
Is it possible to include an _integration test_ for this? One that hits the server? I want to make sure throwing here doesn't crash the client w/ an uncaught exception.

v0idpwn:
Hi, Brian. I'd be happy to, but I'm not sure about __how__, since Postgres itself doesn't return this error. Do you suggest some approach to do it?

For the record, I discovered this because [Supavisor](https://github.com/supabase/supavisor) returns scram errors. I sent a [PR](https://github.com/supabase/supavisor/pull/711) aligning it's behaviour with Postgres, but I decided to open this one too to align node-postgres scram implementation with the others.

brianc:
oh i see - what returns this error? or how can you get this error to trigger?

brianc:
what was worrying to me about this is throwing errors, depending on where they're thrown, might not be caught by the client and then emitted as an error event & instead result in `uncaughtExceptions` - I audited the code and it looks like all this SASL code is wrapped in try/catches & uses throwing to abort in the event of bad, malformed, unexpected responses and the like, so this is actually okay...and if its impossible to write an integration test for I'm okay to merge it like this. I don't love merging code that doesn't actually have an integration test around it with a real backed, but if there's no way to force/trick postgres into returning an `e` attribute, we can do it.

PR #3533, opened 2025-08-22
fix: Avoid retaining buffer for latest parse in reader

charmander:
The buffer can be arbitrarily large, and the parser shouldn’t keep it around while waiting on (and potentially also buffering) the next complete packet.


brianc:
commin' in a bit late here, but appreicate it!

PR #3542, opened 2025-09-07
docs(pg-native): installation

nigrosimone:
Add docker


charmander:
It does already say:

> - On Ubuntu/Debian: `apt-get install libpq-dev g++ make`

Should we just add `python3` to that and mention that it applies to the Debian-based Node images on Docker Hub?

nigrosimone:
Done: https://github.com/brianc/node-postgres/pull/3542/commits/67473c5fc5178f184298361b289d0ec3d5762a41

PR #3547, opened 2025-09-19
Improve Deno compatibility: config-first and safe env access

Jobians:
This PR introduces three changes to make the pg package more compatible with Deno while
keeping full Node.js functionality:

1. **Default user value**  
   - Replace `user: process.platform === 'win32' ? process.env.USERNAME : process.env.USER`  
     with `user: 'postgres'` in defaults.
   - Avoids requiring environment variables in Deno.

2. **Config-first parameter resolution**  
   - Update `val()` in connection-parameters.js to return `config[key]` first, before
     checking environment variables.
   - Prevents Deno errors when `--allow-env` is not granted.

3. **Safe NODE_PG_FORCE_NATIVE check**  
   - Wrap the `NODE_PG_FORCE_NATIVE` check in a `try/catch`.
   - Ensures `process.env` access in Deno doesn’t throw, while preserving Node.js behavior.

These changes maintain Node.js compatibility, preserve the lazy-loading of the native module,
and allow using the package in Deno without requiring `--allow-env`.

### DIFF
--- a/packages/pg/lib/connection-parameters.js
+++ b/packages/pg/lib/connection-parameters.js
@@ -7,6 +7,10 @@ const defaults = require('./defaults')
 const parse = require('pg-connection-string').parse // parses a connection string
 
 const val = function (key, config, envVar) {
+  if (config[key] !== undefined) {
+    return config[key]
+  }
+
   if (envVar === undefined) {
     envVar = process.env['PG' + key.toUpperCase()]
   } else if (envVar === false) {
@@ -15,7 +19,7 @@ const val = function (key, config, envVar) {
     envVar = process.env[envVar]
   }
 
-  return config[key] || envVar || defaults[key]
+  return envVar || defaults[key]
 }
 
 const readSSLConfigFromEnvironment = function () {
--- a/packages/pg/lib/index.js
+++ b/packages/pg/lib/index.js
@@ -34,31 +34,37 @@ const PG = function (clientConstructor) {
   this.utils = utils
 }
 
-if (typeof process.env.NODE_PG_FORCE_NATIVE !== 'undefined') {
-  module.exports = new PG(require('./native'))
-} else {
-  module.exports = new PG(Client)
+let clientConstructor = Client
 
-  // lazy require native module...the native module may not have installed
-  Object.defineProperty(module.exports, 'native', {
-    configurable: true,
-    enumerable: false,
-    get() {
-      let native = null
-      try {
-        native = new PG(require('./native'))
-      } catch (err) {
-        if (err.code !== 'MODULE_NOT_FOUND') {
-          throw err
-        }
-      }
-
-      // overwrite module.exports.native so that getter is never called again
-      Object.defineProperty(module.exports, 'native', {
-        value: native,
-      })
-
-      return native
-    },
-  })
+try {
+  if (process.env.NODE_PG_FORCE_NATIVE) {
+    clientConstructor = require('./native')
+  }
+} catch {
+  // ignore, e.g., Deno without --allow-env
 }
+
+module.exports = new PG(clientConstructor)
+
+// lazy require native module...the native module may not have installed
+Object.defineProperty(module.exports, 'native', {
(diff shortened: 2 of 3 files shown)


packages/pg/lib/connection-parameters.js
```diff
@@ -7,6 +7,10 @@ const defaults = require('./defaults')
 const parse = require('pg-connection-string').parse // parses a connection string
 
 const val = function (key, config, envVar) {
+  if (config[key] !== undefined) {
```
charmander, line 10: Original expression used truthiness:

```suggestion
  if (config[key]) {
```

packages/pg/lib/index.js
```diff
@@ -34,31 +34,37 @@ const PG = function (clientConstructor) {
...
-      } catch (err) {
-        if (err.code !== 'MODULE_NOT_FOUND') {
-          throw err
-        }
+try {
+  if (process.env.NODE_PG_FORCE_NATIVE) {
+    clientConstructor = require('./native')
+  }
+} catch {
+  // ignore, e.g., Deno without --allow-env
```
charmander, line 44: Or `require('./native')` failed. The `try` should only cover the `process.env` access (if we actually want to do it this way at all).

charmander (changes requested):

Jobians:
@charmander

Jobians:
??? I will close this is you don't respond again

packages/pg/lib/index.js
```diff
@@ -34,31 +34,40 @@ const PG = function (clientConstructor) {
...
-          throw err
-        }
+let forceNative = false
+try {
+  forceNative = !!process.env.NODE_PG_FORCE_NATIVE
+} catch {
+  // ignore, e.g., Deno without --allow-env
+}
+
+if (forceNative) {
```
NickCrews: If I am reading this right, if you are in a deno environment with no --allow-env, then the above will always error, and forceNative will always remain in its default of false, and so it will be impossible to use the native driver. I'm not sure how much of a dealbreaker this actually is.

NickCrews (commented):
I think the biggest downside of the old pattern, and this suggested implementation, is that they bind too early. We don't need to, and therefore shouldn't decide which backend to use (either the js backend or the native backend) until the user actually tries to create a Client. If we defer this decision until then, then we open up the possibility for the user to request a different backend using a different API beside env vars (because if you decide at import time, this package has no APIs available yet for the user to inform us what they want).

So what about something like the following, which then provides an actual, non-env way to configure which backend the user wants? I don't love the use of a global variable to define the the behavior, but it is the least api-disruptive

```javascript

const PG = function (clientConstructor) {
  this.defaults = defaults
  this.Client = clientConstructor
  this.Query = this.Client.Query
  this.Pool = poolFactory(this.Client)
  this._pools = []
  this.Connection = Connection
  this.types = require('pg-types')
  this.DatabaseError = DatabaseError
  this.TypeOverrides = TypeOverrides
  this.escapeIdentifier = escapeIdentifier
  this.escapeLiteral = escapeLiteral
  this.Result = Result
  this.utils = utils
  // A new property/config item we add? Not needed if you don't want to
  // increase your API surface area, or maybe requires some more thought.
  this.forceNative = false
}

function makeClient(config) {
    // start with the current config setting
    let useNative = module.exports.forceNative
    try {
        useNative = typeof process.env.NODE_PG_FORCE_NATIVE !== 'undefined'
    } catch {
        // ignore, eg in Deno sandbox without --allow-env
    }
    const actualClient = useNative ? require('./native') : Client
    return actualClient(config)
}

module.exports = new PG(makeClient)

// lazy require native module...the native module may not have installed
  Object.defineProperty(module.exports, 'native', {
...
```

eg user does

```javascript
import {Client, forceNative} from 'pg'

forceNative = true
const nativeClient = Client(...)
forceNative = false
const jsClient = Client(...)
```

Jobians:
Hello what is the issue? If you don't want to merge just close it

Issue #3194, opened 2024-04-11
Performance issue with the `escapeLiteral` utility


mrmlnc:
Hello,

In my application, I use a field with the jsonb type. The size of this field can be more than 1000 characters. There can be many text-based columns in the table.

I've noticed that insert queries with long strings take a long time. My investigation led me to the `escapeLiteral` function. This function iterates over the entire line and builds it again.

Here are some measurements (node@20) showing the growth of operation time depending on the length of the string:

| Input string length | Function execution time |
| -------- | ------- |
| 11  | 0.053ms |
| 110 | 0.065ms    |
| 1100 | 0.11ms |
| 11_000    | 0.554ms   |
| 110_000 | 4.946ms |

<details>

```js
var input = "hello world".repeat(10000)

console.time('\nperf_escape_literal')
utils.escapeLiteral(input)
console.timeEnd('\nperf_escape_literal')

console.dir(input.length)
```

</details>

An example of a simplified string that is written to the `jsonb` column. We also have users with a large amount of data (the string will be longer x2…10).

```js
[
 {"type": "local", "action": "Open projects page", "expectation": ""},
 {"type": "local", "action": "Click on button \"Create new project\"", "expectation": "You are being redirected to the new project form"},
 {"type": "local", "action": "Fill the form", "expectation": "Project code should be generated automatically"},
 {"type": "local", "action": "Click on \"Create project\" button", "expectation": "You are being redirected to repository page with no cases or suites"},
 {"hash": "1_4", "type": "shared", "action": "Do action 0", "version": 23, "expectation": "Action 0 is indicated as 0.13738214216469524 on the ui", "sharedStepsGroupId": 1},
 {"hash": "1_4", "type": "shared", "action": "Swipe 0.4240689256483676", "version": 23, "expectation": "The sun is shining with the color temperature of 0.8792701581028319", "sharedStepsGroupId": 1},
 {"hash": "1_4", "type": "shared", "action": "Rotate at 0.3493667640291416 degrees", "version": 23, "expectation": "Wind speed is changed to 0 m/s", "sharedStepsGroupId": 1}
]

perf_escape_literal: 0.104ms
input_length: 1014
```

At the moment, these are small numbers, but escaping is called for each text-based column. In total, these are significant numbers. For example, when adapting the solution below, we got an speed-up from ~950ms to ~550ms for some insert queries.

[...]

Issue #957, opened 2016-03-03
How to insert multiple rows with a single query?


zettam:
I have `n` number of items. I would like to add them all with a single insert query.

`insert into testtable(id,name) values (1,"jack"),(2,"john"),(3,"jill");`

I have an array for my rows, and I like to add them all in. So, if I provide `pg` the correct string (after manually forming it in a variable) it works perfectly fine, but this all seemed a bit pointless to me. I expect pg to handle arrays for me as well, but I'm unable to find the option for that. So I'm asking, is this possible?

How can I insert multiple rows from an array of rows, with a single insert query with this library?

Would love to know. Thank you.

vitaly-t:
No such thing supported. This is not an ORM, this library just executes queries.

zettam:
hm, okay.
What's the best way to handle this situation in this case? Sending multiple queries is pretty costy hen you can do the same with a single query.

vitaly-t:
You can concatenate your inserts. However, since this library doesn't provide a separate query formatting, you can't really do it correctly.

You can do it with [pg-promise](https://github.com/vitaly-t/pg-promise) though. Article [Performance Boost](https://github.com/vitaly-t/pg-promise/wiki/Performance-Boost) should echoes all your concerns nicely.

langpavel:
Prepare your insert statement

rpedela:
Using [pg-format](https://github.com/datalanche/node-pg-format) with a nested array:

```js
var format = require('pg-format');

var values = [
    [ 1, 'jack' ],
    [ 2, 'john' ],
    [ 3, 'jill' ],
];
console.log(format('INSERT INTO test_table (id, name) VALUES %L', values));
// INSERT INTO test_table (id, name) VALUES ('1', 'jack'), ('2', 'john'), ('3', 'jill')
```

Disclaimer: I wrote [pg-format](https://github.com/datalanche/node-pg-format).

awenger:
You could also do something like:
```
var query = 'INSERT INTO table (SELECT (data->>'c1')::int, (data->>'c2')::text FROM (SELECT json_array_elements($1::json) AS data) tmp)'
var data = JSON.stringify([{c1: 1, c2: 'text1'},{c1: 2, c2: 'text2'}])

client.query(query, [data], ...)
```

Formatted version of the query:
```
INSERT INTO table (
  SELECT 
    -- extract the props from the json object so that they match the table cols
    (data->>'c1')::int, (data->>'c2')::text
[...]

PR #3553, opened 2025-10-03
Improve escapeLiteral performance of long strings by 30x

Llorx:
I noticed that on long strings, `escapeLiteral` goes really slow. I created my own `escapeLiteral` and I decided to push it. I saw this issue when searching: https://github.com/brianc/node-postgres/issues/3194

On short strings, the performance loss is very low compared to the benefit on long strings. On short strings the original is only 1.17x faster, but on long strings the new one is 30x (!!) faster, so at the moment that the strings are a bit longer than 10 characters, the performance profit is going to kick in, and the performance loss on short strings is negligible.

Anyway, with this information I decided to add a quick check: If the string is shorter than 13 characters, use the original one, and if it is longer, use the optimized one. It has the performance profit in both situations:
image

EDIT: Had a typo in the benchmark, it is *only* 30x faster, not 55x.

Here the benchmark code:
```javascript
const { IsoBench } = require("iso-bench");

function escapeLiteralOriginal(str) {
  let hasBackslash = false
  let escaped = "'"

  if (str == null) {
    return "''"
  }

  if (typeof str !== 'string') {
    return "''"
  }

  for (let i = 0; i < str.length; i++) {
    const c = str[i]
    if (c === "'") {
      escaped += c + c
    } else if (c === '\\') {
      escaped += c + c
      hasBackslash = true
    } else {
      escaped += c
    }
  }

  escaped += "'"

  if (hasBackslash === true) {
    escaped = ' E' + escaped
  }

  return escaped
}
const escapeLiteralNew = function (str) {
  if (typeof str !== 'string') {
    return "''"
  }
  let hasBackslash = false
  if (str.length < 13) {
    let escaped = "'"

    for (let i = 0; i < str.length; i++) {
      const c = str[i]
      if (c === "'") {
        escaped += c + c
      } else if (c === '\\') {
        escaped += c + c
        hasBackslash = true
      } else {
        escaped += c
      }
    }
  
    escaped += "'"
  
    if (hasBackslash === true) {
      escaped = ' E' + escaped
    }
    return escaped
  } else {
    let escaped = str
        .replace(/\\/g, () => {
            hasBackslash = true
            return '\\\\'
        })
        .replace(/'/g, "''")
    
    if (hasBackslash) {
        escaped = ` E'${escaped}'`
    } else {
        escaped = `'${escaped}'`
    }
    return escaped
  }
}
function newArray(size, backslash, quote, count = 100) {
    return new Array(count).fill(0).map((_, i) => {
        const str = "a".repeat(size);
        if (backslash) {
            const pos = i % size;
            str[pos] = "\\";
        }
        if (quote) {
            const pos = (i + 2) % size;
            str[pos] = "'";
        }
        return str;
    });
}
new IsoBench()
    .add("escapeLiteralNew", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralNew(arr[i]);
        }
    }, () => newArray(10, false, false))
    .add("escapeLiteralOriginal", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralOriginal(arr[i]);
        }
    }, () => newArray(10, false, false))
.endGroup("short string, backslash = false, quote = false")

    .add("escapeLiteralNew", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralNew(arr[i]);
        }
    }, () => newArray(10, true, false))
    .add("escapeLiteralOriginal", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralOriginal(arr[i]);
        }
    }, () => newArray(10, true, false))
.endGroup("short string, backslash = true, quote = false")

    .add("escapeLiteralNew", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralNew(arr[i]);
        }
    }, () => newArray(10, false, true))
    .add("escapeLiteralOriginal", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralOriginal(arr[i]);
        }
    }, () => newArray(10, false, true))
.endGroup("short string, backslash = false, quote = true")

    .add("escapeLiteralNew", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralNew(arr[i]);
        }
    }, () => newArray(10, true, true))
    .add("escapeLiteralOriginal", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralOriginal(arr[i]);
        }
    }, () => newArray(10, true, true))
.endGroup("short string, backslash = true, quote = true")

    .add("escapeLiteralNew", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralNew(arr[i]);
        }
    }, () => newArray(1000, false, false))
    .add("escapeLiteralOriginal", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralOriginal(arr[i]);
        }
    }, () => newArray(1000, false, false))
.endGroup("long string, backslash = false, quote = false")

    .add("escapeLiteralNew", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralNew(arr[i]);
        }
    }, () => newArray(1000, true, false))
    .add("escapeLiteralOriginal", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralOriginal(arr[i]);
        }
    }, () => newArray(1000, true, false))
.endGroup("long string, backslash = true, quote = false")

    .add("escapeLiteralNew", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralNew(arr[i]);
        }
    }, () => newArray(1000, false, true))
    .add("escapeLiteralOriginal", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralOriginal(arr[i]);
        }
    }, () => newArray(1000, false, true))
.endGroup("long string, backslash = false, quote = true")

    .add("escapeLiteralNew", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralNew(arr[i]);
        }
    }, () => newArray(1000, true, true))
    .add("escapeLiteralOriginal", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralOriginal(arr[i]);
        }
    }, () => newArray(1000, true, true))
.endGroup("long string, backslash = true, quote = true")
    .consoleLog()
    .run();
```

### DIFF
--- a/packages/pg/lib/utils.js
+++ b/packages/pg/lib/utils.js
@@ -183,13 +183,7 @@ const escapeLiteral = function (str) {
           return '\\\\'
       })
       .replace(/'/g, "''")
-
-  if (hasBackslash) {
-      escaped = ` E'${escaped}'`
-  } else {
-      escaped = `'${escaped}'`
-  }
-  return escaped
+  return hasBackslash ? ` E'${escaped}'` : `'${escaped}'`
 }
 
 module.exports = {


Llorx:
*I don't understand the error the linter is returning*

charmander:
> ```js
> if (backslash) {
>     const pos = i % size;
>     str[pos] = "\\";
> }
> if (quote) {
>     const pos = (i + 2) % size;
>     str[pos] = "'";
> }
> ```

`str` is a string (immutable), so this benchmark doesn’t actually test any quotes or backslashes.

Llorx:
> > ```js
> > if (backslash) {
> >     const pos = i % size;
> >     str[pos] = "\\";
> > }
> > if (quote) {
> >     const pos = (i + 2) % size;
> >     str[pos] = "'";
> > }
> > ```
> 
> `str` is a string (immutable), so this benchmark doesn’t actually test any quotes or backslashes.

Ah my bad. I was working with buffers and failed miserably when doing the benchmark haha. Here is the fixed benchmark:
image

Backslash penalizes way more than quote because of the callback to avoid another iteration to find a backslash. Still a great performance improvement, specially when there's nothing to clean (which is the usual case).

The code:
```javascript
const { IsoBench } = require("iso-bench");

function escapeLiteralOriginal(str) {
  let hasBackslash = false
  let escaped = "'"

  if (str == null) {
    return "''"
  }

  if (typeof str !== 'string') {
    return "''"
  }

  for (let i = 0; i < str.length; i++) {
    const c = str[i]
    if (c === "'") {
      escaped += c + c
    } else if (c === '\\') {
      escaped += c + c
      hasBackslash = true
    } else {
      escaped += c
    }
  }

  escaped += "'"

  if (hasBackslash === true) {
    escaped = ' E' + escaped
  }

  return escaped
}
const escapeLiteralNew = function (str) {
  if (typeof str !== 'string') {
    return "''"
  }
  let hasBackslash = false
  if (str.length < 13) {
    let escaped = "'"

    for (let i = 0; i < str.length; i++) {
      const c = str[i]
      if (c === "'") {
        escaped += c + c
      } else if (c === '\\') {
        escaped += c + c
        hasBackslash = true
      } else {
        escaped += c
      }
    }
  
    escaped += "'"
  
    if (hasBackslash === true) {
      escaped = ' E' + escaped
    }
    return escaped
  } else {
    let escaped = str
        .replace(/\\/g, () => {
            hasBackslash = true
            return '\\\\'
        })
        .replace(/'/g, "''")
    
    if (hasBackslash) {
        escaped = ` E'${escaped}'`
    } else {
        escaped = `'${escaped}'`
    }
    return escaped
  }
}
function newArray(size, backslash, quote, count = 100) {
    return new Array(count).fill(0).map((_, i) => {
        let str = "a".repeat(size);
        if (backslash) {
            const pos = i % size;
            str = `${str.substring(0, pos)}\\${str.substring(pos)}`;
        }
        if (quote) {
            const pos = (i + 2) % size;
            str = `${str.substring(0, pos)}'${str.substring(pos)}`;
        }
        return str;
    });
}
new IsoBench()
    .add("escapeLiteralNew", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralNew(arr[i]);
        }
    }, () => newArray(10, false, false))
    .add("escapeLiteralOriginal", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralOriginal(arr[i]);
        }
    }, () => newArray(10, false, false))
.endGroup("short string, backslash = false, quote = false")

    .add("escapeLiteralNew", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralNew(arr[i]);
        }
    }, () => newArray(10, true, false))
    .add("escapeLiteralOriginal", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralOriginal(arr[i]);
        }
    }, () => newArray(10, true, false))
.endGroup("short string, backslash = true, quote = false")

    .add("escapeLiteralNew", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralNew(arr[i]);
        }
    }, () => newArray(10, false, true))
    .add("escapeLiteralOriginal", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralOriginal(arr[i]);
        }
    }, () => newArray(10, false, true))
.endGroup("short string, backslash = false, quote = true")

    .add("escapeLiteralNew", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralNew(arr[i]);
        }
    }, () => newArray(10, true, true))
    .add("escapeLiteralOriginal", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralOriginal(arr[i]);
        }
    }, () => newArray(10, true, true))
.endGroup("short string, backslash = true, quote = true")

    .add("escapeLiteralNew", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralNew(arr[i]);
        }
    }, () => newArray(1000, false, false))
    .add("escapeLiteralOriginal", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralOriginal(arr[i]);
        }
    }, () => newArray(1000, false, false))
.endGroup("long string, backslash = false, quote = false")

    .add("escapeLiteralNew", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralNew(arr[i]);
        }
    }, () => newArray(1000, true, false))
    .add("escapeLiteralOriginal", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralOriginal(arr[i]);
        }
    }, () => newArray(1000, true, false))
.endGroup("long string, backslash = true, quote = false")

    .add("escapeLiteralNew", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralNew(arr[i]);
        }
    }, () => newArray(1000, false, true))
    .add("escapeLiteralOriginal", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralOriginal(arr[i]);
        }
    }, () => newArray(1000, false, true))
.endGroup("long string, backslash = false, quote = true")

    .add("escapeLiteralNew", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralNew(arr[i]);
        }
    }, () => newArray(1000, true, true))
    .add("escapeLiteralOriginal", (arr) => {
        for (let i = 0; i < arr.length; i++) {
            escapeLiteralOriginal(arr[i]);
        }
    }, () => newArray(1000, true, true))
.endGroup("long string, backslash = true, quote = true")
    .consoleLog()
    .run();
```

Llorx:
> I’m against the complexity of having a second path for short strings. (I don’t even think this function should have a second path for strings without backslashes, but that’d have to be a pg 9 change.) There are probably cleaner optimization options, since IIRC the original implementation here was just ported from C.

Yeah, then I can go for the first one, which is way less complex.

> Also not sure how much performance matters past a certain point: as mentioned in #3194, people should pretty much always be using parameters instead of `escapeLiteral` anyway.

The problem is that the performance of parameters is very low. I have a 4x penalty with 5k strings being 100k characters long each one, so if we can help the users to improve performance by just writing optimized code, I think that it should be the way to go. They still have the option to go the low performant way in this case with parameters if they want, but if they need performance (which is pretty usual when you are moving hundreds of thousands of characters), they can use the included utils in the library.

If we keep `escapeLiteral` the old way, the only options that you have is a very very low performant way with the included `escapeLiteral` function, or a *just very* low performant way using parameters. Or to write your own `escapeLiteral` function, which already having a `escapeLiteral` function in the library feels very redundant...

This function is actually the one I use because I need that performance, and I'm not alone as there's another ticket with that question. And also there are going to be users that just struggle with the performance because there's nothing better but don't open a issue: "this insert query lasts 5 seconds. It is what it is. Next ticket". So I just decided to publish it to help users gain this performance they didn't knew they were losing.

> The linter is complaining about trailing spaces.

Ok, problems of the GitHub editor. I've done it online directly. Going to find the trailing spaces.

EDIT: Can't find the trailing spaces in the editor. If you like this, I'll clone and `--fix` it for you to merge.

charmander:
> The problem is that the performance of parameters is very low. I have a 4x penalty with 5k strings being 100k characters long each one, so if we can help the users to improve performance by just writing optimized code, I think that it should be the way to go. They still have the option to go the low performant way in this case with parameters if they want, but if they need performance (which is pretty usual when you are moving hundreds of thousands of characters), they can use the included utils in the library.

How are you using parameters? Do you have a benchmark you can share for this too? It’s possible that a parameter-based approach is inherently slower with PostgreSQL, but it’s also possible that pg is just missing several optimizations that we could implement[^1], with a much cleaner result than serializing data into code – and/or that the way you’re using parameters is suboptimal.

> Or to write your own `escapeLiteral` function, which already having a `escapeLiteral` function in the library feels very redundant...

I know it feels that way, but consider that the library’s `escapeLiteral` acts the way it does for compatibility with a previous bug (#3489) and for compatibility with libpq’s `PQescapeLiteral`, which itself is bogged down by compatibility concerns. If inserting data into your code truly turns out to be the best option and you want maximum performance, consider [asserting that `standard_conforming_strings` is `on`](https://www.postgresql.org/docs/current/runtime-config-compatible.html#GUC-STANDARD-CONFORMING-STRINGS) and using a simple and fast implementation:

```js
const escapeLiteralStandard = str =>
  `'${str.replaceAll("'", "''")}'`;
```

or

```js
const escapeLiteralStandard = str => {
  if (typeof str !== 'string') {
    throw new TypeError('string expected');
  }

  let escaped = "'";
  let start = 0;

  for (let i; (i = str.indexOf("'", start)) !== -1;) {
    escaped += str.substring(start, i);
    escaped += "''";
    start = i + 1;
  }

  escaped += str.substring(start);
  escaped += "'";

  return escaped;
};
```

If being one configurable setting away from SQL injection makes you uneasy, I’d still opt for consistently returning C-escape string literals:

```js
const escapeLiteralPg = str => {
  if (typeof str !== 'string') {
    throw new TypeError('string expected');
  }

  let escaped = "E'";
  let start = 0;
  const re = /[\\']/g;

  for (let m; (m = re.exec(str)) !== null;) {
    escaped += str.substring(start, m.index);
    escaped += '\\';
    escaped += m[0];
    start = m.index + 1;
  }

  escaped += str.substring(start);
  escaped += "'";

  return escaped;
};
```

or

```js
const escapeLiteralPg = str => {
  if (typeof str !== 'string') {
    throw new TypeError('string expected');
  }

  let escaped = "E'";
  let start = 0;

  for (;;) {
    const i = str.indexOf("'", start);
    const sub = i === -1 ? str : str.substring(0, i);

    for (let j; (j = sub.indexOf('\\', start)) !== -1;) {
      escaped += str.substring(start, j);
      escaped += '\\\\';
      start = j + 1;
    }

    if (i === -1) {
      break;
    }

    escaped += str.substring(start, i);
    escaped += "\\'";
    start = i + 1;
  }

  escaped += str.substring(start);
  escaped += "'";

  return escaped;
};
```

[^1]: I worked on this recently, and pg *is* missing several optimizations. But it’d be nice to know if they’d solve your problem here too.

Llorx:
> How are you using parameters? Do you have a benchmark you can share for this too? It’s possible that a parameter-based approach is inherently slower with PostgreSQL, but it’s also possible that pg is just missing several optimizations that we could implement[1](#user-content-fn-1-fe63bf233be2d94f706c3304debde445), with a much cleaner result than serializing data into code – and/or that the way you’re using parameters is suboptimal.

I don't have a benchmark. Just a `console.time/timeEnd()` code:
```javascript
const { Pool } = require("pg");

const arr = new Array(5000).fill(0).map((_, i) => ({ id: i, value: "a".repeat(100000) }));
const pool = new Pool({
    connectionString: "postgres://postgres:<password>@127.0.0.1:5432/test"
});
async function test1() {
    await client.query("CREATE TABLE IF NOT EXISTS my_test (id SERIAL PRIMARY KEY, value TEXT);");
    console.time("custom escapeLiteral");
    await client.query(`INSERT INTO my_test ` +
        `(id, value) ` +
        `VALUES ${arr.map(sample => {
            return `(${sample.id}, ${escapeLiteral(sample.value)})`
        }).join(",")}`);
    console.timeEnd("custom escapeLiteral");
    await client.query("DROP TABLE my_test");
    client.release();
}
async function test2() {
    await client.query("CREATE TABLE IF NOT EXISTS my_test2 (id SERIAL PRIMARY KEY, value TEXT);");
    console.time("parameters");
    await client.query(`INSERT INTO my_test2 ` +
        `(id, value) ` +
        `VALUES ${arr.map((sample, i) => {
            return `(${sample.id}, $${i + 1})`
        }).join(",")}`, arr.map(sample => sample.value));
    console.timeEnd("parameters");
    await client.query("DROP TABLE my_test2");
    client.release();
}
const escapeLiteral = function (str) {
    let hasBackslash = false
    let escaped = str
        .replace(/\\/g, () => {
            hasBackslash = true
            return '\\\\'
        })
        .replace(/'/g, "''")
    return hasBackslash ? ` E'${escaped}'` : `'${escaped}'`
}
test1().then(test2);
```

PD: This is weird, now suddenly they are going on-par. yesterday the first one was 9 seconds and the second one 35 seconds.
PD2: I remember now. It was not 5k strings, but 2 batches of 5k strings because the first one can't handle so much characters in the query. And I guess that some kind of optimization/cache/whatever is kicking in because subsequent iterations run way faster. Today I went with a 10k parameters test again and it was 31 seconds, but the second time just 9 seconds.

Knowing this, I think that I'm going to go with parameters then. I will let the SO or database do their optimizations and that's it.

> I know it feels that way, but consider that the library’s `escapeLiteral` acts the way it does for compatibility with a previous bug (#3489) and for compatibility with libpq’s `PQescapeLiteral`, which itself is bogged down by compatibility concerns. If inserting data into your code truly turns out to be the best option and you want maximum performance, consider [asserting that `standard_conforming_strings` is `on`](https://www.postgresql.org/docs/current/runtime-config-compatible.html#GUC-STANDARD-CONFORMING-STRINGS) and using a simple and fast implementation:

I don't have control over the database :-( that's why I'm testing extreme cases, because some users have weird structures like storing big JSON strings, and that's out of our control, but I don't want them to suddenly have a loss in performance and have to cope with them saying "you are doing weird things, you should change everything to this optimal way". I prefer for the software to be the most performant possible in the most weird situations, and let them be happy with their uberJSONs (or whatever big string they are using).

charmander:
Using array parameters instead of generating queries with large, variable numbers of parameters might help: https://github.com/brianc/node-postgres/issues/957#issuecomment-295583050

pg currently serializes arrays with a text representation (less efficient than the binary representation) and copies them more than it should, though.

Issue #1184, opened 2016-12-13
Validity of callback not checked


Carbonyte:
```js
client.query("SELECT now()", [], {});
```

This will happily execute, attempt to call the object, and crash Node. This isn't really the fault of node-postgres, but there's pretty much no way to catch the error besides process.on('uncaughtException'), and the Node docs explicitly advice against attempting to recover from any errors caught from there. Considering that this is a perfectly recoverable situation, this is not ideal, so node-postgres should check if the callback is actually a function before proceeding with the query. Even if it just throws an error, doing it before the query runs ensures the error's call stack contains code written by the developer, which allows it to be caught properly.

brianc:
Adding to 7.0 milestone.

brianc:
Thinking about this a bit more - I don't think this is a backwards incompatible change. Previous this would just break & destroy your app without you having a way to recover, so it doesn't need to be in 7.0.

brianc:
_note: I still think it's important to fix - just trying to prioritize 7.0 work myself. Happy to take a PR to fix this though as well!_

PR #3561, opened 2025-11-10
Add error handling for non-function callback

dotob:
catch callback not a function earlier to get a proper callstack. later when executing the callback the stack may be wrong/insufficient.


dotob:
this may fix: https://github.com/brianc/node-postgres/issues/1184

brianc:
Thanks @dotob

PR #3571, opened 2025-12-15
chore(pg-native): use ES6 classes

cesco69:
This improves readability and may allow V8 to optimize better.

### DIFF
--- a/packages/pg-native/index.js
+++ b/packages/pg-native/index.js
@@ -5,7 +5,7 @@ const types = require('pg-types')
 const buildResult = require('./lib/build-result')
 const CopyStream = require('./lib/copy-stream')
 
-class ClientClass extends EventEmitter {
+class Client extends EventEmitter {
   constructor(config) {
     super()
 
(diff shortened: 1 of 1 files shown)


packages/pg-native/index.js
```diff
@@ -1,331 +1,330 @@
...
-  this._results = undefined
-
-  this._resultCount = 0
+// export the version number so we can check it in node-postgres
+module.exports.version = require('./package.json').version
 
-  if (cb) {
-    cb(err, rows || [], results)
-  }
+module.exports = function Client(config) {
```
charmander: Breaks `instanceof`/inheritance and useful properties like `prototype` and the `version` set on `Client` immediately above as far as existing `const Client = require('pg-native')` users are concerned. I think it’s best to make this a breaking change in a new major version.

charmander (changes requested):

packages/pg-native/index.js
```diff
@@ -1,331 +1,332 @@
...
-  this._results = undefined
-
-  this._resultCount = 0
-
-  if (cb) {
-    cb(err, rows || [], results)
-  }
-}
+module.exports = createClient
+module.exports.Client = Client
```
charmander: Among the issues I pointed out, that only fixes `version`. It’s not fixable in a clean way overall. Pretty sure we should just do a regular class with no associated compatibility function and merge it for pg-native 4.0.0.

PR #3572, opened 2025-12-16
perf: no need to create a new array and new anonymous function

cesco69:
`this._idle.slice().map( ... )` create a new array and new anonymous function every time


packages/pg-pool/index.js
```diff
@@ -125,9 +125,10 @@ class Pool extends EventEmitter {
     if (this.ending) {
       this.log('pulse queue on ending')
       if (this._idle.length) {
```
charmander: `if` is now fully redundant.

PR #3574, opened 2025-12-22
Fix error handling for exceptions on values parsing.

### LINKED ISSUE #3573: Exception in values serialisation causes next query to use invalid statements.
There are two bugs in different parts of the library which, can lead to unpredictable behavior if value serialization fails.

1. Serializer state is not cleaned up on error

In the serializer, `writer` and `paramWriter` may not be cleaned up if an exception is thrown in the `writeValues` function:
https://github.com/brianc/node-postgres/blob/ecff60dc8aa0bd1ad5ea8f4623af0756a86dc110/packages/pg-protocol/src/serializer.ts#L155

As a result, parameters from one query (which failed, but may was processed partially) can be sent to next query. I am not an expert in the PostgreSQL protocol, but it appears that reusing global writers in this way can theoretically lead to other unpredictable behavior as well.

When I made a local change to avoid reusing the writer, a second issue became apparent.

2. Prepared statement is not properly closed on bind error

In the query `prepare` function, if an error occurs during the `bind` call (triggered by the first issue), the prepared statement is not properly closed:

https://github.com/brianc/node-postgres/blob/ecff60dc8aa0bd1ad5ea8f4623af0756a86dc110/packages/pg/lib/query.js#L231

When using a pooled client, this causes the client to remain in an active state indefinitely, effectively leaking the connection.

I prepared PR to address those issues, but it would be nice to hear a feedback from active maintainers on this.

APshenkin:
Fixes https://github.com/brianc/node-postgres/issues/3573

- Fix: Avoid connection leaks by ensuring parse closure on error.
- Refactor: Move `Writer` instantiation to improve isolation and prevent state reuse issues.

### DIFF
--- a/packages/pg-protocol/src/buffer-writer.ts
+++ b/packages/pg-protocol/src/buffer-writer.ts
@@ -82,4 +82,9 @@ export class Writer {
     this.buffer = Buffer.allocUnsafe(this.size)
     return result
   }
+
+  public clear(): void {
+    this.offset = 5
+    this.headerPosition = 0
+  }
 }
--- a/packages/pg-protocol/src/outbound-serializer.test.ts
+++ b/packages/pg-protocol/src/outbound-serializer.test.ts
@@ -273,4 +273,62 @@ describe('serializer', () => {
     const expected = new BufferList().addInt16(1234).addInt16(5678).addInt32(3).addInt32(4).join(true)
     assert.deepEqual(actual, expected)
   })
+
+  describe('bind error recovery', () => {
+    const throwingMapper = () => {
+      throw new Error('valueMapper error')
+    }
+
+    it('produces correct bind output after a valueMapper exception', () => {
+      assert.throws(() => {
+        serialize.bind({
+          values: ['fail'],
+          valueMapper: throwingMapper,
+        })
+      }, /valueMapper error/)
+
+      const actual = serialize.bind({
+        portal: 'bang',
+        statement: 'woo',
+        values: ['1', 'hi', null, 'zing'],
+      })
+      const expectedBuffer = new BufferList()
+        .addCString('bang')
+        .addCString('woo')
+        .addInt16(4)
+        .addInt16(0)
+        .addInt16(0)
+        .addInt16(0)
+        .addInt16(0)
+        .addInt16(4)
+        .addInt32(1)
+        .add(Buffer.from('1'))
+        .addInt32(2)
+        .add(Buffer.from('hi'))
+        .addInt32(-1)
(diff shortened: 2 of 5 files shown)


charmander (changes requested):
Please add tests if possible. (Ideally, point out the minimal fix for each bug and which changes are just cleanup, e.g. by separating into 2–3 commits, but the tests will help make that more clear anyway.)

APshenkin:
@charmander 

I would really appreciate to get a feedback first, that those changes are in right direction. Those two changes are tight together as they both are related to serialization issue. Just because I solved global writer issue first, zombie client appeared next in PgPool, because without fixing writers you just write sync/close in next query

I'm not an expert of pg protocol and just investigated this issue few hours.
Maybe there were a reasons to use shared writers? If not and those changes makes sence, would be happy to do what you request

charmander:
Yes, these changes make sense. The shared writer was probably an attempt at improving performance.

brianc:
> * Refactor: Move `Writer` instantiation to improve isolation and prevent state reuse issues.

If this is something which can be acomplished without reinstantiating the writer I would greatly prefer it. This is extremely hot path code. Is there a way to fix this w/o? Also: needs tests before merging. 

Thanks for the eyes on this tho! I'll leave it open for a bit as it might be something I can quickly fix (IF there is a repoducable test case on how to trigger the issue)

APshenkin:
@charmander @brianc 

Sorry, I missed these comments, my bad.

I'll try to take a look on this back in a next few weeks (split to different commits, adding tests)

Re
> If this is something which can be acomplished without reinstantiating the writer I would greatly prefer it.

I think it should be doable without reinstantiating the writer, however the fix will be fragile then for any future changes. Also code will be a bit ugly (each time when you have some error you should not forget to clean it properly). I'll try to play around it and come back with something

brianc:
> I think it should be doable without reinstantiating the writer, however the fix will be fragile then for any future changes. Also code will be a bit ugly (each time when you have some error you should not forget to clean it properly). I'll try to play around it and come back with something

sounds good - i think we can just re-instantiate it then. allocating something is pretty cheap.

brianc:
Is it possible to write a couple tests for this?  I'd like to get it merged.

APshenkin:
@brianc 

I implemented change in a way to not reinstantiate writers, added tests for both issues and split two fixes into two different commits. Let me know if that will work for you

APshenkin:
@brianc any updates on this?

brianc (approved):
sorry for the delay - this is perfect now, ty!

Issue #3577, opened 2026-01-14
Latest release dropped support for < node@16


alechirsch:
This commit was recently published in the new version which imports `node:util`, which is not added until node version 16. The supported import should be just `util`.

https://github.com/brianc/node-postgres/commit/27a27547872f7371cfaa2a334fe822b2a3f278f2#r174639831

@brianc

brianc:
thank you. sorry. will fix right now.

brianc:
Strange the tests still run against node 16: https://github.com/brianc/node-postgres/actions/runs/21004704726/job/60383942375

Oh I see what's going on. Yea I'm not supporting versions of node older than 16 at all.  You're on your own with versions older than 16 as node 14 went end of life in 2023. I'll fix this for now & publish a patch version, but please don't expect node@14 compatibility going forward. I recommend you hard-pin your deps in your lockfile and package.json.

alechirsch:
Understood, however the docs do say support for node v4: https://node-postgres.com/guides/upgrading#node-version-support

brianc:
ohhh thanks. I will update that!! lol @ node version 4, right?. node v4. ahhh the good ol days. 🙃

alechirsch:
I would recommend doing a major version bump if you are going to drop support, just my two cents

brianc:
> I would recommend doing a major version bump if you are going to drop support, just my two cents

I hear ya & while I think that's probably technically correct, I do everything in my power not to do semver major unless its significant breaking changes....if someone's node version is so out of date it falls out of the LTS lifetime....I feel like they have bigger problems. I will at least stick to semver minor. I updated the docs to outline my thinking a bit more. Thanks for pointing it out!  PR updated w/ docs. Lets continue discussion there if you have more!

alechirsch:
You are correct about having bigger problems, just trying not to add more 😄 , I would say dropping node support is a significant breaking change, regardless of LTS support

brianc:
hey! fixed in `pg@8.17.1` thanks for the report & the discussion

PR #3578, opened 2026-01-14
Don't use prefix import as it breaks in old nodes.

brianc:
This is a one time fix for #3577 because its easy to do. node < v16 is not supported any more as it went EOL in 2023.


charmander (approved):
Seems like a good practical policy with good advice! 🎉

brianc:
> Seems like a good practical policy with good advice! 🎉

ty for the review!

PR #3595, opened 2026-02-07
chore: upgrade devcontainer at node 22 and add python3 for pg-native

nigrosimone:
pg-native don't works in dev container


brianc:
LGTM! TY!

Issue #724, opened 2015-02-09
Get connection status to avoid wasting clients


rkaw92:
Hi,

Currently, your transaction example checks if the ROLLBACK failed (by looking at the error flag) or if the COMMIT failed (indirectly - the "return client to pool" callback is passed as a callback to the COMMIT query) and returns the client to the pool for re-use iff its state is clean, which is currently defined as lack of error passed to the client release callback.

This is not necessary, and could harm the performance of applications that get a lot of COMMIT failures (serialization failures?). It should be sufficient to check whether the transaction has really ended, and make sure that network errors are all handled properly.

PostgreSQL is supposed to return a transaction status update with each executed query, so our client library should be able to determine whether the transaction has finished. I think this is related to this C interface: http://www.postgresql.org/docs/9.4/static/libpq-status.html - though it may not be very relevant when it comes to a native JS implementation.

Consider the following turn of events, all on a single Client obtained from a pool:

1. BEGIN
2. some_statement_that_succeeds
3. COMMIT **(this fails)**

Now, if the COMMIT has failed, but the transaction has still finished (perhaps because it was turned into a ROLLBACK upon detecting a serialization failure), we can safely return the client to the pool for re-use, because it essentially behaves the same as a successful ROLLBACK. The same should be true for failed ROLLBACKS, provided that the server correctly indicates the transaction status. What I'm getting at is that only clients which are left in the middle of a transaction should be removed from the pool - other clients are just fine to use if I understand the semantics of Postgres.

I've skimmed through this module's codebase and have not found a single variable to look at that would let us determine the transaction state. Is there any? If not, it might be reasonable to keep track of it and export via some public property of the Client.

yoonghm:
I support this great feature. Serializable Snapshot Isolation (SSI) is the only way to make concurrent update to the database safe. 
However, for node.js app, i use new to instantiate a client to postgres instead of from client pool, as i could not find a way to pass the client (from client pool) around. It would not be a problem, unless you use multiple node apps.

Issue #391, opened 2013-07-03
Reset pooled sessions between uses


strk:
When using pooled connections the session configuration in a session will affect operations performed on what the client considers a _new_ connection (but it is not, due to pooling).

For comparison, pg_bouncer default configuration is to issue a "DISCARD ALL" whenever a connection is released into the pool.

This ticket is for supporting a similar mechanism at the node pooled connection level. A "session reset" command could be configurable so that users using pg_bouncer could avoid doing it again.

See https://github.com/Vizzuality/CartoDB-SQL-API/issues/98

sarus:
Just saw this and would love to see some way to "reset" client's session state upon the client being returned to the pool.  For my use case I'm setting `session_replication_role` to `replica` for bulk deletions and inserts and would like it to revert to `origin` (the default) after the client is returned to the pool so the setting does not affect the next query to be called using the same client.

poshest:
I've got a situation where node is calling a postgres function in a loop (a basic job polling mechanism), and that function uses temp tables, which are dropped within the function, but resources for which don't seem to be released until the node server is stopped. 

The node-postgres connections stay open indefinitely and because temp table resources aren't released, eventually the server runs out of memory and falls over.

`DISCARD ALL` [seems to be exactly what we need](http://www.postgresql.org/docs/9.4/static/sql-discard.html): since it "Releases all temporary resources associated with the current session".

poshest:
CORRECTION: `DISCARD ALL` did nothing when we called it manually after each transaction commit. So it's not needed for this use case.

brianc:
If temp tables are kept alive for the length of a connection id recommend
opening & closing a skngle client in your loop instead of using a pooled
client - that way you have control over the lifecycle of everything.

poshest:
Well, that will solve the immediate problem, but isn't a scalable long term solution.  

Is there a way with node-postgres to configure a connection recycle time (TTL)?  Ie some time within X-Y minutes connections will be closed and re-opened?  See, my server is dying after several hours.  If pooled connections had a TTL of between 10 and 20 minutes, they could be recycled in a staggered way that didn't disrupt the application, and problem solved.

brianc:
I think your best solution is to use the client directly and implement your
own pooling behavior. I used the "generic-pool" module. There are likely
other generic pool modules that support TTL behavior.

poshest:
I'm going down that route as we speak! :)  Alternatively there's this
feature request in generic-pool
[...]

PR #3597, opened 2026-02-09
feat: evict connections with open transactions on pool release

panga:
**Summary**

- Add a new pool option `evictOnOpenTransaction` (disabled by default for backwards compatiblity)
- Check `client.getTransactionStatus()` in pg-pool's `release()` method to evict connections not in idle (`I`) state instead of returning them to the pool
- Add a new log indicating that a connection was evicted from the pool due to open transaction
- Add tests covering the new methods and eviction behavior

**Behavior Changes**

- When `evictOnOpenTransaction: true`, a connection is closed and removed from the pool if it is released with an active transaction (BEGIN without COMMIT/ROLLBACK).

**Problem**

When a client with an active transaction (BEGIN without COMMIT/ROLLBACK) is released back to the pool, the connection is returned in a non-idle transaction state. The next consumer that checks out this connection inherits the open transaction, which can cause:

- Row-level lock retention: Rows locked by SELECT ... FOR UPDATE or UPDATE inside the uncommitted transaction remain locked, blocking other connections that try to access the same rows.
- Connection pool exhaustion: In a pool with max: N connections, poisoned connections that hold locks can cause other connections to block waiting on those locks, eventually starving the pool.
- Silent data visibility issues: The next consumer operates inside someone else's transaction with a stale MVCC snapshot, potentially reading stale data or having writes silently rolled back.

This is a known as "leaked/poison connection" problem in connection pools. This feature was recently added in Go pgx https://github.com/jackc/pgx/pull/2481

**Notes**

- This change was also tested with PgBouncer in transaction mode to verify it is working as expected.

### DIFF
--- a/docs/pages/apis/pool.mdx
+++ b/docs/pages/apis/pool.mdx
@@ -63,6 +63,11 @@ type Config = {
   // middleware so that you can rotate the underlying servers. The default is disabled (value of zero).
   maxLifetimeSeconds?: number
 
+  // When true, automatically removes connections with open or failed transactions from the pool
+  // when they are released. This provides protection against "connection poisoning" where
+  // uncommitted transactions leak across different requests. Default is false.
+  evictOnOpenTransaction?: boolean
+
   // Called once when a new client is created, before it is made available to the pool.
   // The client is fully connected and queryable at this point.
   // Can be a regular function or an async function.
@@ -84,6 +89,7 @@ const pool = new Pool({
   idleTimeoutMillis: 30000,
   connectionTimeoutMillis: 2000,
   maxLifetimeSeconds: 60,
+  evictOnOpenTransaction: true,
 })
 ```
 
--- a/packages/pg-pool/index.js
+++ b/packages/pg-pool/index.js
@@ -124,6 +124,10 @@ class Pool extends EventEmitter {
     return this._clients.length > this.options.min
   }
 
+  _hasActiveTransaction(client) {
+    return client && (client._txStatus === 'T' || client._txStatus === 'E')
+  }
+
   _pulseQueue() {
     this.log('pulse queue')
     if (this.ended) {
@@ -394,12 +398,14 @@ class Pool extends EventEmitter {
       this.ending ||
       !client._queryable ||
       client._ending ||
-      client._txStatus === 'T' ||
-      client._txStatus === 'E' ||
+      this._hasActiveTransaction(client) ||
       client._poolUseCount >= this.options.maxUses
     ) {
       if (client._poolUseCount >= this.options.maxUses) {
         this.log('remove expended client')
+      }
+      if (this._hasActiveTransaction(client)) {
+        this.log('remove client with leaked transaction')
       }
 
       return this._remove(client, this._pulseQueue.bind(this))
--- a/packages/pg-pool/index.js
+++ b/packages/pg-pool/index.js
@@ -389,7 +389,15 @@ class Pool extends EventEmitter {
     this.emit('release', err, client)
 
     // TODO(bmc): expose a proper, public interface _queryable and _ending
-    if (err || this.ending || !client._queryable || client._ending || client._poolUseCount >= this.options.maxUses) {
+    if (
+      err ||
+      this.ending ||
+      !client._queryable ||
+      client._ending ||
+      client._txStatus === 'T' ||
+      client._txStatus === 'E' ||
+      client._poolUseCount >= this.options.maxUses
+    ) {
       if (client._poolUseCount >= this.options.maxUses) {
         this.log('remove expended client')
       }
--- a/packages/pg-pool/index.js
+++ b/packages/pg-pool/index.js
@@ -126,7 +126,7 @@ class Pool extends EventEmitter {
   }
 
   _hasActiveTransaction(client) {
-    return client && (client._txStatus === 'T' || client._txStatus === 'E')
+    return client && (client.getTransactionStatus() === 'T' || client.getTransactionStatus() === 'E')
   }
 
   _pulseQueue() {
--- a/packages/pg-pool/index.js
+++ b/packages/pg-pool/index.js
@@ -91,6 +91,7 @@ class Pool extends EventEmitter {
     this.options.maxUses = this.options.maxUses || Infinity
     this.options.allowExitOnIdle = this.options.allowExitOnIdle || false
     this.options.maxLifetimeSeconds = this.options.maxLifetimeSeconds || 0
+    this.options.evictOnOpenTransaction = this.options.evictOnOpenTransaction || false
     this.log = this.options.log || function () {}
     this.Client = this.options.Client || Client || require('pg').Client
     this.Promise = this.options.Promise || global.Promise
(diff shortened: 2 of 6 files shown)


charmander:
Related: #391, #724

How about throwing an error when attempting to return a client with an open transaction to the pool, since that indicates an app bug in current pg, and pg doesn’t have the information to make the optimal choice by itself?

panga:
@charmander thanks for the review. I added a log to align with the behavior in other cases. I couldn’t find a clean way to surface an error at that point since `release` returns `void` and usually called on final. Given the trade-offs around pool correctness, I think it’s preferable to prevent an invalid state rather than silently ignoring it.

Do you think this behavior change should be behind a pool option and disabled by default?

brianc:
> Do you think this behavior change should be behind a pool option and disabled by default?

Though I think how the pool behaves now (doesn't have any concept of the client being in a transaction and will happily rent it out to another caller later) is almost always wrong, I think it risks weird behavioral breaking changes if we change the default behavior in the `8.x` branch....so putting it behind a config where the default value currently is to do what it does now (which is incorrect) is there, but one can opt into auto-aborting & closing clients which are returned to the pool while a transaction is still pending.

So for pg@8.x it would be

```
// not sure what a good config property name is...but something like this:
const pool = new Pool({ evictOnOpenTransaction: true })
```

To opt into the behavior. Then in [9.x](https://github.com/brianc/node-postgres/discussions/3598) we can make this behavior the default & actually remove the feature flag entirely since I don't see any reason to allow the old, invalid behavior long term.

> How about throwing an error when attempting to return a client with an open transaction to the pool, since that indicates an app bug in current pg, and pg doesn’t have the information to make the optimal choice by itself?

I agree w/ @panga that `.release()` seems like a disruptive place to throw an error as its usually in a cleanup phase & historically doesn't throw.  I think one nice thing though would be to actually put `.transactionStatus()` as an actuall method (or getter property) on the client itself and make it public. That would allow libraries or users to perform the check if they wanted to and throw on their side.  What do you think about that?

charmander:
> I couldn’t find a clean way to surface an error at that point since `release` returns `void` and usually called on final.

My suggestion was to throw an error synchronously from `release`. Yes, it’s disruptive if it happens, but that’s the point, since it indicates a serious bug (that would not necessarily even be fixed by discarding the client from the pool). I can’t think of a legitimate reason to return a connection with an open transaction to a pool. But for strict backward compatibility, it could be a deprecation warning in pg 8, with a possible option to opt-in to pg 9 behaviour early.

Exposing the status as a public property is a good idea for other reasons, though.

brianc:
> My suggestion was to throw an error synchronously from `release`. Yes, it’s disruptive if it happens, but that’s the point, since it indicates a serious bug (that would not necessarily even be fixed by discarding the client from the pool). I can’t think of a legitimate reason to return a connection with an open transaction to a pool. But for strict backward compatibility, it could be a deprecation warning in pg 8, with a possible option to opt-in to pg 9 behaviour early.

Yeah that makes sense too. I'll work on a PR soon for the `9.x` branch. I agree it should be considered an error when you return a client that's in a transaction...and closing the users connection is kinda more of a surprise and "silent error" than just throwing as its almost certainly a mistake in the code w/ a dangling transaction.

panga:
I implemented the recommended changes for the new pool option `evictOnOpenTransaction` and for `client.getTransactionStatus()`, including the corresponding documentation updates.

I'm concerned that throwing errors from `release()` may be overly disruptive for many Node.js applications, because in a number of codebases this can surface as an uncaught exception and terminate the process. For example, a common implementation pattern I see looks like this:

```javascript
try {
    return await client.query(...);
} catch (err) {
    client.release(err);
} finally {
    client.release();
}
```

Edit: The above example is not correct, it need to catch the err to avoid double release.
e.g:
```javascript
try {
...
} catch (err) {
  clientErr = err
} finally {
  client.release(clientErr)
}
```

charmander:
> I'm concerned that throwing errors from `release()` may be overly disruptive for many Node.js applications, because in a number of codebases this can surface as an uncaught exception and terminate the process.

Yes, that’s normal JavaScript behaviour for error conditions. Note that this would not be an *uncatchable* exception, and would typically not terminate the process in applications that care about surviving unexpected error conditions.

> For example, a common implementation pattern I see looks like this: […]

That pattern is broken and should not be used, but it’s not incompatible with the general approach of throwing an exception as long as the second `release` is a no-op.

panga:
@charmander you’re right. The simplified example I shared above is not correct because it can result in a double release. In practice, the implementation should explicitly handle the error path and ensure the resource is released exactly once, typically by passing the error object to `release(err)`.

If @brianc is aligned, I can add a throw in this case since it’s behind a feature flag, which should avoid introducing breaking changes for the current version.

panga:
The PR is ready for another review. Could you share what the expectations are, and whether it can go into a minor release now that it’s behind a config flag? I’m a first-time contributor, so I’m just trying to understand the process.

panga:
@charmander I’ve added native support in https://github.com/brianc/node-postgres/pull/3597/commits/b10746f06de3b78ed85b5e66a9d373abc5ad0dda. It depends on https://github.com/brianc/node-libpq/pull/123

brianc:
@panga thanks for doing the work here! and for doing the work on the native side too that's top notch! I hate to be a monster about this...but do you think we could separate this out into 1 PR that just adds the transaction status but doesn't add the optional stuff to the pool and then a different one to add the pool behavior? I'm kinda thinking on the pool behavior one what if instead of having a config option that evicts it we add an `onRelease` callback (I just added `onConnect` [here](https://github.com/brianc/node-postgres/pull/3620)) and it might be more generic to have a callback that's passed the client. By default the callback is undefined but then the user could implement whatever they wanted:


```js
onRelease: async (client) => {
  if (client.getTransactionStatus() !== 'I') {
  console.error("YO YOUR CODE IS DOING SOMETHING BAD. TRANSACTION AUTOMATICALLY ROLLED BACK")
  client.release()
  }
}
```

or 
```js
onRelease: (client) => {
  console.error('transaction pending client released. destroying the universe')
   process.exit(-1)
}
```

etc...

Thoughts? I know its more work...I just want to keep opinions out kinda and extensibility as the focus. Personally I'd probably throw in the release function (and that throw should make the `pool.release` or `client.release` function throw. But everyone might have diff opinions.  That's totally backwards compatible too so don't need a major bump or anything (your change didn't either but i feel like this might be more future-proof?)

panga:
I agree with splitting PRs. This PR expanded beyond the initial scope and is combining multiple changes. Keeping it together can make it easier to review the end-to-end flow and move faster, but I can split out the client transaction function and the pool change into separate PRs.

Creating a new `onRelease` callback is a reasonable direction. I’m not opposed to it, since it provides a clear hook to manage client state before returning it to the pool. That said, to make it broadly useful, it likely needs a corresponding `onCheckout` callback so callers can:

- Initialize session-level variables on checkout and reliably clean them up on release (particularly valuable in multi-tenant scenarios).
- Validate the client state and fail fast by destroying the client (i.e., surfacing an error) when something is incorrect, such as an invalid transaction status.

This would be a net new feature and can be implemented with full backwards compatibility.

If you agree this is the right direction, please confirm. I can start by splitting out the PR for the client transaction status. Will you take ownership of implementing the `onRelease` callback as a follow-up to your current work?

brianc:
> If you agree this is the right direction, please confirm. I can start by splitting out the PR for the client transaction status. Will you take ownership of implementing the `onRelease` callback as a follow-up to your current work?

I can do the on release thing yup.  Actually not sure it can be async though as currently `client.release()` isn't async. That's fine it will still allow to abort, just wont allow a query to be issued. I think I can do `onCheckout` pretty easily as well. 

Yeah direction sounds good!

panga:
@brianc Sorry for the delay. I’ve now split the changes into a new PR that contains only the client-side updates: https://github.com/brianc/node-postgres/pull/3645

brianc:
@panga I'll see your delay and raise you _over a month!_ Sorry about that. I've merged the other change w/ the useful `getTransactionStatus()` - once you rebase this to focus just on the evict changes I think we can revisit the discussion!

panga:
@brianc thank you, don't worry we all get busy lives. I've rebased, the change is much more clean now, let me know what do you think.

PR #3603, opened 2026-02-12
Deprecate Client's internal query queue

brianc:
The internal query queue was a left over from a long ago API decision. In my experience its not used on purpose as nearly everyone is using a flow control library or async/await. It clutters up the internals of the client implementation, makes certain features like pipelining & error handling more difficult, and is a bad & confusing API, so we're removing it in 9.x.

### DIFF
--- a/packages/pg-cursor/test/close.js
+++ b/packages/pg-cursor/test/close.js
@@ -16,11 +16,11 @@ describe('close', function () {
   it('can close a finished cursor without a callback', function (done) {
     const cursor = new Cursor(text)
     this.client.query(cursor)
-    this.client.query('SELECT NOW()', done)
     cursor.read(100, function (err) {
       assert.ifError(err)
       cursor.close()
     })
+    this.client.once('drain', done)
   })
 
   it('can close a finished cursor a promise', function (done) {
@@ -37,11 +37,11 @@ describe('close', function () {
   it('closes cursor early', function (done) {
     const cursor = new Cursor(text)
     this.client.query(cursor)
-    this.client.query('SELECT NOW()', done)
     cursor.read(25, function (err) {
       assert.ifError(err)
       cursor.close()
     })
+    this.client.once('drain', done)
   })
 
   it('works with callback style', function (done) {
--- a/packages/pg-pool/test/lifetime-timeout.js
+++ b/packages/pg-pool/test/lifetime-timeout.js
@@ -12,7 +12,6 @@ describe('lifetime timeout', () => {
     const pool = new Pool({ maxLifetimeSeconds: 1 })
     pool.query('SELECT NOW()')
     pool.on('remove', () => {
-      console.log('expired while idle - on-remove event')
       expect(pool.expiredCount).to.equal(0)
       expect(pool.totalCount).to.equal(0)
       done()
@@ -22,7 +21,6 @@ describe('lifetime timeout', () => {
     const pool = new Pool({ maxLifetimeSeconds: 1 })
     pool.query('SELECT pg_sleep(1.4)')
     pool.on('remove', () => {
-      console.log('expired while busy - on-remove event')
       expect(pool.expiredCount).to.equal(0)
       expect(pool.totalCount).to.equal(0)
       done()
--- a/packages/pg-protocol/src/buffer-reader.ts
+++ b/packages/pg-protocol/src/buffer-reader.ts
@@ -1,7 +1,5 @@
-const emptyBuffer = Buffer.allocUnsafe(0)
-
 export class BufferReader {
-  private buffer: Buffer = emptyBuffer
+  private buffer: Buffer = Buffer.allocUnsafe(0)
 
   // TODO(bmc): support non-utf8 encoding?
   private encoding: string = 'utf-8'
--- a/packages/pg/lib/client.js
+++ b/packages/pg/lib/client.js
@@ -1,5 +1,3 @@
-'use strict'
-
 const EventEmitter = require('events').EventEmitter
 const utils = require('./utils')
 const nodeUtils = require('util')
@@ -14,23 +12,28 @@ const crypto = require('./crypto/utils')
 
 const activeQueryDeprecationNotice = nodeUtils.deprecate(
   () => {},
-  'Client.activeQuery is deprecated and will be removed in a future version.'
+  'Client.activeQuery is deprecated and will be removed in pg@9.0'
 )
 
 const queryQueueDeprecationNotice = nodeUtils.deprecate(
   () => {},
-  'Client.queryQueue is deprecated and will be removed in a future version.'
+  'Client.queryQueue is deprecated and will be removed in pg@9.0.'
 )
 
 const pgPassDeprecationNotice = nodeUtils.deprecate(
   () => {},
-  'pgpass support is deprecated and will be removed in a future version. ' +
+  'pgpass support is deprecated and will be removed in pg@9.0. ' +
     'You can provide an async function as the password property to the Client/Pool constructor that returns a password instead. Within this funciton you can call the pgpass module in your own code.'
 )
 
 const byoPromiseDeprecationNotice = nodeUtils.deprecate(
   () => {},
-  'Passing a custom Promise implementation to the Client/Pool constructor is deprecated and will be removed in a future version.'
+  'Passing a custom Promise implementation to the Client/Pool constructor is deprecated and will be removed in pg@9.0.'
+)
+
+const queryQueueLengthDeprecationNotice = nodeUtils.deprecate(
+  () => {},
+  'Calling client.query() when the client is already executing a query is deprecated and will be removed in pg@9.0. Use asycn/await or an external async flow control mechanism instead.'
 )
 
 class Client extends EventEmitter {
@@ -680,6 +683,9 @@ class Client extends EventEmitter {
       return result
     }
 
+    if (this._queryQueue.length > 0) {
+      queryQueueLengthDeprecationNotice()
+    }
     this._queryQueue.push(query)
     this._pulseQueryQueue()
     return result
--- a/packages/pg/lib/native/client.js
+++ b/packages/pg/lib/native/client.js
@@ -1,5 +1,4 @@
-'use strict'
-
+const nodeUtils = require('util')
 // eslint-disable-next-line
 var Native
 // eslint-disable-next-line no-useless-catch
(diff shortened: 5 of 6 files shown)


stephenh:
@brianc naive question, but isn't this going to be break `Promise.all` usages like:

```
await Promise.all([
  client.query(...),
  client.query(...),
});
```

...I guess point is "those were never actually parallel anyway".

But then if I have a `GET /authors` endpoint that is doing ~some async work, and do a `request.ctx = { client: pool.connect() }` (all SQL calls on this request go through this connection, i.e. maybe its a transaction), I might have multiple methods in my `GET /authors` want to `select * from books` / `select * from tags`, in very disparate places in the `GET /authors` code, but all "sharing the connection", and now things will break?

I'd have to build some sort of "anyone that wants to execute a connection on this client needs to gate/serialize access through this `SharedClient`" -- which I guess in retrospect is exactly what `PoolClient` has been doing for us :sweat_smile: .
 
Fwiw when I heard murmurs of "removing the internal query queue", I was like "oh that is probably something really old / only for callback-based users" ... but now that I look a little closer, it seems like Promise users that want to "get a connection & have it shared by multiple routines/async methods on the request" will trip over this as well?

...that said, if this speeds up landing of the `pipeline` PR, that effectively means we go back to "just drop queries on the client" (with no concern for whether an existing query is in-flight), then maybe it's all good, b/c I'm looking forward to the pipeline support!

dchekanov:
I ran into the exact issue described by @stephenh:

```javascript
function operation(item, tx) {
  return tx.client.query('...');
}

const tx = {client};

// this produces the deprecation message
// the order of execution does not matter, it just needs to perform the same operation on each item
// Promise.all is faster than for ... of  when operation includes non-blocking async calls (FS, API)
await Promise.all(items.map(item => operation(item, tx)));
```

I ended up adding a mutex to each transaction:

```javascript
import {Mutex} from 'async-mutex';

function operation(item, tx) {
  return tx.mutex.runExclusive(() => tx.client.query('...'));
}

const tx = {client, mutex: new Mutex()};

await Promise.all(items.map(item => operation(item, tx)));
``` 

It's probably worth mentioning this case in the migration guide.

brianc:
I really appreicate the comments and insights here y'all!  I'll try and address them from my point of view. I don't want to sound like I'm arguging - I'm often wrong on things and if this ends up being a huge grenade that screws over tons of people I can roll it back..but I don't think it will be that way.  The main reason is the "queryQueue" is just doing async management "secretly" internally which I think is actually kind of a confusing API. I think if you are sending multiple queries with `Promise.all` you should either use `pool.query` (which does queueing and all that stuff, and it totally should!) or use some util function or library to do an async style like `mapAsync`. 

> I might have multiple methods in my `GET /authors` want to `select * from books` / `select * from tags`, in very disparate places in the `GET /authors` code, but all "sharing the connection", and now things will break?

Absolutely things wont break...unless you are doing `Promise.all` on the various subroutines that run those queries or not awaiting the promise result of running a query. I would probably just caution against that in general.  If you need a single client for transactions it's probably also important your queries run in order. You can always send multiple queries on a single client, it's just no longer allowed to queue up multiple queries

One thing to add tho I rarely connect a single client. Even if I only need 1 I use a pool (and optionally set `max: 1` on the pool options). 



> ...that said, if this speeds up landing of the `pipeline` PR, that effectively means we go back to "just drop queries on the client" (with no concern for whether an existing query is in-flight), then maybe it's all good, b/c I'm looking forward to the pipeline support!

totally! Yeah that's the thinking too - much clearer to have a "you can send queries while others are in flight while pipelining. If you try otherwise, that's now how things are actually working under the hood and you might be misunderstanding what's happening internally."

stephenh:
> run statements in order

Right; for us, we generally run "a bunch of `SELECT`s to load data" in ~adhoc/parallel fashion (so likely to hit this Promise.all footgun), but then make sure our INSERTs/UPDATEs/DELETEs are all done by a single "flush" that orders them correctly.

(...okay, technically we nearly always run our `SELECT`s outside of a txn, so that it's "more web scale" :sweat_smile: , but I'm thinking of prior systems where we purposefully ran `SELECT`s within the txn to get the dbs transaction isolation benefits.)

> doing async management internally / secretly

Yeah; fwiw, I was thinking "I've never used a database API where I had to do 'my own queue management'" -- but I realized those other APIs are all OG Java/C# blocking APIs, which I guess ... b/c they block, they probably don't have their own internal query queues either, and I was just never truly using a `client` in parallel before. :thinking: 

(Within the node ecosystem, I've done prototypes with postgres.js, solely to get its pipeline support, and it allows "multiple queries at once", albeit it is likely just pipelining right onto the wire anyway? :thinking:)

> speeds up landing the pipeline PR

My naive assertion is that shipping "the query queue is actually gone" in 9.x is most safe/realistic (in terms of avoiding breaking changes) if 9.x also has pipelining shipped & _enabled by default_, so that users would experience the same API semantics...

(AFAIU the db semantics would be the same as well--which ever query arrives first, is sent to pg first, is executed by pg first -- there is just no node-pg internal queuing.)

Happy to be wrong about that!

stephenh:
I'm seeing the warnings this morning :sweat_smile: and so ran the `--deprecation-warning` to see where I should go fix things:

image

Turns out the `Promise.all` is happening within the 3rd party pg-structure library we use:

https://github.com/ozum/pg-structure/blob/master/src/main.ts#L348

Where the code is, as expected, "using a `client` like a pool" and issuing mass/parallel SELECTs.

(Fwiw I'm 100% in favor of internally simplifying the node-pg codebase, and love @brianc that you're still here grinding out improvements for us, just providing data points -- ty!)

dchekanov:
Hi @brianc, thanks for your comment! 

> I think if you are sending multiple queries with Promise.all you should either use pool.query (which does queueing and all that stuff, and it totally should!) or use some util function or library to do an async style like mapAsync.

I just want to highlight something so that our concern about the deprecation is more clear: 

1. Using the same client might be necessary because operations must be performed within the same transaction.
2. Operations might be complex enough to benefit from parallel async execution.

Here's an illustration:

```javascript

await client.query('BEGIN');

await Promise.all(
  [1,2,3].map(i => {
    return fsp.readFile(`${i.txt}`)
      .then(buffer => api.parse(buffer))
      .then(parsed => client.query('INSERT INTO files (id, data) VALUES ($1, $2)', [i, parsed]));
  })
);

await client.query('COMMIT');

client.release();
```

In this simplified example you can easily extract the FS+API part to run in parallel and then INSERT the results sequentially. However, in real-world applications, the processing logic might be far more convoluted and harder to refactor.

I’m not saying the deprecation is a bad idea - the rationale makes sense - but users will need to carefully review their code when this becomes a breaking change and the transition might not be straightforward.

Thanks a lot for your work!

charmander:
> My naive assertion is that shipping "the query queue is actually gone" in 9.x is most safe/realistic (in terms of avoiding breaking changes) if 9.x also has pipelining shipped & _enabled by default_, so that users would experience the same API semantics...

Yes, I would expect this to be the case.

> Right; for us, we generally run "a bunch of SELECTs to load data" in ~adhoc/parallel fashion (so likely to hit this Promise.all footgun), but then make sure our INSERTs/UPDATEs/DELETEs are all done by a single "flush" that orders them correctly.

> Promise.all is faster than for ... of  when operation includes non-blocking async calls (FS, API)

Note that there was already a footgun: the query queue is an array that gets `shift`ed from, instead of a proper queue. With enough queries, dequeuing becomes a bottleneck.

caseywebdev:
I hear your rationale, makes sense, though I'll need to find a workaround for the same behavior. Is there a way through public APIs on the client to replicate the current behavior if `_queryQueue` goes away? Another rub here is that we're using knex.js on top of pg and I don't believe they're exposing the raw connection object of the driver that's being used.

We have many scenarios where we batch queries running on the same transaction. The order does not matter, and the code running these queries does not know/care about the other queries running on the same transaction, they just need to all succeed or be rolled back if any of them fail. Currently this works well with the query queue, but we'd need to figure out a way to replicate it if it goes away.

Ideally I would be able to check if a client is currently executing and if it is, be able to await a promise for the current execution before attempting to run the next query.

brianc:
I really appreicate all the discussion here. I haven't actually done the work to remove the queue yet so lemme ruminate on it a bit & see if there are easy solutions. I don't want to introduce a change that makes upgrading painful or error prone for others with the main benifit being a more simplified internals and less surprising external api.  I deprecated it as early as I could to let it shake out & see what feedback was. Not fully committed to removing it yet if it causes too much headache.

pasieronen:
I'm worried that there might be quite a lot of existing code that assumes you can send new queries while previous ones are still in progress. (And since the function signature does not change, TypeScript won't help in finding the code that needs changes.)

For example, if you take the example from https://github.com/brianc/node-postgres/blob/master/docs/pages/features/transactions.mdx and make a small, harmless-looking change:

```
try {
  await client.query('BEGIN')
  const [res, something] = await Promise.all([
    client.query(...)
    doSomethingElse() // <-- something async, but does not make any database queries
  ])
  // …
  await client.query('COMMIT')
} catch (e) {
  throw e
} finally {
  client.release()
}
```

There's now a subtle race condition here: doSomethingElse() could fail before client.query completes; if that happens (presumably rarely), then client.query('ROLLBACK') would be called before the first query has completed... 

Myself, I'm using node-postgres via the pg-promise wrapper, and I have literally 100+ places in my code that use tx.batch(), like shown in this example from pg-promise docs (https://github.com/vitaly-t/pg-promise?tab=readme-ov-file#transactions)

```
db.tx(t => {
    // creating a sequence of transaction queries:
    const q1 = t.none('UPDATE users SET active = $1 WHERE id = $2', [true, 123]);
    const q2 = t.one('INSERT INTO audit(entity, id) VALUES($1, $2) RETURNING id', ['users', 123]);

    // returning a promise that determines a successful transaction:
    return t.batch([q1, q2]); // all of the queries are to be resolved;
})
```

brianc:
yah i thought about this a lot while i was on vacation...I think I might roll back the deprecation  & leave the queryQueue existing as it currently is. It's a feature since v1.0 and doesn't actually hurt much other than a less clear and kinda "too clever" api surface. But it also is probably widely used and disruptive to remove and a low maintenance cost on me. I can do some things wrt refactoring to make the internals a bit more clear.  I really hate breaking changes & the cost of this one might be bigger than the benefit. 

I'll work on all the other 9.0 features & changes and leave this until the end and continue to think about it. But i'm leaning towards reverting this PR.

stephenh:
Awesome! Appreciate all your work @brianc , ty!

charmander:
- no deprecation
- replace query queue with automatic pipelining in pg 9

seems reasonable, yeah? A changelog entry will cover it, and almost all uses should keep working.

SirBernardPhilip:
Hey @brianc! Is reverting the deprecation still the plan?

brianc:
> Hey @brianc! Is reverting the deprecation still the plan?

Yeah I am going to revert this warning next week & do this in a less disruptive way!

PR #3605, opened 2026-02-23
Typo fix in PgPass deprecation (funciton)

bigpresh:
Just a really simple typo fix in a deprecation warning:

```
(node:1065439) DeprecationWarning: pgpass support is deprecated and will be removed in a future version. You can provide an async function as the password property to the Client/Pool constructor that returns a password instead. Within this funciton you can call the pgpass module in your own code.
```

(Note `funciton`)


brianc:
Thanks!

PR #3609, opened 2026-02-24
docs: `keepAlive` and `keepAliveInitialDelayMillis`

charmander:
https://github.com/brianc/node-postgres-docs/pull/173


brianc:
Ah ty!

charmander:
@brianc :+1:

This was the last open PR or issue on node-postgres-docs, so you can safely archive it now (to stop the security alerts and Dependabot, for example).

PR #3623, opened 2026-03-05
Don't encourage people to use `pool.on('connect', ...)` for setup anymore

danieldiekmeier:
Following the changes in #3620, and with the probable upcoming removal of the internal query queue, I think the docs should already point users to `onConnect` and explain why.


brianc:
nice! Yeah I don't think that event listener is very valuable.

PR #3643, opened 2026-03-10
feat: Add multihost support for native js driver

maxbronnikov10:
### Summary
- Add host/port array support to Connection.connect() in the pure JS driver, with automatic TCP-level failover to the next host on connection error
- Add targetSessionAttrs config option to Connection and wire it through Client
- Intercept protocol events during startup to collect in_hot_standby and default_transaction_read_only from ParameterStatus messages and validate them against targetSessionAttrs before signalling readyForQuery to the caller
- Fall back to issuing SHOW transaction_read_only; SELECT pg_catalog.pg_is_in_recovery() for older servers that don't advertise those params

### Behavior Changes
When host and port are arrays, the driver tries each host in order. TCP-level errors before the first successful connect silently trigger a retry on the next host. With targetSessionAttrs set, hosts that connect successfully but don't match the session requirement are also silently skipped. If the list is exhausted without a match, an error is emitted.

### Problem
The pure JS driver previously only accepted a single host and had no targetSessionAttrs support, unlike the native libpq driver. This made it impossible to use driver for high-availability Postgres setups where you want to connect to whichever replica is currently the primary, or explicitly route to a read replica.

The lack of multihost support meant users had to implement their own failover logic outside the driver, or switch to the native bindings just to get this feature. 

### Related issues
https://github.com/brianc/node-postgres/issues/1470

### DIFF
--- a/packages/pg/lib/client.js
+++ b/packages/pg/lib/client.js
@@ -81,6 +81,8 @@ class Client extends EventEmitter {
         keepAlive: c.keepAlive || false,
         keepAliveInitialDelayMillis: c.keepAliveInitialDelayMillis || 0,
         encoding: this.connectionParameters.client_encoding || 'utf8',
+        targetSessionAttrs: c.targetSessionAttrs || this.connectionParameters.targetSessionAttrs || null,
+        trustParameterStatus: c.trustParameterStatus || false,
       })
     this._queryQueue = []
     this.binary = c.binary || defaults.binary
@@ -155,7 +157,7 @@ class Client extends EventEmitter {
       }
     }
 
-    if (this.host && this.host.indexOf('/') === 0) {
+    if (!Array.isArray(this.host) && this.host && this.host.indexOf('/') === 0) {
       con.connect(this.host + '/.s.PGSQL.' + this.port)
     } else {
       con.connect(this.port, this.host)
@@ -542,7 +544,7 @@ class Client extends EventEmitter {
     if (client.activeQuery === query) {
       const con = this.connection
 
-      if (this.host && this.host.indexOf('/') === 0) {
+      if (!Array.isArray(this.host) && this.host && this.host.indexOf('/') === 0) {
         con.connect(this.host + '/.s.PGSQL.' + this.port)
       } else {
         con.connect(this.port, this.host)
--- a/packages/pg/lib/client.js
+++ b/packages/pg/lib/client.js
@@ -8,6 +8,7 @@ const ConnectionParameters = require('./connection-parameters')
 const Query = require('./query')
 const defaults = require('./defaults')
 const Connection = require('./connection')
+const MultiConnection = require('./multi-connection')
 const crypto = require('./crypto/utils')
 
 const activeQueryDeprecationNotice = nodeUtils.deprecate(
@@ -86,16 +87,23 @@ class Client extends EventEmitter {
 
     this.enableChannelBinding = Boolean(c.enableChannelBinding) // set true to use SCRAM-SHA-256-PLUS when offered
     this.scramMaxIterations = coerceNumberOrDefault(c.scramMaxIterations, sasl.DEFAULT_MAX_SCRAM_ITERATIONS)
+    const targetSessionAttrs = c.targetSessionAttrs || this.connectionParameters.targetSessionAttrs || null
+    const connectionConfig = {
+      stream: c.stream,
+      ssl: this.connectionParameters.ssl,
+      sslNegotiation: this.connectionParameters.sslnegotiation,
+      keepAlive: c.keepAlive || false,
+      keepAliveInitialDelayMillis: c.keepAliveInitialDelayMillis || 0,
+      encoding: this.connectionParameters.client_encoding || 'utf8',
+      targetSessionAttrs: targetSessionAttrs,
+    }
+    const needsMultiConnection =
+      Array.isArray(this.host) ||
+      Array.isArray(this.port) ||
+      Boolean(targetSessionAttrs && targetSessionAttrs !== 'any')
+
     this.connection =
-      c.connection ||
-      new Connection({
-        stream: c.stream,
-        ssl: this.connectionParameters.ssl,
-        sslNegotiation: this.connectionParameters.sslnegotiation,
-        keepAlive: c.keepAlive || false,
-        keepAliveInitialDelayMillis: c.keepAliveInitialDelayMillis || 0,
-        encoding: this.connectionParameters.client_encoding || 'utf8',
-      })
+      c.connection || (needsMultiConnection ? new MultiConnection(connectionConfig) : new Connection(connectionConfig))
     this._queryQueue = []
     this.binary = c.binary || defaults.binary
     this.processID = null
@@ -170,7 +178,9 @@ class Client extends EventEmitter {
       }
     }
 
-    if (this.host && this.host.indexOf('/') === 0) {
+    if (con instanceof MultiConnection || Array.isArray(this.host)) {
+      con.connect(this.port, this.host)
+    } else if (this.host && this.host.indexOf('/') === 0) {
       con.connect(this.host + '/.s.PGSQL.' + this.port)
     } else {
       con.connect(this.port, this.host)
@@ -566,7 +576,7 @@ class Client extends EventEmitter {
     if (client.activeQuery === query) {
       const con = this.connection
 
-      if (this.host && this.host.indexOf('/') === 0) {
+      if (!Array.isArray(this.host) && this.host && this.host.indexOf('/') === 0) {
         con.connect(this.host + '/.s.PGSQL.' + this.port)
       } else {
         con.connect(this.port, this.host)
--- a/packages/pg/lib/connection.js
+++ b/packages/pg/lib/connection.js
@@ -1,15 +1,25 @@
 'use strict'
 
+const net = require('net')
 const EventEmitter = require('events').EventEmitter
-
 const { parse, serialize } = require('pg-protocol')
 const { getStream, getSecureStream } = require('./stream')
+const MultiHost = require('./multihost')
 
 const flushBuffer = serialize.flush()
 const syncBuffer = serialize.sync()
 const endBuffer = serialize.end()
 
 // TODO(bmc) support binary mode at some point
+const PROBE_SHOW_TX_READ_ONLY = serialize.query('SHOW transaction_read_only')
+const PROBE_SELECT_RECOVERY = serialize.query('SELECT pg_catalog.pg_is_in_recovery()')
+
+const PHASE = {
+  STARTUP: 'startup',
+  PROBE: 'probe',
+  DONE: 'done',
+}
+
 class Connection extends EventEmitter {
   constructor(config) {
     super()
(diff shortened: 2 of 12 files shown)


packages/pg/lib/connection.js
```diff
@@ -34,76 +38,200 @@ class Connection extends EventEmitter {
     })
   }
 
+  _newStream() {
+    return this._streamFactory ? this._streamFactory({}) : getStream(this.ssl)
+  }
```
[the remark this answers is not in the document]
maxbronnikov10: fixed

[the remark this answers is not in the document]
maxbronnikov10: test name - stream factory receives same config on failover streams

maxbronnikov10:
@brianc @charmander hello! can u please check this PR

packages/pg/lib/connection.js
```diff
@@ -34,78 +46,249 @@ class Connection extends EventEmitter {
...
+    }
 
-    const reportStreamError = function (error) {
-      // errors about disconnections should be ignored during disconnect
-      if (self._ending && (error.code === 'ECONNRESET' || error.code === 'EPIPE')) {
-        return
+    const attemptConnect = () => {
+      const currentHost = hosts[hostIndex]
+      const currentPort = ports[Math.min(hostIndex, ports.length - 1)]
+      let connected = false
```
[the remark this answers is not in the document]
maxbronnikov10: fixed in connection-parameters.js

packages/pg/lib/connection.js
```diff
@@ -216,6 +395,66 @@ class Connection extends EventEmitter {
...
+    case 'read-write':
+    case 'read-only':
+      return 'tx_read_only'
+    case 'primary':
+    case 'standby':
+    case 'prefer-standby':
+      return 'is_in_recovery'
+    default:
+      return null
+  }
```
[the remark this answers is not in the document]
maxbronnikov10: fixed in connection-parameters.js

maxbronnikov10:
@brianc @charmander hello! can u please check this PR

packages/pg/lib/connection.js
```diff
@@ -1,23 +1,36 @@
...
 const EventEmitter = require('events').EventEmitter
-
 const { parse, serialize } = require('pg-protocol')
 const { getStream, getSecureStream } = require('./stream')
 
 const flushBuffer = serialize.flush()
 const syncBuffer = serialize.sync()
 const endBuffer = serialize.end()
 
-// TODO(bmc) support binary mode at some point
```
charmander, line 12: Please don’t remove comments unrelated to the pull request, regardless of the quality of the comments. (More instances later in the file.)

charmander (commented):
I’m not sure if/when I’ll get the chance to give this a thorough review, but the changes to `Connection` look kind of messy. Is there a reason the logic can’t be something that operates *on* `Connection`s?

packages/pg/lib/connection.js
maxbronnikov10: okay, i will return the comments.

what exactly looks confusing? It's so hard to determine what needs to be improved. I don't see any point in taking this out of Connection, it looks like we'll have to refine the pg-pool and then the logic will diverge from Connection, which will have to be refined in two places.

maxbronnikov10: UPD: i refactored implementation in more simpler approach - less nested conditions and functions

maxbronnikov10: also @brianc need your opinion about approach

maxbronnikov10:
@brianc hello! can u please check this PR

brianc:
> can u please check this PR

will do now. sorry for delay 🙏

charmander:
> what exactly looks confusing?

It introduces a bunch of state on `Connection` to implement a feature that, as far as I’m aware, is a process that logically attempts multiple unrelated `Connection` instances.

> I don't see any point in taking this out of Connection, it looks like we'll have to refine the pg-pool and then the logic will diverge from Connection, which will have to be refined in two places.

Are you thinking of `Client`?

brianc:
It doesn't look like vibe code which is good! Looks like you put a lot of thought into this which I appreciate. I think I share some of the concerns @charmander has about it being a lot of additional stuff in connection. I'm kinda okay with it though at first glance. I'm pulling down the branch to open it in my editor to kinda "feel it" more. From _my_ perspective what I'd like to see...though I'm really not sure of the feasibility of it...is there any way to have any _integration tests_ for this? Where maybe you try to connect to several hosts, only one of them being valid in the tests? But actually connect to a backend? I would love it if we could have a docker compose setup w/ multiple postgres's that have failover and readonly settings and all that but that's probably a bit out of scope and wouldn't work great w/ local testing (unless one were to run docker for local testing).

maxbronnikov10:
I believe that logic in `Client` would create a messier situation. Client operates at a higher level (queries, transactions), while the host iteration, temporary event listeners, and targetSessionAttrs checks are all low-level transport concerns. If I tried to hoist that into Client, I'd have to manually detach and reattach EventEmitter listeners between connection attempts, coordinate stream cleanup, and expose protocol-level details that Client really shouldn't need to know about. In my view, that coupling would be more fragile and leaky than keeping the complexity sealed inside Connection, where it's already responsible for managing the transport lifecycle.

On integration tests — I completely agree they'd be valuable, and I'm keen to dig into that. To do it properly we'd need to set up CI with a multi-node environment (something like Patroni with a primary and read replicas), which would take a bit of infrastructure work. I'm absolutely willing to spend time on that, but I'm conscious it could become a significant addition and risk blocking this PR. Would you be comfortable if I tackled that as a follow-up PR, so we don't inflate the scope and complexity here?

brianc:
> On integration tests — I completely agree they'd be valuable, and I'm keen to dig into that. To do it properly we'd need to set up CI with a multi-node environment (something like Patroni with a primary and read replicas), which would take a bit of infrastructure work. I'm absolutely willing to spend time on that, but I'm conscious it could become a significant addition and risk blocking this PR. Would you be comfortable if I tackled that as a follow-up PR, so we don't inflate the scope and complexity here?

yeah for sure. there are _some_ tests covering it already, and all existing tests pass which are big first steps. I am still churning through my backlog of PRs and issues today, returning to computer, so I might not get to do a comprehensive dive until tomorrow. If you don't hear back sometime in the afternoon USA time please ping me at brian.m.carlson@gmail.com or get on [the discord](https://discord.gg/2afXp5vUWm) and ping me directly there. My inbox is an absolute nightmare of github notification emails and things can get lost in the shuffle, but I want to bring this over the finish line one way or another as its pretty big & I don't want to break it w/ merge conflicts.

maxbronnikov10:
@brianc hello! can u please check this PR

vanchenkin:
@brianc Hi! We would be happy if you could merge this PR.

packages/pg/lib/multihost.js
```diff
@@ -0,0 +1,68 @@
+'use strict'
+
+class MultiHost {
```
charmander: This class doesn’t represent something that’s instantiated and should just be top-level module exports.

packages/pg/test/unit/client/multihost-tests.js
```diff
@@ -0,0 +1,80 @@
...
+  const con = makeFakeConnection()
+  con.connect = function (path) {
+    con.connectCalls.push({ path })
+  }
+  const client = new Client({ connection: con, host: '/tmp/', port: 5432 })
+  client._connect(function () {})
+  assert.ok(con.connectCalls[0].path.startsWith('/tmp/'), 'should use domain socket path')
+})
+
+// --- array host does NOT trigger domain socket path ---
```
charmander: Shouldn’t it? I don’t see any reason to exclude Unix sockets from this. (Terminology note for this file, by the way: they’re “Unix domain sockets” as in “sockets in the Unix domain”, not “domain sockets”. They can be shortened to “Unix sockets”.)

packages/pg/lib/client.js
```diff
@@ -81,6 +81,8 @@ class Client extends EventEmitter {
         keepAlive: c.keepAlive || false,
         keepAliveInitialDelayMillis: c.keepAliveInitialDelayMillis || 0,
         encoding: this.connectionParameters.client_encoding || 'utf8',
+        targetSessionAttrs: c.targetSessionAttrs || this.connectionParameters.targetSessionAttrs || null,
+        trustParameterStatus: c.trustParameterStatus || false,
```
charmander: What use case is this undocumented option’s `false` value needed to support? Does libpq have something similar?

charmander (changes requested):
I still haven’t done an in-depth review, but remain fairly confident that implementing this inside `Connection` creates an unnecessary mess that will harm maintainability. Maybe a new `MultiConnection` class that creates `Connection` instances would help (but I have a suspicion that a single async function with a loop might be able to do the trick).

packages/pg/lib/client.js
maxbronnikov10: Option for support PG multihost versions less than 14. I do not find that node-postgres is not supported "some" versions.

maxbronnikov10:
How would this work with the existing Connection class?

Multihost is still one logical connection: try hosts sequentially, run the same startup lifecycle, validate targetSessionAttrs, and emit readyForQuery only for the selected host.

A separate MultiConnection would still need to coordinate or proxy the full Connection lifecycle, so I do not yet see how it reduces complexity. I am open to a concrete alternative, but an unspecified refactoring would block this functionality without a clear maintainability benefit.

maxbronnikov10:
We can refactor this, but separating it into MultiConnection would also require changing Client and the existing event lifecycle I think. That risks breaking compatibility with custom connections and other code relying on the current Connection interface, or adding extra boilerplate for library users.

maxbronnikov10:
@brianc @charmander  hello! can u please check new version of featuru - i separated multi host logic from connection class

maxbronnikov10:
Hi! Ping

charmander:
Thank you for giving that refactoring a try! I’m putting together a patch to suggest on top of it, and hope to have it ready within a few more days.

maxbronnikov10:
@charmander Sure, I’ll wait. I’m also happy to make any further changes if needed.

maxbronnikov10:
@charmander Hi! Do u need any help?

maxbronnikov10:
Ping

maxbronnikov10:
@brianc @charmander Hi, guys! Just pinging for check PR. It is been waiting already 5 months. After refactoring I don't understand exactly what needs to be done to accept the PR.

maxbronnikov10:
Hi! Ping

maxbronnikov10:
Ping

maxbronnikov10:
Ping

maxbronnikov10:
Ping

maxbronnikov10:
@brianc @charmander If I add integration tests, will it increase the chances of merging?

charmander:
Hi, sorry I haven’t had the time to work on a refactoring lately. I really do recommend a cleaner implementation though, and (as just one potential reviewer, mind) I’m not going to review the logic without a cleaner implementation.

But yes, integration tests would be valuable/required, and should be safe to implement since they’re independent of the feature implementation.

maxbronnikov10:
> Hi, sorry I haven’t had the time to work on a refactoring lately. I really do recommend a cleaner implementation though, and (as just one potential reviewer, mind) I’m not going to review the logic without a cleaner implementation.
> 
> But yes, integration tests would be valuable/required, and should be safe to implement since they’re independent of the feature implementation.

@charmander Hi! just pushed a refactor — replaced MultiConnection with an async loop and added integration tests against real Postgres. Could you take another look and let me know if this is closer to what you had in mind?

maxbronnikov10:
ping

maxbronnikov10:
Ping

maxbronnikov10:
@brianc @charmander ping

PR #3596, opened 2026-02-07
feat(pg-native): pipeline mode

nigrosimone:
same as https://github.com/brianc/node-postgres/pull/3591 but for pg-native (complete https://github.com/brianc/node-postgres/pull/3592)

```js
const Client = require('pg-native')

const client = new Client({
  pipelineMode: true,
  pipelineMaxQueries: 100  // default: 1000
})

client.connectSync() // or: client.connect(callback)

client.query('SELECT $1::int as num', [1], (err, rows) => {
  console.log(rows)
}) // or: await client.query('SELECT 1 as num')
```

# Benchmark
```
node index.js
============================================================
pg-native Benchmark
============================================================
Pipeline mode supported: true

Warming up...
Warmup complete.

--- Iteration 1/3 ---

Simple query (SELECT 1):
  pg.native:         5263 qps (26314 queries in 5000ms)
  Native:            6858 qps (34290 queries in 5000ms)
  Native+Pipeline:   5866 qps (29328 queries in 5000ms)

Complex query (generate_series):
  pg.native:         1074 qps (5368 queries in 5000ms)
  Native:            1035 qps (5174 queries in 5000ms)
  Native+Pipeline:   1117 qps (5586 queries in 5000ms)

Concurrent queries (10 queries in parallel):
  pg.native:         6873 qps (34370 queries in 5001ms)
  Native:            N/A (concurrent queries not supported without pipeline)
  Native+Pipeline:   20363 qps (101820 queries in 5000ms)

--- Iteration 2/3 ---

Simple query (SELECT 1):
  pg.native:         6538 qps (32693 queries in 5000ms)
  Native:            6990 qps (34952 queries in 5000ms)
  Native+Pipeline:   6255 qps (31277 queries in 5000ms)

Complex query (generate_series):
  pg.native:         928 qps (4642 queries in 5001ms)
  Native:            771 qps (3858 queries in 5001ms)
  Native+Pipeline:   687 qps (3435 queries in 5000ms)

Concurrent queries (10 queries in parallel):
  pg.native:         6582 qps (32920 queries in 5001ms)
  Native:            N/A (concurrent queries not supported without pipeline)
  Native+Pipeline:   19832 qps (99160 queries in 5000ms)

--- Iteration 3/3 ---

Simple query (SELECT 1):
  pg.native:         6560 qps (32799 queries in 5000ms)
  Native:            6891 qps (34458 queries in 5000ms)
  Native+Pipeline:   6447 qps (32234 queries in 5000ms)

[...]

PR #3652, opened 2026-04-10
Add opt-in query pipelining

mcollina:
Adds opt-in query pipelining via `client.pipelining = true` (or `new Pool({ pipelining: true })`).

Pipelined queries are sent to the server without waiting for previous responses, using PostgreSQL's extended query protocol with per-query `Sync` for error isolation. ~2-3x throughput on batches of simple queries locally.

- `pg` — three-queue state machine (`_queryQueue` → `_sentQueryQueue` → `_activeQuery`), named-statement dedup across in-flight Parse, graceful `end()`, `query_timeout` unblocks the pipeline
- `pg-pool` — `pipelining` option sets the flag on every client the pool creates
- Unit + integration tests, benchmark, and docs (features + API reference)


nigrosimone:
Hi @mcollina , take a look at these two PRs who proposed pipeline mode, maybe they can help you:
- https://github.com/brianc/node-postgres/pull/3596
- https://github.com/brianc/node-postgres/pull/3591

mcollina:
@nigrosimone do you know why they are stuck?

cc @brianc how do you want to proceed?

nigrosimone:
> @nigrosimone do you know why they are stuck?

Maybe they want to remove the query queue in v. 9. My implementation is totally based on query queue.

cesco69:
A test fails with: 
> PQsendQuery not allowed in pipeline mode

 from [docs](https://www.postgresql.org/docs/current/libpq-pipeline-mode.html#LIBPQ-PIPELINE-USING)
> PQsendQuery is also disallowed, because it uses the simple query protocol

In pipeline mode always use `PQsendQueryParams` (also for query without parameters, eg.: `SELECT 1 as Num`) or `PQsendPrepare`

mcollina:
@brianc what would be the next steps here?

brianc:
Just gotta review it & get it merged in, friend! Maybe tomorrow, definitely by Wendesday!

charmander:
Is the plan not still to replace the query queue with pipelining in pg 9?

luk-bmpi:
Really looking forward to pipelining in the pg node client. If it's working as expected, we will be able to meet our SLAs also with PostgreSQL - currently MySQL with the mariadb client is the only working combination for us. So: thanks, @mcollina, @brianc and all the other contributors!

brianc:
yo sorry for the massive delay on my side here, i've been on sabbatical and really not much computer interaction for the past 3-4 months. It's been pretty nice, but things have naturally backed up for me w/ pg. I'm recitifying that now.  I'll pull down this PR & play with it today, but from initial inspection it looks shippable & I'm looking forward to getting this out ASAP (hopefully right after I review it provided there aren't problems).

brianc:
Okay so the only thing I'm seeing here is the api of `client.pipelining = true` versus passing it as an option to the client constructor. Is there a reason for this design over

```js
const client = new Client({ host: 'foobar.com', pipelining: true })
```

also maybe rename it to `pipeline` from `pipelining`? I am bad at spelling & already misspelled it just in this comment 🙃

charmander:
Is there any good reason not to remove the query queue first and make pipelining always-on? Seems like that would greatly simplify the implementation.

brianc:
> Is there any good reason not to remove the query queue first and make pipelining always-on? Seems like that would greatly simplify the implementation.

I think the disruption introduced by removing it going to be very high, tbh. Even when I removed it myself I had to change over 100 tests (granted, they old and relied on it out of them being written before async/await).  I'm having second thoughts about removing the query queue at all...simply because it shipped so long ago and there's just a mountain of code written assuming it exists. And it doesn't hurt _that bad_ to maintain. I don't _like it_ and I think it's an unclear API...but there were some valid points raised on [this PR](https://github.com/brianc/node-postgres/pull/3603#issuecomment-4350401082) that made me have second thoughts.

SO what I'm thinking is in 9.x you can turn _off_ query pipelining and then it will throw if you try to send while another is in flight, but pipelining is on by default....but we can technically ship the "turn it on now as an opt in" in the 8.x branch and the get tires kicked. Reading the documentation for pipelining I don't see any drawbacks to having it be on by default.

What do you think?

charmander:
> I think the disruption introduced by removing it going to be very high, tbh.

Sorry for the bad phrasing; what I meant was replacing the query queue with pipelining in one step in pg 9, with no option to turn pipelining off (unless someone wants that for some reason?).

mcollina:
We have to support turning pipelining off, as some old pg might not have it yet.

mcollina:
> Okay so the only thing I'm seeing here is the api of `client.pipelining = true` versus passing it as an option to the client constructor. Is there a reason for this design over
> 
> 
> 
> ```js
> 
> const client = new Client({ host: 'foobar.com', pipelining: true })
> 
> ```
> 
> 
> 
> also maybe rename it to `pipeline` from `pipelining`? I am bad at spelling & already misspelled it just in this comment 🙃 

I'll do the change/remame.

charmander:
> We have to support turning pipelining off, as some old pg might not have it yet.

Could you expand on this?

mcollina:
> Could you expand on this?

PostgreSQL 14+ is needed.

charmander:
https://www.postgresql.org/docs/18/libpq-pipeline-mode.html:

> While libpq's pipeline API was introduced in PostgreSQL 14, it is a client-side feature which doesn't require special server support and works on any server that supports the v3 extended query protocol.

mcollina:
Given how many "postgresql compatible" servers are out there, I would want to keep the escape hatch when flipping the default. If nobody complains, remove the queue.

charmander:
> Given how many "postgresql compatible" servers are out there, I would want to keep the escape hatch when flipping the default. If nobody complains, remove the queue.

Ah, weird, this comment didn’t show up for me until somewhat recently. (I got an e-mail notification but assumed it was deleted.)

Is this an entirely theoretical incompatibility? A “PostgreSQL-compatible” server would have to be especially badly implemented for pipelining not to work. I think the best way forward is to create a clean and more straightforwardly correct implementation in the new major version, and if anyone *does* complain on a configuration worth supporting, that can be fixed.

mcollina:
@brianc ping?

mcollina:
@brianc ping?

nigrosimone:
@brianc any update on this? Is there anything blocking the merge, or something we can help with?

If you're not ready for a stable release, publishing it under a beta or alpha tag on npm would allow the community to test it more easily and provide feedback before it lands in the main release.

brianc:
> If you're not ready for a stable release, publishing it under a beta or alpha tag on npm would allow the community to test it more easily and provide feedback before it lands in the main release.

sorry I don't think this is a breaking change so it could probaby go out in a minor. Looking at this now. I've had some life stuff come up recently which has caused me an unfortunate amout of slowdown across most areas not directly related to it.

PR #3656, opened 2026-04-13
Fix pg prototype pollution via server supplied column names

hjr3:


packages/pg/lib/result.js
```diff
@@ -61,7 +61,7 @@ class Result {
   }
 
   parseRow(rowData) {
-    const row = { ...this._prebuiltEmptyResultObject }
+    const row = Object.assign(Object.create(null), this._prebuiltEmptyResultObject)
```
charmander: Backward compat break, and I don’t think it’s necessary as long as `_prebuiltEmptyResultObject` is built correctly.

packages/pg/lib/result.js
```diff
@@ -102,7 +102,7 @@ class Result {
       }
     }
 
-    this._prebuiltEmptyResultObject = { ...row }
+    this._prebuiltEmptyResultObject = Object.assign(Object.create(null), row)
```
charmander: Can be reverted, similar to previous comment. Only the initial `const row = Object.create(null)` is necessary.

packages/pg/test/integration/client/simple-query-tests.js
```diff
@@ -19,7 +19,7 @@ test('simple query interface', async function () {
     })
     query.once('row', function (row) {
       test('returned right columns', function () {
-        assert.deepStrictEqual(row, { name: row.name })
+        assert.deepStrictEqual(Object.keys(row), ['name'])
```
charmander: & revert test loosening accordingly

charmander (changes requested):

packages/pg/lib/result.js
hjr3: yes, great point. reverted.

packages/pg/lib/result.js
hjr3: reverted

packages/pg/test/integration/client/simple-query-tests.js
hjr3: for sure. i was fighting with test failures that were not matching my local. i never intended to keep this as-is. reverted

brianc:
Hey nice - sorry for the delay! Been on a sabbatical 💤

PR #3660, opened 2026-04-22
remove deprecated 'version' attribute from docker-compose.yml

The-Alchemist:
Removes the warning:

```
WARN[0000] .../node-postgres/.devcontainer/docker-compose.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion
```

The `version` attribute has been removed as of v2 of the docker compose plugin (https://github.com/compose-spec/compose-spec/blob/main/spec.md#version-top-level-element-obsolete).


brianc:
TY!

PR #3661, opened 2026-04-27
Explicit Resource Management for pg pool client

### LINKED ISSUE #3515: Symbol.asyncDispose support
I propose support `AsyncDisposable` on `Client` (https://github.com/tc39/proposal-explicit-resource-management) so that developers can opt-into ending the connection automatically at the end of the current scope.

**before**

```ts
import { Client } from 'pg'
const client = new Client()
await client.connect()
 
try {
   const res = await client.query('SELECT $1::text as message', ['Hello world!'])
   console.log(res.rows[0].message) // Hello world!
} catch (err) {
   console.error(err);
} finally {
   await client.end()
}
```

**after**

```ts
import { Client } from 'pg'
await using client = new Client()
await client.connect()
 
try {
   const res = await client.query('SELECT $1::text as message', ['Hello world!'])
   console.log(res.rows[0].message) // Hello world!
} catch (err) {
   console.error(err);
}
```

hyperair:
Add support for [Explicit Resource Management](https://tc39.es/proposal-explicit-resource-management/) for `pg.Pool` clients.

This is implemented using `[Symbol.dispose]` instead of `[Symbol.asyncDispose]` because `client.release()` is a synchronous function, not async.

Additionally, add a second property `client.destroyOnDispose` so that `Symbol.dispose` supports calling `client.release(true)`. This is useful for a usecase that changes connection parameters in a way that is troublesome to reset upon release to the pool, e.g.

```js
const pool = new Pool({
    database: 'foo',
    user: 'foo',
    password: 'foo',
    port: 5432,

    statement_timeout: 10_000,
});

async function getSlowClient() {
    await client.query("SET statement_timeout = 0");
    client.destroyOnDispose = true;
}

async function performReallySlowQuery() {
    using client = await getSlowClient();

    client.query(`select pg_sleep_for('30 minutes')`);
}
```

Fixes: #3515

Note: This PR has been rebased on top of https://github.com/brianc/node-postgres/pull/3662, so there are some irrelevant commits until that PR is merged.

### DIFF
--- a/docs/pages/apis/pool.mdx
+++ b/docs/pages/apis/pool.mdx
@@ -207,8 +207,8 @@ const pool = new Pool()
 
 {
   // check out a single client
-  using client = await pool.connnect()
-  client.destroyOnDispose = true;
+  using client = await pool.connect()
+  client.destroyOnDispose = true
 
   // client.release(true) implicitly called at the end of scope
 }
--- a/docs/pages/apis/pool.mdx
+++ b/docs/pages/apis/pool.mdx
@@ -187,6 +187,10 @@ assert(pool.idleCount === 0)
 assert(pool.totalCount === 0)
 ```
 
+`client.destroyOnDispose: boolean`
+
+`client[Symbol.dispose]() => void`
+
 Alternatively, pool clients support [Explicit Resource Management](https://tc39.es/proposal-explicit-resource-management/), which means you can use the `using` syntax to release them if your runtime supports it.
 
 ```js
@@ -195,13 +199,22 @@ import { Pool } from 'pg'
 
 const pool = new Pool()
 {
-
   // check out a single client
   using client = await pool.connect()
 
   // client.release() implicitly called at the end of scope
 }
+
+{
+  // check out a single client
+  using client = await pool.connnect()
+  client.destroyOnDispose = true;
+
+  // client.release(true) implicitly called at the end of scope
+}
 ```
+
+If you want the client to be destroyed instead of being returned to the pool at the end of the scope, (i.e. calling `client.release(true)` at the end of the scope), set the `destroyOnRelease` property to `true`.
 
 <Alert>
   <div>
--- a/docs/pages/apis/pool.mdx
+++ b/docs/pages/apis/pool.mdx
@@ -187,6 +187,22 @@ assert(pool.idleCount === 0)
 assert(pool.totalCount === 0)
 ```
 
+Alternatively, pool clients support [Explicit Resource Management](https://tc39.es/proposal-explicit-resource-management/), which means you can use the `using` syntax to release them if your runtime supports it.
+
+```js
+
+import { Pool } from 'pg'
+
+const pool = new Pool()
+{
+
+  // check out a single client
+  using client = await pool.connect()
+
+  // client.release() implicitly called at the end of scope
+}
+```
+
 <Alert>
   <div>
     You <strong>must</strong> release a client when you are finished with it.
--- a/packages/pg-pool/index.js
+++ b/packages/pg-pool/index.js
@@ -341,6 +341,12 @@ class Pool extends EventEmitter {
 
     client.release = this._releaseOnce(client, idleListener)
 
+    if (Symbol.dispose) {
+      client[Symbol.dispose] = function () {
+        this.release()
+      }
+    }
+
     client.removeListener('error', idleListener)
 
     if (!pendingItem.timedOut) {
--- a/packages/pg-pool/index.js
+++ b/packages/pg-pool/index.js
@@ -342,8 +342,9 @@ class Pool extends EventEmitter {
     client.release = this._releaseOnce(client, idleListener)
 
     if (Symbol.dispose) {
+      client.destroyOnDispose = false
       client[Symbol.dispose] = function () {
-        this.release()
+        this.release(this.destroyOnDispose)
       }
     }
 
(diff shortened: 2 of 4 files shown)


hyperair:
Hmm, looks like eslint needs to be upgraded before it'll parse the `using` test

packages/pg-pool/test/disposable-clients.js
```diff
@@ -0,0 +1,25 @@
+const Pool = require('../')
+
+const expect = require('expect.js')
+
+describe('disposable clients', () => {
+  it('defines a callable [Symbol.dispose]() when symbol is present', async () => {
+    const pool = new Pool({ max: 1 })
+    const client = await pool.connect()
+
+    if (Symbol?.dispose) {
```
charmander:
```suggestion
    if (Symbol.dispose) {
```

packages/pg-pool/index.js
```diff
@@ -341,6 +341,13 @@ class Pool extends EventEmitter {
 
     client.release = this._releaseOnce(client, idleListener)
 
+    if (Symbol.dispose) {
```
charmander, line 344: I think this would normally go on the prototype, but pg-pool’s approach is already a bit messy (it’s lacking a `PooledClient` object that’s specific to one acquisition operation) so it’s also not the end of the world if the patch goes in as-is.

docs/pages/apis/pool.mdx
```diff
@@ -187,6 +187,35 @@ assert(pool.idleCount === 0)
...
+  // check out a single client
+  using client = await pool.connect()
+
+  // client.release() implicitly called at the end of scope
+}
+
+{
+  // check out a single client
+  using client = await pool.connnect()
+  client.destroyOnDispose = true;
```
charmander:
```suggestion
  using client = await pool.connect()
  client.destroyOnDispose = true
```

charmander (commented):
Not sure about `destroyOnDispose`, i.e. whether its job is common enough to justify it and whether it’s the best API for the job. Would suggest initializing to `false` if kept. ~~Note `.release(true)` continues to work for error cases given that double release is a no-op~~; I recognize that it doesn’t satisfy the use case you mentioned. Maybe release the feature without that API initially, since it can be added easily?

Edit: Sorry, I don’t know where I got that from – we haven’t even done `PooledClient` yet.

packages/pg-pool/index.js
hyperair: Agreed, and that's why I implemented it like this.

`@types/pg` has a `PooledClient` type with the `release` signature, and I intend to make a PR there to add the `[Symbol.dispose]` addition later.

hyperair:
> Not sure about destroyOnDispose, i.e. whether its job is common enough to justify it and whether it’s the best API for the job. Would suggest initializing to false if kept. Note .release(true) continues to work for error cases given that double release is a no-op; I recognize that it doesn’t satisfy the use case you mentioned. Maybe release the feature without that API initially, since it can be added easily?

Yeah I wasn't sure about the API either but I couldn't think of a better way to do it. Maybe a `client.setError()` to set an error state that `client.release()` acknowledges as well?

That said, unfortunately I don't think relying on double-release is workable either, because the second release, which will be called from within `Symbol.dispose` will throw. Even if `[Symbol.dispose]()` wraps that in a try-catch to swallow the error, needing to put the `client.release(true)` in a finally clause obviates any purpose of using the `using` syntax.

olsonpm:
I got here because I'm implementing dispose on a client wrapper.

One question, why implement `destroyOnDispose` as a mutable property instead of an internal one initialized by `pool.connect()`?  e.g.

```js
using client = await pool.connect({ destroyOnDispose: true });
```

Maybe I'm not understanding the use case you mention.

hyperair:
@olsonpm That does work for my usecase, I just didn't think of it.

But on the other hand it's still a shift from how late you can make the decision not to return the client to the pool -- with `client.release(true)`, you're able to make this decision all the way at the end of your session (starting from pool.connect until client.release) with this client, but with a `pool.connect` parameter, you need to commit to destroying the client all the way at the beginning of your session, before you even have the client instance.

An example where it might be desirable to destroy the client rather than release it back to the pool would be if you made a `client.query` and caught an error that indicated that the client was no longer usable. You might then want to set `destroyOnDispose = true` instead of letting the client go back to the pool.

olsonpm:
Ah thanks for that use case Hyperair.  I was aware the decision would change the semantics as you mentioned, but couldn't think of why you would want that flexibility.

brianc:
Hey thanks for this! I actually set out to do this a while ago and due to the linting and it not working on older versions of node I grew frustrated and thought I'd do it later. And now it's done!

I'm also not sold on `destroyOnDispose` though...it is kinda a weird API to me.  I can't really think of a better way to handle it though. I found it kinda frustrating actually in the `dispose` ecmascript design there's no way to know if an error was thrown in the scope or if the scope exited "naturally."  I took a stab at having a `using client.transaction()` type thing but not being able to issue a rollback on error was a show stopper.

One possible alternative is to just allow ended clients to be returned to the pool (via the disposal or via `client.release()`) and if the pooled client has had 'end' called then it is evicted automatically from the pool. This would prevent likely error scenarios of misuse anyway where one accidentally returns an ended client (obviously not recommended, but could happen) and then the code would look like this

```js
using client = await pool.connect()
try {
  client.query("EXPLODE!")
} catch (e) {
  client.end()
  throw e // (or just log it if you want or whatever)
}
```

Thoughts on that?

hyperair:
@brianc, thanks for the feedback.

Hmm, I like the idea of `client.end()` working that way, but it still doesn't allow me to mark a client as `dirty` early in the session (i.e. before queries are made) as per the use-case in the PR description. Do you have any ideas on how to support that sort of usecase?

PR #3662, opened 2026-04-27
Upgrade eslint and typescript

hyperair:
This is required for tests that use the new `using` syntax that was introduced in
https://tc39.es/proposal-explicit-resource-management/, because the current eslint version doesn't support it (see
errors in #3661), and upgrading eslint requires an upgrade of typescript as well.

Changes:
- Upgrade `eslint`, `typescript`, `@types/node`, and some related packages
- eslint: Port config to new flat config format
- Fix preserve-caught-error eslint warning
- Drop unused eslint-disable-line
- Fix various new typescript errors
- Workaround https://github.com/microsoft/TypeScript/issues/63447


eslint.config.mjs
```diff
@@ -0,0 +1,66 @@
+import { defineConfig, globalIgnores } from "eslint/config";
```
charmander: Looks like ESLint needs at least a one-off run on its own configuration :)

hyperair: Hmm yes lemme do that, thanks

hyperair: Fixed in [a4ff1bf](https://github.com/brianc/node-postgres/pull/3662/commits/a4ff1bf595d0a36b6a1e65b2bffb4245dc36917a)

charmander (approved):

brianc:
Thank you so much!! This kinda work can be tedious & I appreciate it. ❤️

Issue #3663, opened 2026-04-27
supporting ECMAScript Temporal


jochemd:
Word is that [Temporal](https://tc39.es/proposal-temporal/docs/) has reached [Stage 4](https://github.com/tc39/proposals/blob/main/finished-proposals.md) and will be included in the next ECMAScript draft. I think it looks very promising and will allow for much better support for some PostgreSQL datatypes by mapping them to proper types instead of strings.
I took Temporal for a spin with the latest NodeJS nightly and ran some [really basic happy path tests](https://github.com/jochemd/pg-temporal/tree/main) for the no-brainer datatype mappings:
1. [Temporal.Duration](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/Duration) to PostgreSQL `interval`
2. [Temporal.PlainDate](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/PlainDate) to PostgreSQL `date`
3. [Temporal.PlainTime](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/PlainTime) to PostgreSQL `time`

Other mappings might not be so obvious and/or a matter of personal preference (e.g. I have no use for [Temporal.ZonedDateTime](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/ZonedDateTime) and strongly prefer to get everything out of the database as [Temporal.Instant](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/Instant)). And obviously backward compatibility is its own can of worms.

Temporal support was also requested in the [pg@9.0 breaking changes ideas](https://github.com/brianc/node-postgres/discussions/3598#discussioncomment-15932019) discussion, but it has not been added to the [pg@9.0 milestone](https://github.com/brianc/node-postgres/milestone/6).  Is there a specific reason temporal not added to the milestone? Sorting out pg_types first?

Is there a template/guidance on how to propose changes that are not backward-compatible and can be enabled using a simple mechanism so it is possible to just add one option / env variable that switches all datetime parsing to Temporal datatypes instead?

james-pre:
[...]

PR #3664, opened 2026-04-28
RFC: Modernize the monorepo (ESM-only, TypeScript, pnpm, vitest, Node 22+)

### LINKED ISSUE #1796: Merge TypeScript declarations into repo
First of all big kudos out for this great piece of software! I don't know what I would do without it.

Now considering the recent growth of TypeScript usage across the JS / node eco system and the revival of static typing in general:
*Would you consider a PR that merges the definitely-typed TS declarations into the main repo?*

### Background

- It's tricky to keep the typings package version and the actual package's version in sync over
- Not only for TypeScript users, but also for useful hints in the IDE
- Coherence: Keep together what's tightly coupled
- Submitting a PR to that beast of definitely typed monorepo is a nightmare
- One `npm install` instead of two

Let me know what you think :)

Happy holidays!

### LINKED ISSUE #1930: Query callbacks have incorrect type definitions
The callback functions for queries have type definitions that say that two arguments will be provided, one of type Error and one of type QueryResult. However, one of these arguments will be null which allows typescript code to be written in strict mode that does not account for these nullable arguments.

### LINKED ISSUE #2137: Add support for native ESM modules
Hi and thank you for the great library and endless support!

It would be great to have this library [export native named exports](https://nodejs.org/api/esm.html#esm_conditional_exports) so we could import it natively on new Node versions.

Here is an example of how you would do it now, without that support: https://github.com/Urigo/typescript-node-es-modules-example/commit/98304173e964713955be3a92b4e355a857376dca#diff-f41e9d04a45c83f3b6f6e630f10117feR10-R11

### LINKED ISSUE #2353: Add ESM support
When `import`ing a CommonJS library, it's flattened into a single object, and you're unable to take advantage of any ESM features. However if you add a few lines to your `package.json` and a `esm.mjs` wrapper file that simply re-exports the contents of `index.js`, then ESM users will be able to take full advantage of ESM features without any changes needed to your source code, and without affecting any CommonJS users. The `package.json` changes would be as simple as:
```js
"main": "./index.js",
"exports": {
  ".": [
    {"require": "index.js", "import": "esm.mjs"},
    "index.js"
  ],
  "./": "./"
}
```
The `"index.js"` fallback is only necessary to avoid `package.json` parsing errors in some older versions of Node, which would affect CommonJS users as well. If this library ever supports only Node 14+ in the future, the fallback could be removed.

### LINKED ISSUE #2662: Rewrite in TypeScript
@brianc Happy to help if this was considered.

### LINKED ISSUE #2763: TypeScript: how to import the `NoticeMessage` type?
I'm using the "pg" package in my TypeScript program. I'm using `.on('notice', ...)` to listen for notices:

```ts
client.on('notice', (notice: NoticeMessage) => ...);
```

What's the right way to import `NoticeMessage`? It looks like it's not explicitly exported from pg-protocol: [pg-protocol/src/index.ts](https://github.com/brianc/node-postgres/blob/68160a29bd8dfe97c74ab9a74000977da7783d6f/packages/pg-protocol/src/index.ts).

This seems to work:

```ts
import {NoticeMessage} from 'pg-protocol/dist/messages';
```

But peeking into the "dist" subfolder feels brittle.

### LINKED ISSUE #2871: Migrating some packages to TypeScript.
Hi @voodootikigod @jeremyevans @dmnd @pshc @mikl are you planning to migrate completely to TypeScript ? If not I can raise a pull request for this as a contribution.
Thanks!

### LINKED ISSUE #3487: Bind does not generate the result column format encoding in pg-protocol

### LINKED ISSUE #3492: Cannot find module 'pg-protocol/dist/messages' or its corresponding type declarations.
the file `node_modules/@types/pg/index.d.ts` make use of an import that does not seem to works.

A valid import is `import { NoticeMessage } from 'pg-protocol/dist/messages.js'`. Maybe you changed the package.json of `pg-protocol` and it broke the exports

### LINKED ISSUE #3641: Publish pg@9 to JSR
Hello 👋🏼 

It would be nice to publish this package into [jsr.io](https://jsr.io/). It is a new registry for Deno, Node.js, Bun and other runtimes. 

One of the cool things, you may write the entire package in Typescript and forget the transpilation process. JSR will be on charge of that so any runtime can grab and get the proper JavaScript + typings.

productdevbook:
Closes #2871
Closes #2353
Closes #2137
Closes #2662
Closes #1796
Closes #3492
Closes #2763
Closes #1930
Closes #3641
Closes #3487

Hi Brian — first off, thank you for ~15 years of node-postgres (and for the fact that you're still actively maintaining it — the recent v8.20 release is proof). It's been the backbone of countless projects.

This PR is intentionally large and I'm opening it as an **RFC / proof-of-concept**, not as something I expect to merge as-is. The aim is to share a working implementation of what a modernized node-postgres monorepo could look like, so it can serve as a concrete starting point for discussion (or get cherry-picked piece by piece, or rejected outright — all fair).

**This is still active work in progress on my side.** I'll keep pushing fixes here so the diff stays current.

I rebuilt the monorepo end-to-end to match the toolchain I use across my own libraries (etiket, hucre, misina, sumak). The branch is at https://github.com/productdevbook/ts-node-postgres/tree/feat/modernize-esm-ts-vitest if you'd rather just browse the diff there.

## What changed at a glance

| Area | Before | After |
|------|--------|-------|
| Module system | CJS + thin ESM wrappers | **ESM-only** (\"type\": \"module\") |
| Node baseline | >= 16 | **>= 22.11** |
| Package manager | yarn + lerna | **pnpm 10 workspaces** |
| TypeScript | 4.0.3 (3 packages) | **6.x everywhere** |
| Build | tsc --build | **obuild** (transform → .mjs + .d.mts) |
| Lint / format | eslint + prettier | **oxlint + oxfmt** |
| Test | mocha + chai + Makefile | **vitest** (flat test/ per package) |
| Typecheck | tsc | **tsgo** (@typescript/native-preview) |
| Versioning | lerna independent | **bumpp -r** + changelogithub |
| Imports | require('./x') | import { x } from './x.ts' (per misina convention) |
| Exports | main / types | granular subpath exports map (./client, ./pool, ./driver/*, …) |

## Per-package status

All 8 packages converted (lib/ → src/, JS → TS, mocha → vitest):

- **pg** → 9.0.0 — 133 JS files ported to TS, src/ split into client/, connection/, query/, types/, crypto/, native/, utils. Public surface preserved (Client, Pool, Connection, types, DatabaseError, defaults, Query, Result).
- **pg-pool** → 4.0.0 — single-file CJS rewritten as a typed Pool class.
- **pg-cursor** → 3.0.0
- **pg-query-stream** → 5.0.0 — already TS, tightened strict mode.
- **pg-protocol** → 2.0.0 — already TS, ESM-only output.
- **pg-cloudflare** → 2.0.0 — workerd condition preserved; non-workerd fallback ships a no-op CloudflareSocket with the same public shape so consumers can `import { CloudflareSocket }` from any runtime.
- **pg-connection-string** → 3.0.0 — JS + .d.ts fused into a single TS module.
- **pg-native** → 4.0.0 — libpq binding still works via ESM default import.

## Issues this addresses

If this lands (in any form), the following long-standing issues become resolvable:

- **#2871** — Migrating some packages to TypeScript (umbrella tracker)
- **#2353** — Add ESM support
- **#2137** — Add support for native ESM modules
- **#2662** — Rewrite in TypeScript
- **#1796** — Merge TypeScript declarations into repo (TS source means types are intrinsic, no more @types/pg drift)
- **#1930** — Query callbacks have incorrect type definitions
- **#2763** — TypeScript: how to import the NoticeMessage type? (granular subpath exports: \`pg-protocol/messages\`)
- **#3492** — Cannot find module 'pg-protocol/dist/messages' or its corresponding type declarations
- **#3641** — Publish pg@9 to JSR (now JSR-publishable)
- **#3487** — Binary-mode INT[] roundtrip returning empty array (fixed during the migration; was a real regression in the JS→TS port of Result#parseRow)

## Open / superseded PRs in flight

- **#3662** — \"Upgrade eslint and typescript\" → eslint replaced by oxlint, TS bumped to 6.x; superseded.
- **#3646** — \"build(deps-dev): bump lerna 3 → 9\" → lerna removed entirely; obsolete.
- **#3648** — \"build(deps-dev): bump eslint-plugin-prettier\" → eslint+prettier replaced; obsolete.

I haven't touched **#3168** (\"build pg-cloudflare as a CommonJS module\") because it's solving the same Workers test issue from the opposite direction — that ticket can stay open as a fallback if this RFC isn't the path forward.

## Verification

GitHub Actions: **Lint & Typecheck pass · Test (Node 22) pass · Test (Node 24) pass.** Locally with the included docker-compose:

    corepack enable
    pnpm install
    docker compose up -d
    set -a && source .env.test && set +a
    pnpm typecheck   # 0 errors
    pnpm lint        # 0 errors, 0 warnings
    pnpm build       # 8/8 packages emit .mjs + .d.mts
    pnpm test        # exit 0

| Package | Pass / Total |
|---------|--------------|
| pg-protocol | 73/73 |
| pg-connection-string | 71/71 |
| pg-cloudflare | 6/6 |
| pg-native | 75/75 |
| pg-pool | 85/85 (2 native-only skipped) |
| pg-cursor | 37/37 |
| pg-query-stream | 40/40 |
| pg | 426/426 (6 native-only skipped) |
| pg-esm-test | 22/22 |
| **Total** | **835/835 (8 platform-skipped)** |

\`pnpm test\` exits 0 cleanly — lint, typecheck, build and the full suite all pass.

## Remaining skips

The 8 skipped tests are all native-only paths gated on the optional \`pg-native\` peer being present. They do execute (and pass) on CI thanks to the \`libpq-dev\` install + \`pnpm rebuild libpq\` step in \`.github/workflows/ci.yml\`; the per-test skips guard the cases where individual native-client wiring would otherwise fail when the dependency wasn't loaded.

## Local infra added

- \`docker-compose.yml\` — postgres-ssl service with healthcheck.
- \`scripts/init-scram.sql\` — bootstraps the SCRAM test role the integration suite expects.
- \`.env.test\` — convenience env file (\`set -a && source .env.test\`).

## Breaking changes (intentional)

- **CJS consumers are broken.** require('pg') no longer works. This is the core trade-off of going ESM-only. Major-version bump for every package reflects that.
- engines.node >= 22.11. Node 16/18/20 dropped.
- Bundle layout changed from dist/index.js (CJS) to dist/index.mjs + dist/index.d.mts.
- Several internal modules now have explicit subpath exports; users reaching into pg/lib/... will need to update.

## Why I'm sending this

I wanted to put a working, end-to-end modernization on the table rather than just opinions on Twitter. If any of this is useful — even just the workflow files or one package's conversion — please cherry-pick freely. If the ESM-only direction isn't where you want node-postgres to go, that's a totally fair answer; closing this PR doesn't bother me.

Happy to break it up into smaller, more reviewable PRs (per package, or just the tooling), to host the result as a fork under a different name, or to walk through any specific decision. Whatever's least friction for you.

— Wind (@productdevbook)

---

PS: I'm currently looking for a full-stack role — CV / portfolio: https://productdevbook.com 🙂

🤖 Generated with [Claude Code](https://claude.com/claude-code)


brianc:
Holy smokes this is amazing. I am looking forward to looking this over. I'm just wrapping up some final travel - took a bit of a sabbatical after work, then will dive head first into node-postgres soon! Def not stepping back, just needed a computer break after doing the startup sprint stuff for years on end.

james-pre:
Might I suggest sticking with regular `typescript` for now? `@typescript/native-preview` is only temporary and TS 7.0 will use the native port for `tsc`. Staying with the regular `typescript` package would decrease maintaince burden and dependency churn.

Also for pg@9 I'm hoping #3663 makes the cut.

1valdis:
> CJS consumers are broken. require('pg') no longer works.

If packages don't have top-level await, it's possible to make them require-able even if they are ES modules. Node supports `require(ESM)` for quite some time now. This works via
```diff
 "exports": {
   ".": {
     "types": "./path-to-es-module.d.ts",
     "import": "./path-to-es-module.js",
+    "module-sync": "./path-to-es-module.js"
   }
}
```
and it's the way I use for a ESM-only library that I maintain.

> import { x } from './x.ts'

I wonder why this is necessary instead of `.js` imports. The only pro I see is it looks kind of more beautiful but if you work with TypeScript for years then you know `.js` takes no additional compiler and dev environment settings and is much more likely to "just work".

james-pre:
Note sure how relavent it is, for `moduleResolution: NodeNext`, having `.js` in imports and exports is required. For projects I maintain (which are ESM), I use TS' `verbatimModuleSyntax` and use `.js` extensions in the TS source. For example `import { Duck } from './duck.js';` inside index.ts.

brianc:
> or rejected outright — all fair

this wont happen. 😄  Many of the things you did here are things I have kinda wanted to do for a while but for life reasons haven't gotten around to.

> All 8 packages converted (lib/ → src/, JS → TS, mocha → vitest):

That's awesome. vitest is very nice. and the typescript conversion is great too.

There are a bunch of things that might block this, particularly as it being a giga-huge PR...I'll try to outline some of them.

1) ESM-only is a breaking change. Would need to be a semver major.
2) Node baseline: I support all LTS versions of node, so we'd need to bring that down to node 18 (or 20 on a semver major bump) but since 20 is still actively maintained being compatible with it is something I have an implicit community contract w/ the folks who've installed `pg` I am committed to maintaing.
3) exports structure changing - also semver major though I was intending something similar with pg@9.0. 
4) some of these tools I'm extremely unfamiliar with (oxlint, oxfmt, obuild, misina) - tooling is such a minor issue that doesn't impact the user base I'd rather change less & endure minorly more pain than churn a bunch of tooling based on preferences or "this is better, but only a bit"
5) this is basically a "close all PRs before this merges" level of disruption of merge conflicts to the source code...which can be okay for sure...i've done it before. Just pointing it out.
6) Uhhh there's so much code to read it's scary to think a vuln could be slipped in here. I would need a full half-day to go through a single PR of this size probably.

There are a bunch of things I really love about this:
1) rewriting everything in TS is excellent
2) standardizing vitest is excellent
3) ditching lerna is 👍 - i use pnpm for my own work these days
4) the local infra added stuff - all of it

Some things I'm still curious about:
1) Would need some updated (and honestly probably a good idea to do this) maintained developer docs on how to publish & do the various mono-repo things w/ the new tooling.
2) why oxlint + oxfmt vs eslint & prettier? I don't usually use the new fancy thing but rather "old standby" on pg since its kinda been around while other "new hotness" things have come & gone.
3) why use tsgo now since it's not released yet? Does it hurt to just use tsc and upgrade to tsgo once it goes stable?

I think it'd be worthwhile discussing a path forward to incrementally land some of these things on the 9.x branch.  There isn't a world where me landing a +26k, -31k PR on this project is going to happen but more incremental PRs would be very useful. I want to be mindful too I've been AWOL for a while on sabbatical so I've got some open PRs  I need to address or merge before anything too disruptive happens.

I think sequencing this out...if I had to do it would be starting w/ less disruptive things to existing PRs.


1) land the local infra changes isolated w/ better developer documentation on how to lean on that
2) `yarn -> pnpm` - this is absolutely a win imo and not too code disruptive
3) `mocha / custom tests -> vitest` - will impact a bunch of code but all test code & should read through and make sure things are still good
4) migrate one single library at a time to typescript in its own PR so its reviewable, and at that time publish the types w/ that library. Probably start at the bottom of the dependency chain (I think its pg-native) and work up from there?
5) change the exports to be `import { Client } from 'pg/native'` instead of `const { Client } = require('pg').native`

Basically break it down into individual, reviewable changes along a path of improvement, target 9.x branch, and we can discuss each piece in detail.

Sorry about the delay here.

james-pre:
My two cents on tooling is that we should stick with `tsc`, `prettier`, and `eslint` since they are a lot more common and imo more reliable.

vfil:
@brianc I would like to be involved in this change. We all have our time constraints and I totally agree with your last comment. I decided to allocate some time for OSS, something I wanted to do for long time.
As I am not very familiar with this project I would skip:

> 1. land the local infra changes isolated w/ better developer documentation on how to lean on that

and go with:

> 2. yarn -> pnpm - this is absolutely a win imo and not too code disruptive
This will help to get more confidence.

What do you think would this be a good-first-issue?

PR #3667, opened 2026-05-08
Node JS 26

cesco69:
- update libpq to 1.11.0 for pg-native on nodejs 26
- add node 26 into test CI (and remove node 25)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
           - '20'
           - '22'
           - '24'
-          - '25'
+          - '26'
         os:
           - ubuntu-latest
     name: Node.js ${{ matrix.node }}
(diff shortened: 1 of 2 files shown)


brianc:
Thank you!

brianc:
I'll publish a new patch version a bit later today

charmander:
Isn’t 1.11.0 inside the `^1.8.15` version range? Might not need to publish anything. (Actually, why did CI fail without the change? Cache?)

brianc:
> Might not need to publish anything.

That was what I was originally thinking, yeah. But oh well I'll just do a patch version.



> CI fail without the change? Cache?

possible, though i actually didn't see it fail

nigrosimone:
It without libpq update, see this pr https://github.com/brianc/node-postgres/pull/3666

charmander:
This probably shouldn’t have been able to pass CI (`yarn install --frozen-lockfile`) without updating yarn.lock, but Lerna might be getting in the way.

PR #3677, opened 2026-05-14
Add scramMaxIterations option to limit SCRAM iteration count

sehrope:
Caps the number of SCRAM iterations the driver will perform during SASL auth, defaulting to 100000. Protects against malicious or misconfigured servers requesting unbounded PBKDF2 work. A value of zero disables the check entirely.

---

Went back and forth on adding it to the `continueSession(...)` vs `startSession(...)`. Ended up going with this approach as it makes it a bit easier to test the defaults and overrides.

We added something similar to this to Java driver pgjdbc recently with the same 100K default.

### DIFF
--- a/packages/pg/lib/crypto/sasl.js
+++ b/packages/pg/lib/crypto/sasl.js
@@ -30,7 +30,9 @@ function saslprep(password) {
   return password.replace(nonAsciiSpace, ' ').replace(mappedToNothing, '').normalize('NFKC')
 }
 
-function startSession(mechanisms, stream) {
+const DEFAULT_MAX_SCRAM_ITERATIONS = 100000
+
+function startSession(mechanisms, stream, scramMaxIterations = DEFAULT_MAX_SCRAM_ITERATIONS) {
   const candidates = ['SCRAM-SHA-256']
   if (stream) candidates.unshift('SCRAM-SHA-256-PLUS') // higher-priority, so placed first
 
@@ -53,6 +55,7 @@ function startSession(mechanisms, stream) {
     clientNonce,
     response: gs2Header + ',,n=*,r=' + clientNonce,
     message: 'SASLInitialResponse',
+    scramMaxIterations,
   }
 }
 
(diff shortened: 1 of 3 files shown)


packages/pg/lib/crypto/sasl.js
```diff
@@ -78,6 +81,18 @@ async function continueSession(session, password, serverData, stream) {
     throw new Error('SASL: SCRAM-SERVER-FIRST-MESSAGE: server nonce is too short')
   }
 
+  const scramMaxIterations =
+    typeof session.scramMaxIterations === 'number' ? session.scramMaxIterations : DEFAULT_MAX_SCRAM_ITERATIONS
```
charmander, line 85: Is `session.scramMaxIterations` always a number here (at least when used via `Client`)? If so, I think omitting the check makes sense. (And if supporting direct use really is a goal, `startSession` seems like the place to do it.)

sehrope: Yes it should always be a number as we do the `coerceNumberOrDefault(...)`.

That would simplify it a bit as the only place it's set is in `startSession` and this part would be just:


```js
  // a value of 0 disables the iteration count check
  if (this.scramMaxIterations !== 0 && sv.iteration > this.scramMaxIterations) {
    throw new Error(
      'SASL: SCRAM-SERVER-FIRST-MESSAGE: iteration count ' +
        sv.iteration +
        ' exceeds scramMaxIterations of ' +
        this.scramMaxIterations
    )
  }

```

sehrope: Oh wait I realize now that I did it like this because a bunch of the tests create their own session objects and call the other sasl functions:

```js
sasl.continueSession(
    {
      message: 'SASLInitialResponse',
      clientNonce: 'a',
    },
    badPasswordValue,
    'r=1,i=1'
  ),
  {
    message: 'SASL: SCRAM-SERVER-FIRST-MESSAGE: client password must be a string',
  }
)
```

So it wouldn't always be populated unless we update all the tests to add the default.

charmander: Maybe a shared wrapper function for test use that’s more faithful to how `Client` will use it?

brianc:
Thank you!

sehrope:
Thanks for merging.

Was going to follow up this week to see if I could tweak it per charmanders feedback. Will circle back to that when I have a some time as it shouldn't change any actual usage and would be good for the test suite anyway.

brianc:
For aure! That would be great, thanks. Yah wanted to get this out with the release (as a new feature) and figured the refactoring could always happen in a follow on.

PR #3678, opened 2026-05-15
Remove compatibility code for unsupported versions of Node (<16)

charmander:
Annoyingly, the default diff doesn’t show that lib/crypto/utils-webcrypto.js was renamed to lib/crypto/utils.js without changes (and that the previous version of the latter was deleted), but that’s what happened.


brianc (approved):
this is nice!

PR #3681, opened 2026-06-02
perf(pg-protocol): encode length-prefixed strings in a single pass

kylecannon:
## What

Adds `Writer.addInt32PrefixedString`, which writes a value's Int32 byte-length prefix immediately followed by its UTF-8 bytes, computing `Buffer.byteLength` once. The previous `addInt32(Buffer.byteLength(s)).addString(s)` pairing scanned each string three times (once for the prefix length, again inside `addString`, then the encode). The new method is used in two places:

1. Bind parameter values (`serializer.writeValues`), the common write path for parameterized queries.
2. The SASL initial response message.

## Why

Bind is on the hot path for every parameterized query, and the cost of the redundant `Buffer.byteLength` scans grows with parameter size. Doing the length prefix and the UTF-8 write in one pass removes two of the three string scans for each string parameter.

## Correctness

The serialized wire output is byte-identical to the previous `addInt32(Buffer.byteLength(s)).addString(s)` sequence:

- The Int32 length prefix is the UTF-8 byte length (`Buffer.byteLength`), not the character count, exactly as before.
- `this.ensure(4 + len)` is called before capturing the local buffer reference, so a backing-buffer growth cannot write through a stale reference.
- `addInt32` and `addString` are left unchanged for all other callers.

No public API or behavior changes. The new method is purely additive.

## Benchmark

Measured with an in-process serialization microbenchmark (no network), alternating between this branch and base each round to cancel thermal drift:

| case | improvement |
| --- | --- |
| bind, 2 small string params | about +16% ops/sec |
| bind, 10 mixed params | about +23% ops/sec |
| bind, multi-byte unicode param | about +14% ops/sec |
| full insert (parse + bind) | about +9% ops/sec |

The gain grows with parameter size, since the eliminated scans are proportional to string length.

## Testing

The `pg-protocol` test suite passes. This change adds a multi-byte unicode bind unit test that asserts the Int32 length prefix equals the UTF-8 byte length rather than the character count (the exact property the single-pass write must preserve, which the existing ASCII-only bind tests could not catch).

### DIFF
--- a/packages/pg-protocol/src/outbound-serializer.test.ts
+++ b/packages/pg-protocol/src/outbound-serializer.test.ts
@@ -129,6 +129,28 @@ describe('serializer', () => {
         .join(true, 'B')
       assert.deepEqual(actual, expectedBuffer)
     })
+
+    it('encodes a multi-byte string param with its UTF-8 byte length, not char length', function () {
+      // Guards the single-pass addInt32PrefixedString write path: the Int32
+      // length prefix must be the UTF-8 byte count, not String.length. 'héllo中🎉'
+      // is 7 code units / 8 chars but 13 UTF-8 bytes.
+      const value = 'héllo中🎉'
+      const bytes = Buffer.from(value, 'utf8')
+      assert.notEqual(bytes.length, value.length) // sanity: the divergence we're testing
+      const actual = serialize.bind({ values: [value] })
+      const expectedBuffer = new BufferList()
+        .addCString('') // portal
+        .addCString('') // statement
+        .addInt16(1) // param format code count
+        .addInt16(0) // format code for the one value (text)
+        .addInt16(1) // value count
+        .addInt32(bytes.length) // 13 — the UTF-8 byte length, NOT value.length (8)
+        .add(bytes)
+        .addInt16(1) // result format code count
+        .addInt16(0) // result format (text)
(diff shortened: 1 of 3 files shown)


packages/pg-protocol/src/outbound-serializer.test.ts
```diff
@@ -129,6 +129,28 @@ describe('serializer', () => {
         .join(true, 'B')
       assert.deepEqual(actual, expectedBuffer)
     })
+
+    it('encodes a multi-byte string param with its UTF-8 byte length, not char length', function () {
+      // Guards the single-pass addInt32PrefixedString write path: the Int32
+      // length prefix must be the UTF-8 byte count, not String.length. 'héllo中🎉'
+      // is 7 code units / 8 chars but 13 UTF-8 bytes.
```
charmander:
```suggestion
      // is 7 code points / 8 UTF-16 code units but 13 UTF-8 bytes.
```

charmander:
Other than that, looks good. Can you share the benchmark you used?

kylecannon:
@charmander here you go:

https://gist.github.com/kylecannon/32f626ef17ee2394bd30d64707852398

charmander (approved):

brianc:
sweet!

PR #3688, opened 2026-06-03
Add support for sslnegotiation=direct (PostgreSQL 17)

### LINKED ISSUE #3346: support PG 17 sslnegotiation=direct
sslnegotiation 

This option controls how SSL encryption is negotiated with the server, if SSL is used. In the default postgres mode, the client first asks the server if SSL is supported. In direct mode, the client starts the standard SSL handshake directly after establishing the TCP/IP connection. Traditional PostgreSQL protocol negotiation is the most flexible with different server configurations. If the server is known to support direct SSL connections then the latter requires one fewer round trip reducing connection latency and also allows the use of protocol agnostic SSL network tools. The direct SSL option was introduced in PostgreSQL version 17.

**postgres**

perform PostgreSQL protocol negotiation. This is the default if the option is not provided.

**direct**

start SSL handshake directly after establishing the TCP/IP connection. This is only allowed with sslmode=require or higher, because the weaker settings could lead to unintended fallback to plaintext authentication when the server does not support direct SSL handshake.

---

- PGJDBC https://github.com/pgjdbc/pgjdbc/pull/3252
- libpq https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLNEGOTIATION

Shion1305:
Adds support for PostgreSQL 17's `sslnegotiation` connection parameter. The default, `postgres`, keeps the existing behavior: the driver sends an `SSLRequest` packet and waits for the server's `S`/`N` reply before starting the TLS handshake. The new `direct` value starts the TLS handshake immediately after the TCP connection is established, saving a network round-trip and allowing protocol-agnostic TLS tooling. As in libpq, `direct` advertises the `postgresql` ALPN protocol during the handshake and requires SSL to be enabled.

Usage mirrors the existing `ssl` options — via config object, `PGSSLNEGOTIATION`, or connection string:

```js
new Client({ host, ssl: { rejectUnauthorized: false }, sslnegotiation: 'direct' })
new Client('postgres://user:pw@host/db?sslmode=require&sslnegotiation=direct')
```

Closes #3346.

---

A few design notes:

- `direct` requires SSL to be enabled (`ConnectionParameters` throws otherwise), matching libpq's rule that direct negotiation is only allowed with `sslmode=require` or higher so there's no silent plaintext fallback. In a connection string, `sslnegotiation=direct` enables SSL automatically when no other SSL option is present.
- The TLS upgrade in `connection.js` was extracted into `upgradeToSSL()` so both negotiation styles share it; only the `direct` path skips the `SSLRequest`/`S`-byte exchange and sets `ALPNProtocols`.
- `sslnegotiation` is also added to the libpq connection string for the native (`pg-native`) path.
- Tests cover connection-string parsing, `ConnectionParameters` validation, and the connection behavior itself (no `SSLRequest` packet is written in direct mode, and ALPN is set only for direct). I followed pgjdbc/libpq for the parameter semantics.

---

I've verified this against a personal PostgreSQL 17.5 instance configured to require direct SSL. With `sslnegotiation=direct`, the client establishes TLS 1.3, negotiates the `postgresql` ALPN protocol, and runs queries successfully; `pg_stat_ssl` confirms the connection is encrypted server-side. The same server rejects the traditional `sslnegotiation=postgres` handshake with `ECONNRESET`, confirming the two negotiation styles take genuinely different code paths.

### DIFF
--- a/docs/pages/features/ssl.mdx
+++ b/docs/pages/features/ssl.mdx
@@ -49,6 +49,31 @@ const config = {
 }
 ```
 
+## Direct SSL negotiation
+
+By default node-postgres uses the traditional PostgreSQL SSL negotiation: it sends an `SSLRequest` packet, waits for the server to acknowledge it, and only then starts the TLS handshake. PostgreSQL 17 and newer also support _direct_ SSL negotiation, where the TLS handshake begins immediately on connect (similar to HTTPS), saving one network round-trip.
+
+To use direct negotiation, set `sslnegotiation: 'direct'`. SSL must be enabled, and the server must be PostgreSQL 17+ configured to accept direct SSL connections.
+
+```js
+const config = {
+  database: 'database-name',
+  host: 'host-or-ip',
+  ssl: { rejectUnauthorized: false },
+  sslnegotiation: 'direct',
+}
+```
+
+It can also be supplied via a connection string. When `sslnegotiation=direct` is present, SSL is enabled automatically if not otherwise configured:
+
+```js
+const config = {
+  connectionString: 'postgres://user:password@host:port/db?sslmode=require&sslnegotiation=direct',
+}
+```
+
+Direct negotiation requests the `postgresql` ALPN protocol during the TLS handshake, as required by the server. The default value is `'postgres'`, which preserves the traditional `SSLRequest` behavior. You can also set the `PGSSLNEGOTIATION` environment variable.
+
 ## Channel binding
 
 If the PostgreSQL server offers SCRAM-SHA-256-PLUS (i.e. channel binding) for TLS/SSL connections, you can enable this as follows:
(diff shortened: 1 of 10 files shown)


Shion1305:
Hi @brianc @hjr3 👋 Thank you for maintaining this package!

Just a friendly heads up on this PR. I need my app to support a Postgres instance that requires direct SSL negotiation, so this would be really helpful on my end.

I’d really appreciate it if you could take a look when you have a chance. Happy to adjust the implementation if there’s a better direction for node-postgres.

brianc:
This is really nice & comprehensive. Also, doesn't look like you auto-generated the entire PR description w/ a wall of chatbot text which is a nice bonus!  It'd be dope to have a version of postgres in CI that has direct ssl negotation enabled so we have an end to end test there, but historically testing ssl permutations has been difficult to automate, and I appreciate the nice unit test coverage. I'm good to ship this, ty!

sehrope:
> It'd be dope to have a version of postgres in CI that has direct ssl negotation enabled so we have an end to end test there ...

https://github.com/brianc/node-postgres/pull/3693 :smile:

PR #3691, opened 2026-06-17
Stop cstring reading on any falsy value

brianc:
I got a massive, obviously LLM generated security vuln report generated about a potential MITM attack where if you connect pg to a compromised backend or a malicious MITM and the cstring buffer does not have a null terminator anywhere in it it will cause your process to infinite loop.  Pretty unlikely to happen in practice and honestly you're probably in more trouble than just that if your db connection is being tampered with maliciously, but I figure its still worth fixing since the fix itself is very straight forward.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ dist
 /.eslintcache
 .vscode/
 manually-test-on-heroku.js
+tsconfig.tsbuildinfo
--- a/packages/pg-protocol/src/buffer-reader.ts
+++ b/packages/pg-protocol/src/buffer-reader.ts
@@ -45,7 +45,7 @@ export class BufferReader {
     const start = this.offset
     let end = start
     // eslint-disable-next-line no-empty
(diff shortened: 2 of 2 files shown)


sehrope:
So the change here is to stop when the buffer gets `undefined` trying to read past it's max rather than looping forever? LGTM.

Though now I'm also curious if there's other spots where we're reading past the edge of the message or field limits. Will take a peek...

brianc:
> Though now I'm also curious if there's other spots where we're reading past the edge of the message or field limits. Will take a peek...

if memory serves this is the only place. 

as an aside: the empty while loop was a performance "hack" I did a super super long time ago. I forget where I copied it from - maybe the old version of the node mysql driver back in 2009?  Anyways I was curious if it was still the most efficent way to read a cstring out of the result buffer - because this code gets called in the *extremely hot path* I wanted to test it. I thought maybe `Buffer.indexOf` would be better now - but turns out this is still 10% faster than using `Buffer.indexOf` 🤷

PR #3690, opened 2026-06-17
Add docs for supported PostgreSQL versions

vfil:
Adds docs for supported PostgreSQL versions which closes #1529

PR #3693, opened 2026-06-18
Expand CI matrix of PG versions and add direct SSL test

sehrope:
First commit removes the `os` block from the GHA matrix. It's not actually used as we always run on `ubuntu-latest`. I think it may be left over from when we thought we were running on Windows / MacOS but actually were not.

Second updates the CI matrix to test against multiple PG versions. The original matrix of nodes versions is only tested against v18. But the latest node version is tested against v12 to v18. Uses some YAML anchors so we only have to update the latest node / PG versions in one place too.

image

RE: #3690, figure if we're going to say the driver supports older versions we should test them out too. It doesn't go all the way back till v8.4 though as the container images for `ghcr.io/railwayapp-templates/postgres-ssl` only go to v13. We'd have to switch back to the "official" docker hub postgres images without SSL for older versions. Might be worth doing but left it out for now (I think we could do it by just having the image name, and not just the version, in the matrix).

Finally last commit adds a test for direct SSL negotiation. It only runs in the pure JS driver against server v17+ (the min version its available). The test checks `client.connection.stream.alpnProtocol` to verify how we're connected.

RE: the server matrix, a while back I got a JSON endpoint added to the postgresql website that lists out the latest server versions in a machine readable JSON response: https://www.postgresql.org/versions.json Was going to make this all dynamic based on that version list (i.e. something like pick versions >= X) but I like it being explicit in the CI matrix. Makes it a bit easier to reason about what's actually being tested and we'd run into issues if the latest version does not have working containers yet. So best to keep it manual.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,8 +44,6 @@ jobs:
           - '22'
           - '24'
           - '26'
-        os:
-          - ubuntu-latest
     name: Node.js ${{ matrix.node }}
     runs-on: ubuntu-latest
     env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,26 +25,39 @@ jobs:
     needs: lint
     services:
       postgres:
-        image: ghcr.io/railwayapp-templates/postgres-ssl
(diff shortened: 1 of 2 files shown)


brianc:
ohhh yes this is amazing!! TY! ❤️

PR #3694, opened 2026-06-19
Add SASL OAUTHBEARER support with application-supplied tokens

### LINKED ISSUE #3687: Feature Request: SASL OAUTHBEARER support for PostgreSQL 18
PostgreSQL 18 introduced native OAuth 2.0 authentication via the SASL OAUTHBEARER mechanism ([RFC 7628](https://datatracker.ietf.org/doc/html/rfc7628)). `pg` currently has no support for this mechanism, which prevents connections to any PostgreSQL 18 instance configured with `method=oauth` in `pg_hba.conf`.

### Background

PG18 adds `OAUTHBEARER` as a third SASL mechanism alongside the existing `SCRAM-SHA-256` and `SCRAM-SHA-256-PLUS`. When a client connects to an OAuth-configured server, the server advertises `OAUTHBEARER` in its `AuthenticationSASL` message. The client must then conduct a multi-step SASL exchange, either:
- Running a full OAuth flow (e.g. Device Authorization Grant, RFC 8628) and presenting the resulting bearer token, or
- Presenting a pre-fetched bearer token directly in the initial SASL message

Relevant PG18 docs: https://www.postgresql.org/docs/current/auth-oauth.html

### Why the existing password callback isn't sufficient

The dynamic `password` callback added for AWS RDS token auth gets conceptually close, but isn't a workaround here. Even if a valid bearer token were returned from the callback, `pg` would attempt to use it in a SCRAM exchange. The server expects OAUTHBEARER SASL messages instead and will reject the connection at the protocol level.

### What's needed

- Recognition of `OAUTHBEARER` in the server's advertised SASL mechanism list
- Implementation of the SASL OAUTHBEARER message exchange
- New connection parameters for `oauth_issuer` and `oauth_client_id` (mirroring libpq)
- A hook or callback for the caller to supply a bearer token (fetched from user's IdP)

### References

- [RFC 7628 — SASL Mechanisms for OAuth](https://datatracker.ietf.org/doc/html/rfc7628)
- [RFC 8628 — OAuth 2.0 Device Authorization Grant](https://datatracker.ietf.org/doc/html/rfc8628)
- [PG18 OAuth auth docs](https://www.postgresql.org/docs/current/auth-oauth.html)
- [PG18 libpq OAuth client docs](https://www.postgresql.org/docs/current/libpq-oauth.html)
- [PG18 SASL protocol details](https://www.postgresql.org/docs/current/sasl-authentication.html)

proddata:
This adds support for PostgreSQL's SASL `OAUTHBEARER` authentication mechanism when the application supplies a bearer token directly or through a callback.

The implementation intentionally keeps OAuth provider interaction outside of `pg`. OAuth bearer token acquisition depends heavily on the IdP, grant type, scopes, refresh behavior, caching policy, and user interaction model. Instead of embedding that logic in the driver, this change gives applications a focused hook for supplying a valid token while `pg` handles the PostgreSQL SASL protocol exchange.

## Changes

- Add an `oauthBearerToken` client option that accepts either:
  - a token string
  - a sync or async callback returning a token string
- Prefer `OAUTHBEARER` when the server advertises it and `oauthBearerToken` is configured.
- Implement the OAUTHBEARER initial response and server challenge response handling.
- Add a generic `sendSASLResponseMessage` protocol serializer while keeping `sendSCRAMClientFinalMessage` as an alias for SCRAM compatibility.
- Keep `oauthBearerToken` non-enumerable on client, connection parameters, and pool options so token values do not appear in common inspection/serialization output.
- Document the new option and clarify that applications are responsible for fetching and caching OAuth bearer tokens.
- Add unit coverage for OAUTHBEARER mechanism selection, token callbacks, error paths, and response packets.

## Rationale

PostgreSQL 18 adds native OAuth authentication using the SASL `OAUTHBEARER` mechanism. Returning a bearer token from the existing `password` callback is not sufficient because the server expects OAUTHBEARER SASL messages, not a SCRAM exchange.

This change implements the PostgreSQL protocol support needed for callers that already have, or can fetch, a bearer token. Keeping token acquisition in user code keeps the driver small and avoids baking IdP-specific OAuth behavior into `pg`.

## Scope

This does not implement a full OAuth authorization flow, device authorization grant, token discovery, or token refresh logic inside `pg`.

Applications can build those flows externally and provide the resulting token through `oauthBearerToken`.

This support applies to the pure JavaScript client. `pg.native` delegates authentication to libpq and does not currently use `oauthBearerToken`; native OAuth support would require libpq OAuth configuration or new node-libpq auth hook bindings.

Partially resolves https://github.com/brianc/node-postgres/issues/3687

I am not conviced that fully following libpq (oauth_issuer and oauth_client_id) is proper for a low level driver like node-pg. However open for suggestions.

co-authored with codex

### DIFF
--- a/.github/workflows/oauth-integration.yml
+++ b/.github/workflows/oauth-integration.yml
@@ -1,0 +1,86 @@
+name: OAuth integration
+
+on:
+  pull_request:
+    paths:
+      - packages/pg/lib/client.js
+      - packages/pg/lib/connection-parameters.js
+      - packages/pg/lib/connection.js
+      - packages/pg/lib/crypto/sasl.js
+      - packages/pg-protocol/src/serializer.ts
+      - packages/pg/test/integration/client/oauth-tests.js
+      - packages/pg/test/oauth-postgres/**
+      - .github/workflows/oauth-integration.yml
+  push:
+    branches: [master]
+    paths:
+      - packages/pg/lib/client.js
+      - packages/pg/lib/connection-parameters.js
+      - packages/pg/lib/connection.js
+      - packages/pg/lib/crypto/sasl.js
+      - packages/pg-protocol/src/serializer.ts
+      - packages/pg/test/integration/client/oauth-tests.js
+      - packages/pg/test/oauth-postgres/**
+      - .github/workflows/oauth-integration.yml
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    env:
+      OAUTH_TEST_PGHOST: localhost
+      OAUTH_TEST_PGPORT: 55433
+      OAUTH_TEST_PGUSER: oauth_test
+      OAUTH_TEST_PGDATABASE: oauth_test
+      OAUTH_TEST_TOKEN: node-postgres-test-token
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - name: Setup node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: yarn
+      - run: yarn install --frozen-lockfile
+      - run: yarn build
+      - uses: docker/setup-buildx-action@v3
+      - name: Build OAuth PostgreSQL image
+        uses: docker/build-push-action@v6
+        with:
+          context: packages/pg/test/oauth-postgres
+          load: true
+          tags: node-postgres-oauth
+          cache-from: type=gha,scope=oauth-postgres
+          cache-to: type=gha,mode=max,scope=oauth-postgres
+      - name: Start OAuth PostgreSQL
+        run: |
+          docker run --detach --rm \
+            --name node-postgres-oauth \
+            --env POSTGRES_USER=postgres \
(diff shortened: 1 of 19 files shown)


brianc (approved):
This looks good to me. It would be nice if there was a way to configure a postgres instance w/ OAUTH authentication in CI so we could do an end-to-end integration test, but I realize configuring postgres in docker is somewhat non-trivial. Maybe ask an LLM to give that a shot? Otherwise this looks p straight forward and mergeable...ty!

proddata:
> This looks good to me. It would be nice if there was a way to configure a postgres instance w/ OAUTH authentication in CI so we could do an end-to-end integration test, but I realize configuring postgres in docker is somewhat non-trivial. Maybe ask an LLM to give that a shot? Otherwise this looks p straight forward and mergeable...ty!

I don't think there is a Postgres image with a bundled OAUTH validator module (yet). While Postgres does have some validators in the repo for tests https://github.com/postgres/postgres/tree/master/src/test/modules/oauth_validator, I think builidng Postgres for a test might be a bit much. So an approach would be to have some dummy validator and build it during testing and still use the same base Postgres docker image. However that likely means it should not be added to the CI, therefore added as separate workflow. Let me know what you think about it.

PR #3697, opened 2026-06-22
Remove unused “chunky” dev dependency and associated leftovers

charmander:
Unused since 766e48f34a5efaf52cfdc545230aaccfbb3d5107 (Update types & move some configs around).


brianc:
nice!

PR #3710, opened 2026-07-14
feat(types): expose QueryStream.Config

bobnil:
`QueryStreamConfig` is part of the public constructor API but is not currently exported.

This PR exposes it as `QueryStream.Config` using namespace merging, preserving the existing CommonJS export style.


brianc:
Thank you!

Issue #3707, opened 2026-07-06
Error during a `rows`-limited query permanently wedges the connection — `handleError` sends no Sync


drudolf:
- **pg:** 8.22.0
- **Node:** 24.14.1
- **PostgreSQL:** 18.4 (macOS; not server-specific — the wedge follows from the wire protocol)

## Summary

When a query runs with the `rows` option (portal suspension — `client.query({ text, rows: N })`),
an `ErrorResponse` from the server permanently wedges the connection: the query itself rejects
correctly, but `ReadyForQuery` never arrives, `client.readyForQuery` stays `false`, and every
subsequent query on that client queues forever.

Found while developing a wire-protocol bridge, where the message flow made the missing
frame visible.

## Reproduction

```js
const { Client } = require('pg')

async function main() {
  const client = new Client()
  await client.connect()

  // Errors mid-portal: division by zero on row 5, fetched in pages of 2.
  await client
    .query({ text: 'select 10 / (5 - i) as v from generate_series(1, 7) g(i)', rows: 2 })
    .catch((err) => console.log('query rejected:', err.message)) // "division by zero" — fine

  // The connection is now wedged: this never settles.
  const result = await Promise.race([
    client.query('select 1').then(() => 'follow-up resolved'),
    new Promise((r) => setTimeout(() => r('follow-up HUNG'), 5000)),
  ])
  console.log(result) // "follow-up HUNG"
}
main()
```

## Root cause

With `rows` set, `Query.prepare()` drives the extended protocol with **Flush**, not Sync:
`_getRows()` sends `Execute(rows=N)` + `connection.flush()`, and again on every
`PortalSuspended`; Sync is only sent from `handleCommandComplete` / `handleEmptyQuery`
(lib/query.js). When the server answers with `ErrorResponse` instead, it enters
ignore-till-sync — it discards everything until a Sync arrives and withholds
`ReadyForQuery`. `handleError` sends nothing, so the RFQ never comes and the client's
queue is stuck permanently.

Some history, because the code carries a fossil of the old behavior: `handleError` *did*
send Sync from at least 2014 (`// need to sync after error during a prepared statement`
— the comment is still in today's source). The Oct 2020 pipelining redesign (d31486fb,
then dd3ce616) moved Sync into `_getRows()`, pipelined right after Execute, and removed
the `handleError` call — correct for the normal prepared path, where a Sync is already
in flight by the time an error arrives. But the `rows` branch of `_getRows()` pipelines
[...]

PR #3712, opened 2026-07-18
fix: send Sync on error during rows-limited (portal) queries

houzhonglogic:
## Problem

When a query with the `rows` option (portal suspension mode) encounters an `ErrorResponse` from the server, `handleError` returns the error to the caller but **never sends `Sync`**. The server stays in an extended-query subtransaction waiting for `Sync`, `ReadyForQuery` never arrives, and the connection is permanently wedged — every subsequent query on that client queues forever.

This was reported in #3707: "Error during a `rows`-limited query permanently wedges the connection — `handleError` sends no Sync".

## Fix

`handleCommandComplete` already has the `connection.sync()` call for the `rows` path (line 93–94). `handleError` was simply missing the same check. This PR adds it.

## Prior art

The `pg-cursor` package handles this correctly in its own error path. The main `pg` packages `rows` feature uses a different internal path and was overlooked.

## Testing

`pg-cursor` already has error-handling tests that verify the connection recovers after a cursor error. The main packages `rows` path should have the same coverage — this fix brings it in line with cursor behaviour.

### DIFF
--- a/packages/pg/lib/query.js
+++ b/packages/pg/lib/query.js
@@ -125,6 +125,12 @@ class Query extends EventEmitter {
       err = this._canceledDueToError
       this._canceledDueToError = false
     }
+    // send sync to unstick the connection after an error during
+    // a rows-limited (portal suspension) query; without this,
+    // the server waits for Sync and ReadyForQuery never arrives.
+    if (this.rows) {
+      connection.sync()
+    }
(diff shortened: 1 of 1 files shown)


brianc:
I think CI/lint was broken. Can you rebase so we can run the tests here? Sorry for that

PR #3717, opened 2026-07-22
docs: document peer authentication over unix sockets

### LINKED ISSUE #1797: Connection Documentation: Peer Conncetion
Hey! Thanks for your work on a great project.

I went looking for docs on how to manage postgresql's peer authentication. It's mentioned in https://github.com/brianc/node-postgres/issues/613 which provides a solution. Unfortunately that doesn't appear in the documentation. I wondered if we could add a section:

## Password-less Authentication (Peer Auth)

Postgresql supports a number of authentication procedures which can be helpful, especially in development. One of those is Peer Authentication which allows the current OS user to authenticate with a local database. For this to work though, the connection need to be "local", specifically that means using a unix socket instead of a TCP socket.

To connect to the unix socket provided by postgresql you just need to use the path to the directory containing the socket as the host to your client configuration. Like so:

```js
const { Pool, Client } = require('pg')

const pool = new Pool({
  host: '/var/run/posthresql',
  database: 'mydb'
})

pool.query('SELECT NOW()', (err, res) => {
  console.log(err, res)
  pool.end()
})

const client = new Client({
  host: '/var/run/posthresql',
  database: 'mydb'
})
client.connect()

client.query('SELECT NOW()', (err, res) => {
  console.log(err, res)
  pool.end()
})
```

The client will then connect to the database `mydb` on the local postgresql server with the OS user that ran the program.

Hashim1999164:
## Summary
- Expands the unix socket connection docs to explain that `host` should be the socket directory.
- Adds a peer authentication (password-less local) example using `/var/run/postgresql`, matching the request in #1797.
- Fixes the existing unix socket snippet so the client is declared with `const`.

Closes #1797

## Test plan
- [ ] Preview the Connecting docs page and confirm the new Peer authentication section renders correctly
- [ ] Spot-check that the example matches local peer auth setups (`host` = socket dir, no password)

### DIFF
--- a/docs/pages/features/connecting.mdx
+++ b/docs/pages/features/connecting.mdx
@@ -116,18 +116,48 @@ const pool = new Pool({
 
 ### Unix Domain Sockets
 
-Connections to unix sockets can also be made. This can be useful on distros like Ubuntu, where authentication is managed via the socket connection instead of a password.
+Connections to unix sockets can also be made by setting `host` to the directory that contains the PostgreSQL socket (commonly `/var/run/postgresql` or `/tmp`).
 
 ```js
 import pg from 'pg'
 const { Client } = pg
-client = new Client({
+const client = new Client({
+  host: '/cloudsql/myproject:zone:mydb',
   user: 'username',
   password: 'password',
-  host: '/cloudsql/myproject:zone:mydb',
   database: 'database_name',
 })
 ```
+
+### Peer authentication (password-less)
+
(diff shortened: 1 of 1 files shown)


charmander:
I don’t think the extra section adds much correct value. How about just:

> Connections to unix sockets can also be made by setting `host` to the directory that contains the PostgreSQL socket (commonly `/run/postgresql` or `/tmp`). Unix sockets can support [peer authentication](https://www.postgresql.org/docs/current/auth-peer.html) without a password.

(there might be a better link, but it’s nothing node-postgres-specific)

Hashim1999164:
Simplified the section to the concise wording you suggested and removed the extra peer-auth section and examples.

charmander (approved):
Thanks! I’ll also pull the change from the original version of the PR to add the missing declaration for `client` as a separate commit.

Issue #2974, opened 2023-05-11
Parsing Error for Data Types w/ OID Outside of int4 Range


e-rich18:
Values that have a data type OID that is outside of the int4 range (i.e. > 2147483647), do not get parsed correctly. This seems especially true for array types, but it could effect any data type with such an OID.

Steps to reproduce:

1. Create a custom data type in your postgres database (i.e. `CREATE TYPE ...`) that has an OID outside of the int4 range (I don't know exactly how to accomplish that since it doesn't seem to be deterministic so it may take some trial and error)
2. Run a select query using `Client.query` that includes a column having that data type.

**Expected Behavior**
Column parses according to parsing rules for that data type (e.g. an array is parsed into a JS array)

**Actual Behavior**
The column is set to a plain string (e.g. `'{element,element}'` for an array).

Since OIDs are not typically set through manual intervention, this needs to be fixed in the library.

charmander:
https://github.com/brianc/node-postgres/blob/249182ea9fc1d0f3d73ecd5b67b5f2fde2ceab4f/packages/pg-protocol/src/parser.ts#L264

Even though https://www.postgresql.org/docs/current/protocol-message-formats.html specifies Int32, I guess this needs `uint32`?

charmander:
@gus-civitas In which context? Definitely not all of them.

PR #3728, opened 2026-08-03
fix(pg-protocol): read ParameterDescription type OIDs as unsigned

pratik-desgn:
While digging into #2974 (custom-type OIDs above the int4 range not parsing correctly), I found that path is actually already fixed — `parseField()` in `pg-protocol`'s parser reads `dataTypeID` with `reader.uint32()`, and there's already a regression test (`bigOidDescBuff`/`expectedBigOidMessage`) covering an OID over 2^31-1 in RowDescription.

`parseParameterDescriptionMessage()` never got the same fix though. It still reads each entry in `dataTypeIDs` with the signed `reader.int32()`, so a prepared statement parameter whose type OID exceeds 2^31-1 (custom extension types can get OIDs that high) comes back as a negative number instead of the real OID.

This mirrors the RowDescription fix exactly: swap `int32()` for `uint32()`, and add the same kind of big-OID regression test for ParameterDescription that already exists for RowDescription. Confirmed the new test fails on the old code (asserts a negative dataTypeID) and passes with the fix; full `pg-protocol` suite (78 tests) is green.

### DIFF
--- a/packages/pg-protocol/src/inbound-parser.test.ts
+++ b/packages/pg-protocol/src/inbound-parser.test.ts
@@ -161,6 +161,8 @@ const oneParameterDescBuf = buffers.parameterDescription([1111])
 
 const twoParameterDescBuf = buffers.parameterDescription([2222, 3333])
 
+const bigOidParameterDescBuf = buffers.parameterDescription([3000000003])
+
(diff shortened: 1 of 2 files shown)


brianc (approved):
wow nice really nice find & clean fix. ty!

PR #3732, opened 2026-08-07
fix: make prepared statement caches prototypeless

### LINKED ISSUE #3625: Make `Connection`’s `parsedStatements` a prototypeless object
## make `Connection`’s `parsedStatements` a prototypeless object

(= `namedQueries` on the native `Client`)

Maybe nobody has ever tried to prepare a query with `name: 'constructor'` (let alone the more obscure options on `Object.prototype`), and maybe nobody ever will, but it would be nice not to have to think about this anymore. Technically a breaking change for anyone who’s reaching into pg internals to do an inadvisable `client.connection.parsedStatements.hasOwnProperty(…)` check, for example.

_Originally posted by @charmander in https://github.com/brianc/node-postgres/discussions/3598#discussioncomment-15942909_

pri12ya871:
Closes #3625.

`Connection.parsedStatements` and the native client's `namedQueries` become `Object.create(null)`.

## Why it's user-visible

While writing tests for this I found it surfaces two different ways, depending on whether the query carries text:

**With text** — `submit()` finds a truthy `previous` and rejects the *first ever* use of the name:

```js
await client.query({ name: 'constructor', text: 'select 1', values: [] })
// Error: Prepared statements must be unique - 'constructor' was used for a different statement
```

**Without text** (the "execute the statement I prepared earlier" form) — that check is guarded on `this.text` and so is skipped, but `hasBeenParsed()` is truthy, so the Parse is omitted and the backend receives a Bind naming a statement it never prepared.

Same for `toString`, `valueOf`, `hasOwnProperty`, `isPrototypeOf` and `__proto__`. `__proto__` is the worst of them: the write in `client.js` is silently swallowed rather than stored, so the statement is never cached even once it has been parsed.

## Scope

Every read and write of both objects is a plain index access — `connection.parsedStatements[name]`, `client.namedQueries[self.name] = self.text` — across `lib/query.js`, `lib/client.js` and `lib/native/query.js`. Nothing calls `hasOwnProperty()` on them, spreads them, or passes them to `Object.keys`/`JSON.stringify`, and no type declaration references them, so dropping the prototype needs no other changes.

As noted in the issue this is technically breaking for anyone reaching into `client.connection.parsedStatements` and calling an inherited method on it, hence the `pg@9.0` target.

## Tests

Six cases in `packages/pg/test/unit/client/parsed-statements-prototype-tests.js`, covering both failure paths above, that the cache still works once a statement is parsed, and that genuine duplicate-name detection still errors. They're synchronous against a stubbed connection, so there's nothing to time out.

`make test-unit` passes. I could not run `test-integration` or `test-native` — I'm on Windows without a Postgres instance, and `libpq` won't build here — so the `namedQueries` half of the change is unexercised by anything I ran. Worth a look from someone who can run the native suite.

## Base branch

Based on `9.0` since that's where the milestone points. Happy to retarget to `master` if you'd rather — neither file has diverged, so it's a clean cherry-pick.

### DIFF
--- a/packages/pg/lib/connection.js
+++ b/packages/pg/lib/connection.js
@@ -22,7 +22,9 @@ class Connection extends EventEmitter {
 
     this._keepAlive = config.keepAlive
     this._keepAliveInitialDelayMillis = config.keepAliveInitialDelayMillis
-    this.parsedStatements = {}
+    // prototypeless so a statement named after an inherited property
+    // (`constructor`, `__proto__`, …) cannot be mistaken for one already parsed
+    this.parsedStatements = Object.create(null)
     this.ssl = config.ssl || false
     this._ending = false
     this._emitMessage = false
--- a/packages/pg/lib/native/client.js
+++ b/packages/pg/lib/native/client.js
@@ -55,8 +55,9 @@ const Client = (module.exports = function (config) {
   this.host = cp.host
   this.port = cp.port
 
-  // a hash to hold named queries
-  this.namedQueries = {}
+  // a hash to hold named queries, prototypeless so a query named after an inherited
+  // property (`constructor`, `__proto__`, …) cannot be mistaken for one already prepared
+  this.namedQueries = Object.create(null)
 })
 
 Client.Query = NativeQuery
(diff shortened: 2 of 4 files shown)


charmander:
This code isn’t ideal, and it’ll be faster for me to write it correctly than to review it. Since it’s AI-written, I don’t think anyone will miss out on a learning opportunity as a result.

PR #3734, opened 2026-08-09
fix(pg-protocol): decode binary result fields as raw bytes instead of utf-8

nigrosimone:
Results requested with `binary: true` are silently corrupted. No error is raised, the query
succeeds and returns wrong values:

```js
const client = new Client()
await client.connect()

await client.query({ text: 'SELECT $1::int AS a', values: [127], binary: true })
// { a: 127 }    correct

await client.query({ text: 'SELECT $1::int AS a', values: [128], binary: true })
// { a: 239 }    WRONG

await client.query({ text: 'SELECT $1::int AS a', values: [200], binary: true })
// { a: 239 }    WRONG

await client.query({ text: 'SELECT $1::int AS a', values: [-1], binary: true })
// { a: -272646673 }    WRONG
```

### DIFF
--- a/packages/pg-protocol/src/parser.ts
+++ b/packages/pg-protocol/src/parser.ts
@@ -83,6 +83,8 @@ export class Parser {
   private bufferOffset: number = 0
   private reader = new BufferReader()
   private mode: Mode
+  // Column formats from the most recent RowDescription, or null before any has been seen.
+  private rowDescriptionFormats: Mode[] | null = null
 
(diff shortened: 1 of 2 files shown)


charmander:
Duplicate of #3496. As I understand it, `binary: true` is not actually supported yet, just sort of happens to exist.

nigrosimone:
Duplicate https://github.com/brianc/node-postgres/pull/3496

PR #3736, opened 2026-08-09
fix: fail pipelined queries when the connection dies instead of hanging

nigrosimone:
When a pipelined batch is in flight and the connection goes away, the batch never call back, so every query queued behind it wait forever.

There is also a related bug in node-libpq, brianc/node-libpq#137 The fixes here are independent from that one, but both are needed to survive a lost connection.


brianc (approved):
a bit confusing on first glance but the PR description and tests make it readable. :) thanks a ton!

PR #3737, opened 2026-08-09
fix: reject portal based queries in pipeline mode instead of TypeError

nigrosimone:
A portal stays open across more round trips. In pipeline mode the queries behind it are written without waiting, so the server answer them out of that portal: the rows go to the wrong query, the next read fails with `portal "..." does not exist`, and with pg-cursor the process dies with an uncaught `TypeError: Cannot read properties of null (reading 'push')` inside `handleDataRow`.

Two cases, both reproducible every time on a client with `pipeline: true`:
```
without cursor 1,2,3   (correct)
with cursor:   TypeError: Cannot read properties of null (reading 'push')
               at Cursor.handleDataRow (pg-cursor/index.js:116)
```

### DIFF
--- a/packages/pg/lib/client.js
+++ b/packages/pg/lib/client.js
@@ -744,6 +744,24 @@ class Client extends EventEmitter {
       query._result._types = this._types
     }
 
+    // A query that keeps a portal open across round trips cannot share a pipelined connection: the
+    // queries written behind it are answered out of its portal, so rows land on the wrong query and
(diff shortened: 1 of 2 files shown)


brianc (approved):
this is cool - yeah its kinda unfortunate pipeline mode prevents some other stuff but it also makes sense. pipeline is "make the fast & straight forward things faster" and cursors or streams are "make the fiddly pieces easier for node to deal with"

PR #3739, opened 2026-08-10
feat: let pool.query pipeline on a connection that is already working

### LINKED ISSUE #3738: Pipeline mode for Pool
`pipeline: true` works on a `Client` but a `Pool` cannot use it. `pool.query()` checks a client out, runs one query on it and gives it back only when the result arrives, so there is never a second query on that connection to pipeline with. Passing `pipeline: true` to the `Pool` config changes nothing today.

postgres.js already does this automatically. It keeps the connections in three groups:
- idle
- busy but still accepting
- full
when nothing is idle it sends the query on a busy one. 

It limits itself with a depth cap (`max_pipeline`, default 100), with the socket backpressure, and it refuses to pipeline cursors.

The same is possible here and the public API almost does not change:

```js
const pool = new Pool({ max: 10, pipeline: true, maxPipeline: 100 })
await pool.query('SELECT $1::int', [1]) // same call, can go on a connection already working
```

nigrosimone:
close https://github.com/brianc/node-postgres/issues/3738

Passing `pipeline: true` to a `Pool` today changes nothing for `pool.query()`: every query checks out a connection for itself, so there is never a second query to pipeline behind it.

With this change `pool.query()` can send a query on a connection that is still waiting for results. The pool opens connection up to `max` first, then picks the one with the fewest queries in flight, capped by a new `maxPipeline` option (default 1) and by socket backpressure. `max` keeps its meaning of number of connections. `pool.connect()` still checks out a connection nobody else is using, so transactions are not affected. Off by default.

A connection with queries in flight is not idle: idle timeout, `maxUses`, `maxLifetimeSeconds` and `pool.end()` wait for the pending results before closing it.

usage:
```js
const pool = new Pool({ max: 10, pipeline: true, maxPipeline: 100 })
await pool.query('SELECT $1::int', [1]) // same call, can go on a connection already working
```

**Disclaimer**: made with the help of Opus 5 and a final check with Fable 5. I have done my best to verify that everything is correct and not AI slopped! tested on my side (patching node modules), seem works

### DIFF
--- a/docs/pages/apis/pool.mdx
+++ b/docs/pages/apis/pool.mdx
@@ -31,7 +31,7 @@ type Config = {
   // Maximum number of clients the pool should contain.
   // By default this is set to 10. There is some nuance to setting the maximum size of your pool.
   // See https://node-postgres.com/guides/pool-sizing for more information.
-  // With `pipeline: true` this is still the number of connections, but the number of queries
+  // With `maxPipeline` this is still the number of connections, but the number of queries
   // in flight can reach max * maxPipeline.
   max?: number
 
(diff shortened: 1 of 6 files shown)


nigrosimone:
A good point from @charmander https://github.com/brianc/node-postgres/issues/3738#issuecomment-5236785545

brianc:
this one is a big large & gnarly so i'll try to review it later this week!

nigrosimone:
I pushed a commit https://github.com/brianc/node-postgres/commit/aedb45ebb47ae6547c4bf13b9bfc3e7d79acd90d that separates the two: pipeline stays exactly as it is now, and the pool behaviour is driven by maxPipeline, which defaults to 1. With maxPipeline: 1 the pool works as today, one query per connection at a time. With a higher value pool.query() can send a query on a connection that is already working, and the clients the pool creates get pipeline: true. So new Pool({ pipeline: true }) keeps giving you pipelining clients and nothing else.

PR #3740, opened 2026-08-11
perf: Copy less when serializing `Bind` messages

charmander:
Takes about 15–30% less time on some parameter values in my measurements. Performance here could still be improved by a lot, but many other improvements also need interface changes, so I separated out this patch that was fully backwards-compatible.

### DIFF
--- a/packages/pg-protocol/src/serializer.ts
+++ b/packages/pg-protocol/src/serializer.ts
@@ -117,7 +117,8 @@ const enum ParamType {
 }
 
 const writeValues = function (values: any[], valueMapper: ValueMapper | undefined, formatsOffset: number): void {
-  for (let i = 0; i < values.length; i++) {
+  const len = values.length
+  for (let i = 0; i < len; i++) {
     const mappedVal = valueMapper ? valueMapper(values[i], i) : values[i]
     let formatByte = ParamType.STRING
 
--- a/packages/pg-protocol/src/serializer.ts
+++ b/packages/pg-protocol/src/serializer.ts
@@ -110,34 +110,35 @@ type BindOpts = {
   valueMapper?: ValueMapper
 }
 
-const paramWriter = new Writer()
-
 // make this a const enum so typescript will inline the value
 const enum ParamType {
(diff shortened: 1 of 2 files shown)


packages/pg-protocol/src/serializer.ts
```diff
@@ -110,34 +110,35 @@ type BindOpts = {
...
-
 // make this a const enum so typescript will inline the value
 const enum ParamType {
   STRING = 0,
   BINARY = 1,
 }
 
-const writeValues = function (values: any[], valueMapper?: ValueMapper): void {
+const writeValues = function (values: any[], valueMapper: ValueMapper | undefined, formatsOffset: number): void {
   for (let i = 0; i < values.length; i++) {
```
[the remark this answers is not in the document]
charmander: Not really a supported configuration, and I’m sure this issue is present elsewhere considering existing use of `allocUnsafe`, but I guess this is slightly more plausible for someone to do by accident compared to breaking the serializer with getters/proxies.

packages/pg-protocol/src/serializer.ts
```diff
@@ -110,34 +110,36 @@ type BindOpts = {
...
 // make this a const enum so typescript will inline the value
 const enum ParamType {
   STRING = 0,
   BINARY = 1,
 }
 
-const writeValues = function (values: any[], valueMapper?: ValueMapper): void {
-  for (let i = 0; i < values.length; i++) {
+const writeValues = function (values: any[], valueMapper: ValueMapper | undefined, formatsOffset: number): void {
+  const len = values.length
```
brianc, line 120: surprising this is still a perf improvement in 2026!

brianc (approved):
dang, nice!

packages/pg-protocol/src/serializer.ts
charmander: I’m not sure whether there’s a performance effect, but it was mainly for https://github.com/brianc/node-postgres/pull/3740#discussion_r3755338681.

PR #3742, opened 2026-08-11
fix: preserve row mode in native pipelines

zfaustk:
## What changed

Native pipeline entries now retain each query's array mode, and `pg-native` uses that mode when it builds the corresponding pipeline result.

The regression test mixes `rowMode: 'array'` and the default object mode in one connected pipeline batch, so it also verifies that row mode does not leak between queries.

## Why

The non-pipeline native path applies `NativeQuery._arrayMode` before reading a result, but the pipeline path dropped that per-query value. A `rowMode: 'array'` query therefore returned object rows when `pipeline: true`.

This is observable downstream in drizzle-team/drizzle-orm#6118: interpreted mapping produced `undefined` fields and JIT mapping threw `TypeError: object is not iterable`.

## Verification

- New integration test fails on `ff9d775` in native mode (object row returned instead of `[10]`).
- The focused pipeline integration file passes in both JavaScript and native modes after the fix.
- The original Drizzle reproduction passes for native direct queries plus interpreted and JIT mapping, with and without pipelining.
- Repository-wide ESLint passes; the three changed files pass Prettier check.
- The `pg` unit suite passes.

The canonical full `yarn test` matrix was not completed locally: its Yarn 1 workspace install caches unrelated multi-platform optional binaries beyond this executor's bounded disk budget. GitHub CI remains the authoritative full matrix.

## Compatibility

No public API is added or changed. The native pipeline path now matches the existing per-query `rowMode` behavior of the non-pipeline and JavaScript clients.

## AI assistance

Codex was used for investigation, implementation, and test execution. No human-review attestation is made.

### DIFF
--- a/packages/pg-native/index.js
+++ b/packages/pg-native/index.js
@@ -174,8 +174,8 @@ Client.prototype._stopReading = function () {
   this.pq.removeListener('readable', this._read)
 }
 
-Client.prototype._consumeQueryResults = function (pq) {
-  return buildResult(pq, this._types, this.arrayMode)
+Client.prototype._consumeQueryResults = function (pq, arrayMode = this.arrayMode) {
+  return buildResult(pq, this._types, arrayMode)
 }
 
 Client.prototype._emitResult = function (pq) {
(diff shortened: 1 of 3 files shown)


brianc (approved):
ohhhh very nice! good find 👍

pasha-vuiko:
is there any info about when it is going to be released to NPM?

brianc:
> is there any info about when it is going to be released to NPM?

sorry been out of the country & steady wifi access for a few days. I will release this monday morning CDT so I can be around on the computer & respond if the release has issues for some reason. 😄

pasha-vuiko:
> > is there any info about when it is going to be released to NPM?
> 
> sorry been out of the country & steady wifi access for a few days. I will release this monday morning CDT so I can be around on the computer & respond if the release has issues for some reason. 😄 

Great, thank you!

pasha-vuiko:
Hi, just a kind reminder about the release 🙂

PR #3743, opened 2026-08-12
Normalize query config without copying it

charmander:
Thoroughly copying an object with `create(getPrototypeOf(x), getOwnPropertyDescriptors(x))` just to read some properties from it is a little awkward and inefficient, and still has potential issues with setters for `callback`/`values`, for example. Follow-up to #3720 using a new backwards-incompatible function that can’t be imported from the outside, deprecating the original (now unused) function because it’s not a very good public API. When/if the callback API is removed, that’ll be a good opportunity to clean things up even more.

### DIFF
--- a/packages/pg/lib/query.js
+++ b/packages/pg/lib/query.js
@@ -3,16 +3,19 @@
 const { EventEmitter } = require('events')
 
 const Result = require('./result')
-const utils = require('./utils')
+const { prepareValue } = require('./utils')
+const { normalizeQueryConfig } = require('./internal/utils.js')
 
 class Query extends EventEmitter {
   constructor(config, values, callback) {
     super()
+    ;({
+      config,
+      text: this.text,
+      values: this.values,
+      callback: this.callback,
+    } = normalizeQueryConfig(config, values, callback))
 
-    config = utils.normalizeQueryConfig(config, values, callback)
-
-    this.text = config.text
-    this.values = config.values
     this.rows = config.rows
     this.types = config.types
     this.name = config.name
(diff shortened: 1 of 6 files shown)


packages/pg/lib/query.js
```diff
@@ -3,27 +3,29 @@
...
 
 const Result = require('./result')
-const utils = require('./utils')
+const { prepareValue } = require('./utils')
+const { normalizeQueryConfig } = require('./internal/utils.js')
 
 class Query extends EventEmitter {
   constructor(config, values, callback) {
     super()
+    ;({
```
charmander, line 12: (Prettier removes the newline before this for some reason, but I’m not going to dig into that.)

packages/pg/lib/utils.js
```diff
@@ -219,7 +219,10 @@ module.exports = {
     // by accident, eg: from calling values.map(utils.prepareValue)
     return prepareValue(value)
   },
-  normalizeQueryConfig,
+  normalizeQueryConfig: nodeUtils.deprecate(
+    normalizeQueryConfig,
+    'normalizeQueryConfig is an unused internal function and will be removed in pg@9.0'
+  ),
```
brianc, line 225: 👏

packages/pg/lib/query.js
brianc: yeah prettier is as prettier does 🙃

brianc (approved):
thanks friend! sorry for the delay 😢 - been in and out of town all summer so things are scattered for me

PR #3744, opened 2026-08-12
Report the parse error to pipelined queries that reused the failed statement

nigrosimone:
When a named statement fails to parse, pipelined queries already on the wire that skipped their own Parse for that name receive `prepared statement "x" does not exist` from the backend, an error about pg internals instead of the real one.

The client now marks those queries when the parse error arrives and hands them the original error, the same one they get without pipelining.

Regression test in `pipelining-tests.js`, it fails on master.


brianc (approved):
ohhh nice find. ty!

PR #3746, opened 2026-08-12
Support libpq's `channel_binding` and `require_auth` (and enforce them against every authentication request)

jawj:
This builds on my earlier PR, #3356, which added SCRAM-SHA-256-PLUS behind the `enableChannelBinding` option. As I noted then, that option let us *enable* channel binding but gave no way to *require* it. Being opt-in and non-standard, I suspect rather few users yet get the benefit of it.

This PR moves to using a libpq-style `channel_binding` parameter. It also introduces a libpq-style `require_auth` parameter, and makes sure both are enforced everywhere they need to be.

These changes make some connections more secure by default. They also provide users with the option to make connections even more secure, by specifying `channel_binding=require` or specific authentication modes, such as  `require_auth=scram-sha-256`, that limit the damage that can be done by an MITM attack. And they bring node-postgres's connection options into closer alignment with libpq.

Because `channel_binding=require` authenticates the server, it also silences the deprecation warning that `sslmode=prefer`, `sslmode=require` and `sslmode=verify-ca` currently emit (this matters to Neon users, for example, who see those warnings even though the weaker guarantees they warn about cannot be exploited when the channel is bound).

## `channel_binding`

- Takes libpq's three values, `disable`, `prefer` and `require`, from the client config, a connection string, or `PGCHANNELBINDING`.
- The default is now `prefer`: channel binding is used whenever the connection is over SSL and the server offers SCRAM-SHA-256-PLUS, and an unbound exchange is used otherwise. This matches libpq's default.
- `enableChannelBinding` is still honored, with `true` meaning `"prefer"` and `false` meaning `"disable"` (it also accepts the three levels), but is now documented as deprecated.
- `channel_binding: 'require'` needs SSL, so it is an error to combine it with SSL disabled — unless you are using the native (libpq) client, which negotiates SSL of its own accord and reports for itself if it ends up without.

## `require_auth`

Names the authentication method(s) the server is allowed to ask for, with libpq's semantics: a comma-separated allow-list, or a block-list in which every entry is negated with `!`, plus `none` for a connection where the server asks for nothing. It can be set in the config, a connection string, or `PGREQUIREAUTH` — which an explicit `require_auth: ''` overrides, as in libpq.

- The methods libpq recognizes are `password`, `md5`, `scram-sha-256`, `gss`, `sspi` and `oauth`, of which this client implements the first three. A setting that includes _only_ methods we cannot satisfy (e.g. `require_auth: 'gss'`) is an error at construction-time (but e.g. `require_auth: 'gss,md5'` is fine). The native client passes the setting to libpq and defers to it, so the full set is accepted there.
- `channel_binding: 'require'` implies `require_auth: 'scram-sha-256'`, since no other method supports channel binding. A `require_auth` value that rules out SCRAM contradicts that, so is also an error.

## Enforcing them

The kind of bug I filed [CVE-2025-49146](https://www.cve.org/CVERecord?id=CVE-2025-49146) for against pgjdbc is a risk here — where `channel_binding=require` was honored inside the SCRAM exchange but a server could simply ask for a plain password instead!

So the check lives in one place that every authentication request passes through, modelled on libpq's `check_expected_areq`:

- Requirements are checked before the client answers, so a refused request neither sends a password nor invokes a `password` callback.
- The same check runs on `AuthenticationOk`, and again where the connection becomes usable, so a server cannot evade a requirement by declaring the client authenticated without asking for anything — or by sending no authentication message at all. libpq accepts nothing but an authentication request at that point in its handshake, whereas this client listens for every message from the start, so it needs the second check.
- Authentication is answered once, whatever the two new parameters are set to. A server that asks again, having already said what it wanted and been given it, is refused.
- A refused authentication request now stays refused. A server can pipeline an entire successful login into one packet, and the messages following the refused request were still acted on: the connection callback was invoked a second time, this time reporting success, a `connect` event was emitted, and the client was left looking usable. Such a client now reports the failure once and refuses to answer anything further.
- Hashing an MD5 password and computing a SCRAM proof each take a turn of the event loop, and `Connection#end()` writes its Terminate before ending the stream, so both re-check that the connection has not been given up on before writing their answer.

Two related tightenings in the SCRAM code: the client counts itself channel-bound only once the server's signature has been verified, and only for SCRAM-SHA-256-PLUS, so an unbound exchange cannot satisfy `require`; and a server offering SCRAM-SHA-256-PLUS on a connection that is not encrypted is refused.

## Naming

Both parameters keep libpq's spelling (the same as `client_encoding` and `application_name` do, for example). Since both exist to refuse weak authentication, a camelCased `channelBinding` or `requireAuth` throws rather than being quietly ignored, as does a channel binding level that is not one of the supported three. Both options can also be set on a `Pool`, which passes its options to the clients it creates.

## Changing the default

Changing default behaviour was a concern last time round. `prefer` only ever upgrades a connection that is already over SSL, to a server that offers channel binding, and falls back to an unbound exchange otherwise, so there is nothing for a server to be surprised by. The one case where an outcome changes is a server whose certificate uses a signature algorithm the parser doesn't know (Ed448, for instance): that connection now fails rather than silently authenticating unbound. The same thing can happen in libpq. In either case, `channel_binding: 'disable'` restores the old outcome.

## Tests

- `packages/pg/test/unit/client/auth-flow-tests.js` drives a client through complete exchanges against a scripted SCRAM server using a real certificate, computing the server's replies from what the client actually sent, so it checks the gs2 header and the binding data rather than taking the client's word for either. This helps with @charmander's point on #3356 that the tests there would have passed with no implementation of channel binding at all. It covers each of the refusals above, including the pipelined login and the post-refusal races.
- `require-auth-tests.js` covers the parsing and the requirement check. The `ConnectionParameters` and `pg-connection-string` tests cover resolution from config, connection strings and the environment, precedence between them, and each error.
- The SCRAM integration tests now run over real SSL and assert whether the exchange was bound, and `PGTESTNOSSL` is no longer set in CI, so the SSL and channel binding tests actually run there (perhaps there is a good reason for setting `PGTESTNOSSL` in CI, though?). `packages/pg/script/test-server.sh` starts the same image CI uses under either podman or docker, for anyone who wants to run them locally.

Docs are in `docs/pages/features/ssl.mdx` and the `pg-connection-string` README, with a suggested `CHANGELOG` entry too.

DefinitelyTyped's `@types/pg` will need `channel_binding` and `require_auth` adding. I'll send a PR for that if and when these changes land.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ For richer information consult the commit log on github with referenced pull req
 
 We do not include break-fix version release in this file.
 
+## pg@8.24.0
+
+- Add support for the `channel_binding` connection parameter (in a connection string, the client config or `PGCHANNELBINDING`): `"disable"`, `"prefer"` or `"require"`, following libpq. Also change default from `disable` to `prefer`, so channel binding is used whenever the server offers it. The previous `enableChannelBinding` boolean option is retained but deprecated: `true` maps to `"prefer"`.
+- Also add support for the `require_auth`/`PGREQUIREAUTH`  connection parameter, which specifies which authentication method(s) the client will accept from the server.
+- Both parameters retain libpq's spelling. A camelCased `channelBinding` or `requireAuth` throw an error rather than leaving a security setting silently ignored, and an unrecognized channel binding value is refused the same way.
+- These requirements are enforced against every authentication request, not only SCRAM exchanges, so a server cannot evade `channel_binding=require` by requesting some other kind of authentication (a downgrade recorded against another driver as [CVE-2025-49146](https://www.cve.org/CVERecord?id=CVE-2025-49146)).
+- The native (libpq) client validates `channel_binding` and `require_auth` natively, against libpq's wider range of supported auth types.
+
 ## pg@8.23.0
 
 - Add support for query [`pipelineing`](https://github.com/brianc/node-postgres/pull/3652).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ For richer information consult the commit log on github with referenced pull req
 
 We do not include break-fix version release in this file.
 
-## pg@8.24.0
+## Unreleased
 
 - Add support for the `channel_binding` connection parameter (in a connection string, the client config or `PGCHANNELBINDING`): `"disable"`, `"prefer"` or `"require"`, following libpq. Also change default from `disable` to `prefer`, so channel binding is used whenever the server offers it. The previous `enableChannelBinding` boolean option is retained but deprecated: `true` maps to `"prefer"`.
 - Also add support for the `require_auth`/`PGREQUIREAUTH`  connection parameter, which specifies which authentication method(s) the client will accept from the server.
--- a/packages/pg-connection-string/index.d.ts
+++ b/packages/pg-connection-string/index.d.ts
@@ -2,9 +2,16 @@ import { ClientConfig } from 'pg'
 
 export function parse(connectionString: string, options?: Options): ConnectionOptions
 
+// Use of SCRAM channel binding, as libpq's channel_binding parameter defines it
+export type ChannelBinding = 'disable' | 'prefer' | 'require'
+
 export interface Options {
   // Use libpq semantics when interpreting the connection string
   useLibpqCompat?: boolean
+  // The channel binding setting held by the caller, for cases where it was not
+  // given in the connection string. A value of 'require' suppresses the sslmode
+  // deprecation warning, since the server is then authenticated by the binding.
+  channelBinding?: ChannelBinding
 }
 
 interface SSLConfig {
@@ -23,6 +30,10 @@ export interface ConnectionOptions {
   client_encoding?: string
   ssl?: boolean | string | SSLConfig
   sslnegotiation?: 'postgres' | 'direct'
+  channel_binding?: ChannelBinding
+  // The authentication method(s) the server may ask for, as libpq's require_auth
+  // parameter defines them: a comma-separated list, optionally negated with '!'
+  require_auth?: string
 
   application_name?: string
   fallback_application_name?: string
--- a/packages/pg/lib/channel-binding.js
+++ b/packages/pg/lib/channel-binding.js
@@ -1,0 +1,36 @@
+'use strict'
+
+// Support for libpq's channel_binding parameter, which says whether SCRAM authentication
+// has to be bound to the server's certificate:
+// https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-CHANNEL-BINDING
+
+const defaults = require('./defaults')
+
+const channelBindingLevels = ['disable', 'prefer', 'require']
+
+// Accepts the levels libpq's channel_binding parameter defines, plus the booleans that
+// pg's original enableChannelBinding option took. Any other non-string keeps its
+// historical truthiness, so previously working configs keep working. A string that is not
+// a level is refused rather than read as the weakest one that resembles it.
+const normalizeChannelBinding = function (value) {
+  if (typeof value !== 'string') {
+    return value ? 'prefer' : 'disable'
+  }
+  if (!channelBindingLevels.includes(value)) {
+    throw new Error(
+      `Invalid channel_binding value: "${value}". Valid values are "disable", "prefer" and "require" (or a boolean).`
+    )
+  }
+  return value
+}
+
+// channel_binding, being libpq's own spelling, wins over the older
+// enableChannelBinding option, then the environment, then the default.
+const resolveChannelBinding = function (channelBinding, enableChannelBinding) {
+  const value = [channelBinding, enableChannelBinding, process.env.PGCHANNELBINDING, defaults.channel_binding].find(
+    (candidate) => candidate !== undefined && candidate !== null
+  )
+  return normalizeChannelBinding(value)
+}
+
+module.exports = { channelBindingLevels, normalizeChannelBinding, resolveChannelBinding }
--- a/packages/pg/lib/channel-binding.js
+++ b/packages/pg/lib/channel-binding.js
@@ -4,33 +4,41 @@
 // has to be bound to the server's certificate:
 // https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-CHANNEL-BINDING
 
+const nodeUtils = require('util')
 const defaults = require('./defaults')
 
 const channelBindingLevels = ['disable', 'prefer', 'require']
 
-// Accepts the levels libpq's channel_binding parameter defines, plus the booleans that
-// pg's original enableChannelBinding option took. Any other non-string keeps its
-// historical truthiness, so previously working configs keep working. A string that is not
-// a level is refused rather than read as the weakest one that resembles it.
-const normalizeChannelBinding = function (value) {
-  if (typeof value !== 'string') {
-    return value ? 'prefer' : 'disable'
-  }
+const emitEnableChannelBindingDeprecationNotice = nodeUtils.deprecate(
+  () => {},
+  'enableChannelBinding is deprecated: instead, please set channel_binding to "disable", "prefer" or "require"'
+)
+
+function channelBindingFromDeprecatedBoolean(value) {
+  if (value === undefined) return  // undefined passes straight through, no warning
+  emitEnableChannelBindingDeprecationNotice();
+  // note: we pass through valid string values to avoid confusion, otherwise
+  // "disable" and "require" would both resolve as truthy and mean "prefer"
+  return channelBindingLevels.includes(value) ? value : (value ? 'prefer' : 'disable')
+}
+
+function validatedChannelBinding(value) {
   if (!channelBindingLevels.includes(value)) {
     throw new Error(
-      `Invalid channel_binding value: "${value}". Valid values are "disable", "prefer" and "require" (or a boolean).`
+      `Invalid channel_binding value: "${value}". Valid values are "disable", "prefer" and "require".`
     )
   }
   return value
 }
 
-// channel_binding, being libpq's own spelling, wins over the older
-// enableChannelBinding option, then the environment, then the default.
-const resolveChannelBinding = function (channelBinding, enableChannelBinding) {
-  const value = [channelBinding, enableChannelBinding, process.env.PGCHANNELBINDING, defaults.channel_binding].find(
-    (candidate) => candidate !== undefined && candidate !== null
-  )
-  return normalizeChannelBinding(value)
+function resolveChannelBinding(channelBinding, enableChannelBinding) {
+  const value =
+    channelBinding ??
+    channelBindingFromDeprecatedBoolean(enableChannelBinding) ??
+    process.env.PGCHANNELBINDING ??
+    defaults.channel_binding
+
+  return validatedChannelBinding(value)
 }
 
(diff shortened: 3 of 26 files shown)


jawj:
@hjr3 or @brianc: any thoughts on this?

charmander:
(For what it’s worth: I haven’t reviewed the diff yet, but the PR description sounds perfect.)

packages/pg/lib/channel-binding.js
```diff
@@ -0,0 +1,36 @@
...
+  }
+  return value
+}
+
+// channel_binding, being libpq's own spelling, wins over the older
+// enableChannelBinding option, then the environment, then the default.
+const resolveChannelBinding = function (channelBinding, enableChannelBinding) {
+  const value = [channelBinding, enableChannelBinding, process.env.PGCHANNELBINDING, defaults.channel_binding].find(
+    (candidate) => candidate !== undefined && candidate !== null
+  )
```
charmander:
```suggestion
  const value = channelBinding ?? enableChannelBinding ?? process.env.PGCHANNELBINDING ?? defaults.channel_binding
```

but shouldn’t `enableChannelBinding` be the only one with support for boolean values (and emit a deprecation warning when used)?

```javascript
const value = channelBinding ?? convertBoolean(enableChannelBinding) ?? process.env.PGCHANNELBINDING ?? defaults.channel_binding
if (!channelBindingLevels.includes(value)) {
  throw …
}
return value
```

(where `convertBoolean` returns `null`, `'disable'` + warning, or `'prefer'` + warning)

packages/pg/lib/client.js
```diff
@@ -116,6 +134,30 @@ class Client extends EventEmitter {
     this._connectionTimeoutMillis = c.connectionTimeoutMillis || 0
   }
 
+  get channelBinding() {
+    return this._channelBinding
+  }
+
+  // Changing the level after construction re-derives what the server has to do, so that
```
charmander: Does changing the level after construction need to be supported at all? If it’s just for strict backwards compatibility with `enableChannelBinding` (whose being settable was incidental, yeah?), I think it should only be possible via `set enableChannelBinding` and emit a deprecation warning.

packages/pg/test/unit/client/auth-flow-tests.js
```diff
@@ -0,0 +1,625 @@
...
+
+  return { client, stream, errors, callbacks, connects }
+}
+
+// Sends a message from the server to the client
+const send = function ({ client }, name, msg = {}) {
+  client.connection.emit(name, msg)
+}
+
+const until = async function (predicate, description) {
```
charmander: I know it’s a test, but uses of this busy wait look like they wouldn’t be too hard to replace with a promise resolved in `startClient` and a method provided by `MemoryStream`?

packages/pg/test/unit/client/auth-flow-tests.js
```diff
@@ -0,0 +1,625 @@
...
+
+suite.test('a cleartext password request is answered by default', async function () {
+  const connecting = startClient()
+
+  send(connecting, 'authenticationCleartextPassword')
+  await awaitPackets(connecting, 1)
+  send(connecting, 'authenticationOk')
+  await awaitQuiet()
+
+  assert.deepStrictEqual(connecting.errors, [])
```
charmander: And it looks like this kind of thing should be a flag to `startClient` that no errors are expected.

packages/pg/test/unit/client/auth-flow-tests.js
```diff
@@ -0,0 +1,625 @@
...
+  })
+
+  // The request is permitted, so the lookup begins...
+  send(connecting, 'authenticationCleartextPassword')
+  // ...but the same packet held a request that is not, and by the time the credential
+  // service answers there is nothing left to answer with.
+  send(connecting, 'authenticationMD5Password', { salt: Buffer.from([1, 2, 3, 4]) })
+  const error = await awaitError(connecting)
+  await until(() => release !== undefined, 'the password to be asked for')
+  release()
```
charmander: and `;(await release)()`, with `release` being made into a promise.

packages/pg/test/unit/connection-parameters/creation-tests.js
```diff
@@ -397,6 +397,302 @@ suite.test('sslnegotiation is read from PGSSLNEGOTIATION env var', function () {
...
+    database: 'postgres',
+    ssl: true,
+    channel_binding: 'require',
+  })
+  subject.getLibpqConnectionString(
+    assert.calls(function (err, pgCString) {
+      assert(!err)
+      assert.equal(
+        pgCString.indexOf("channel_binding='require'") !== -1,
+        true,
```
charmander:
```suggestion
      assert.ifError(err)
      assert(
        pgCString.includes("channel_binding='require'"),
```

CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ For richer information consult the commit log on github with referenced pull req
 
 We do not include break-fix version release in this file.
 
+## pg@8.24.0
```
charmander:
```suggestion
## Unreleased
```

charmander (commented):
(Partial review that doesn’t cover the core logic. Test refactoring suggestions needn’t block merge.)

Thanks for implementing all this!

packages/pg-connection-string/index.d.ts
```diff
@@ -2,9 +2,16 @@ import { ClientConfig } from 'pg'
...
+// Use of SCRAM channel binding, as libpq's channel_binding parameter defines it
+export type ChannelBinding = 'disable' | 'prefer' | 'require'
+
 export interface Options {
   // Use libpq semantics when interpreting the connection string
   useLibpqCompat?: boolean
+  // The channel binding setting held by the caller, for cases where it was not
+  // given in the connection string. A value of 'require' suppresses the sslmode
+  // deprecation warning, since the server is then authenticated by the binding.
+  channelBinding?: ChannelBinding
```
hjr3, line 14: We introduced `Options.useLibpqCompat` because we had no other way of signaling a change in the parsing behavior. My hope was to remove `Options` once we made the major version bump.

Setting `Options.channelBinding` is not changing parsing behavior but instead adding another way to modify the resulting config object.

If someone wants to add channel binding, they could parse and then add it to the resulting config object:

```js
const config = parse('...');
config.channel_binding = 'require';
```

If it is to prevent the warning, then:

```js
const url = new URL('pg:///?sslmode=require')
if (url.searchParams.has('channel_binding') === false) {
  url.searchParams.append('channel_binding', 'require')
}
const config = parse(url.toString());
```

hjr3 (commented):
Overall, this looks great. One minor piece of feedback on `Options.channelBinding` but this does not have to be blocking a merge.

packages/pg/lib/channel-binding.js
jawj: > shouldn’t `enableChannelBinding` be the only one with support for boolean values (and emit a deprecation warning when used)

Good call: both changes now made.

packages/pg/lib/client.js
jawj: I agree that it's not _necessary_, but it is already implemented.

Since properties like `host` and `ssl` can be changed after construction, is there a good reason to prevent `channel_binding` from being changed too?

packages/pg/test/unit/client/auth-flow-tests.js
jawj: Fair point — now refactored to do that.

packages/pg/test/unit/client/auth-flow-tests.js
jawj: Certainly this is a common pattern. For now, I've refactored it into a `waitForSuccess` function, for symmetry with `waitForError`.

packages/pg/test/unit/client/auth-flow-tests.js
jawj: The busy wait is removed here by the changes made above (we await a new local promise, `passwordPromise`).

packages/pg/test/unit/connection-parameters/creation-tests.js
jawj: Done.

packages/pg-connection-string/index.d.ts
jawj: The only use of this new `options.channelBinding` parameter is to suppress the warning currently emitted about the forthcoming changed behaviour of `prefer`, `require` and `verify-ca` when channel binding is being enforced.

I assume that once the version bump is made, this warning goes away, and thus so does the need to suppress it? So we won't need `options.channelBinding` any more, and this shouldn't stop you removing `Options`, right?

Or have I missed something?

packages/pg/lib/client.js
charmander: > Since properties like `host` and `ssl` can be changed after construction

I’m not sure how intended that is either… `Client` is awkwardly designed. But yes, there is a small good reason: this complexity :)

jawj: `set enableChannelBinding` already emits a deprecation warning, when it calls `channelBindingFromDeprecatedBoolean`.

And as long as we allow `set enableChannelBinding` for backwards-compatibility, the complexity (which amounts to calling `resolveAuthRequirement` again) doesn't go away.

So my preference is not to change this. But if that's a blocker for you, then I guess what I need to do is:

* Move the logic in `set channelBinding` to `set enableChannelBinding`
* Make `set channelBinding` throw, so that nobody thinks they're getting a more secure connection than they are

Can you confirm?

packages/pg/test/unit/client/auth-flow-tests.js
jawj: @charmander Would you consider this resolved?

packages/pg/lib/client.js
charmander: It’s not strictly a blocker for me, but here’s my case:

> And as long as we allow `set enableChannelBinding` for backwards-compatibility, the complexity (which amounts to calling `resolveAuthRequirement` again) doesn't go away.

Yes, but the intent is to remove that in the imminent pg@9.

> Move the logic in `set channelBinding` to `set enableChannelBinding`

:+1:

> Make `set channelBinding` throw, so that nobody thinks they're getting a more secure connection than they are

No need to throw explicitly, properties with only a getter are read-only by default.

jawj: OK: I've now made it so `channelBinding` can't be set after construction.

> No need to throw explicitly, properties with only a getter are read-only by default.

A getter-only property is read-only, but outside of strict mode it's a no-op rather than an error. Since it's a security issue if `client.channelBinding = "require"` does nothing (from user code not in strict mode), I've kept the explicitly-throwing setter.

jawj:
I think I've addressed all comments, so I'm hopeful this can now be merged. :)

jawj:
Since this PR effectively changes channel binding from opt-in to opt-out, it becomes more important to support all potential certificate hash types.

So I thought it was worth putting in a bit more work to (a) make the hash/digest type parsing in `signatureAlgorithmHashFromCertificate` pretty much exhaustive, and (b) use `node:crypto` as a fallback for `SubtleCrypto` when calculating the hash (since it supports more algorithms).

@charmander and @hjr3 — please do let me know if there are any more changes you'd like to see.

jawj:
SCRAM-SHA-256 upgrades MD5 and SHA-1 to SHA-256, indicating an intention to be at least as secure as SHA-256. My updated signature algorithm-parsing code already had MD2, MD4, MDC2 and SHA-0 as explicitly unsupported. My latest commit disables two additional algorithms: RIPEMD160 and GOST94.

Any remaining blockers to merging?

PR #3747, opened 2026-08-12
fix: support callback as second argument to `CloudflareSocket.write`

zfaustk:
`Connection.end()` calls `write(data, callback)`. `CloudflareSocket.write` previously treated the callback as an encoding, so it never ran.

This adds explicit overloads for `write(data, callback)` and `write(data, encoding, callback)`. The runtime dispatch stays in the existing `write` method; close and event behavior are unchanged.

Validation:

- focused `pg-cloudflare` runtime test: 3 passed;
- `yarn build`, `yarn lint`, Prettier, and `git diff --check`: passed;
- type fixture: valid overloads compile, while `write(data, callback, callback)` fails with TS2345.

The full PostgreSQL/libpq integration suite and a real Cloudflare Worker E2E were not run locally.

Prepared with Codex assistance.

### DIFF
--- a/packages/pg-esm-test/pg-cloudflare.test.js
+++ b/packages/pg-esm-test/pg-cloudflare.test.js
@@ -19,44 +19,17 @@ describe('pg-cloudflare', () => {
     assert.doesNotThrow(() => socket.end())
   })
 
-  it('should emit close when the underlying close call resolves', async () => {
-    const socket = new CloudflareSocket()
-    const underlyingSocket = {
-      closed: new Promise(() => {}),
-      close: () => Promise.resolve(),
-    }
-    socket._cfSocket = underlyingSocket
-    socket._addClosedHandler()
-
-    const close = new Promise((resolve) => socket.once('close', resolve))
-    socket.end()
-
-    await close
-    assert.equal(socket._cfSocket, null)
-  })
-
-  it('should support the write(data, callback) overload', async () => {
+  it('should call the write(data, callback) callback exactly once', async () => {
     const socket = new CloudflareSocket()
     socket._cfWriter = { write: () => Promise.resolve() }
 
-    await new Promise((resolve) => socket.write(Buffer.from('x'), resolve))
-  })
+    let callbackCount = 0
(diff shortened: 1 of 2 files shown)


charmander (changes requested):
Doesn’t look right. Maybe start with a PR for just the `write` overload and a corresponding test that exercises it, ideally typed correctly with an overload instead of accepting `(data, callback, callback)`.

zfaustk:
Thanks — I narrowed this to the `write` overload only.

The updated method now has explicit overloads for `write(data, callback)` and
`write(data, encoding, callback)`. The focused test checks that the two-argument
callback runs exactly once, and a TypeScript fixture confirms that
`write(data, callback, callback)` is rejected. I removed the close-event and
closed-promise changes from this PR.

packages/pg-esm-test/pg-cloudflare.test.js
```diff
@@ -18,4 +18,18 @@ describe('pg-cloudflare', () => {
...
+    socket._cfWriter = { write: () => Promise.resolve() }
+
+    let callbackCount = 0
+    socket.write(Buffer.from('x'), (error) => {
+      assert.ifError(error)
+      callbackCount++
+    })
+
+    await new Promise((resolve) => setImmediate(resolve))
+    assert.equal(callbackCount, 1)
```
charmander:
```suggestion
    const { promise, resolve } = Promise.withResolvers()
    let called = false
    socket.write(Buffer.from('x'), (error) => {
      assert.ifError(error)
      assert(!called)
      called = true
      resolve()
    })

    await promise
```

charmander (changes requested):

packages/pg-esm-test/pg-cloudflare.test.js
zfaustk: Updated — the test now waits for the callback directly, checks the error, and guards against a second call. I used `new Promise(...)` rather than `Promise.withResolvers()` because CI still runs Node 16, 18, and 20.

charmander (approved):

PR #3748, opened 2026-08-12
fix: add default error listeners

### LINKED ISSUE #3630: ## add default no-op `error` listener to pool and client
## add default no-op `error` listener to pool and client

Returning errors from all operations (which is already what happens after an `error` event) is probably enough as a default; no need to crash the process.

_Originally posted by @charmander in https://github.com/brianc/node-postgres/discussions/3598#discussioncomment-15904240_

zfaustk:
## Summary

- add a default no-op `error` listener to every `Client`, native-backed `Client`, and `Pool`
- keep normal EventEmitter semantics for user listeners, including the original `Error` object
- update pool guidance for the new pg@9 default

Fixes #3630.

## Why

Today an otherwise-unobserved background `error` event throws through Node's EventEmitter and can terminate the process. The operation-specific errors have already been returned, so pg@9 can provide a safer default while applications that add an `error` listener continue to observe and handle the same event.

This intentionally changes the behavior of unobserved `error` events and is scoped to the pg@9.0 milestone.

## Testing

- `packages/pg/test/unit/client/configuration-tests.js` (13 checks passed)
- focused `packages/pg-pool/test/events.js` error-listener cases (2 passed)
- public runtime RED-to-GREEN for `Pool`; actual JavaScript and native-backed `Client` constructors verified with the same `Error` reaching a user listener
- ESLint on all changed JavaScript files
- Prettier 3.0.3 on all changed files
- `node --check` on all changed JavaScript files and `git diff --check`

The full `yarn test` matrix was not run locally. The remaining database-backed tests in `packages/pg-pool/test/events.js` require a PostgreSQL service and stopped with `ECONNREFUSED` in this environment; upstream CI covers the Node/PostgreSQL matrix and native libpq suite.

## Authorship

Implemented and tested with Codex assistance. No human-review claim is made.

### DIFF
--- a/docs/pages/apis/pool.mdx
+++ b/docs/pages/apis/pool.mdx
@@ -240,7 +240,8 @@ The number of queued requests waiting on a client when all clients are checked o
 Whenever the pool establishes a new client connection to the PostgreSQL backend it will emit the `connect` event with the newly connected client.
 
 <Alert>
-  The event listener does not wait for promises or async functions. If you want to run setup commands on each new client, use the `onConnect` option. (See documentation above.)
+  The event listener does not wait for promises or async functions. If you want to run setup commands on each new
+  client, use the `onConnect` option. (See documentation above.)
 </Alert>
 
 ### acquire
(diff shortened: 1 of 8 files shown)


charmander:
bad tests, wrong target branch. i’ll just do it when the time comes

PR #3749, opened 2026-08-12
fix: use pg-protocol CommonJS entrypoint

### LINKED ISSUE #3700: pg 8.22.0 CJS entry can resolve pg-protocol as ESM under Cloudflare/Vite worker tests
### Summary

When `pg@8.22.0` is bundled by the `@cloudflare/vitest-pool-workers` / Vite worker-test pipeline in a pnpm workspace, the CommonJS entry can fail while loading `pg-protocol` from `pg/lib/connection.js`:

```text
SyntaxError: Cannot use import statement outside a module
  at node_modules/.pnpm/pg@8.22.0/node_modules/pg/lib/connection.js?mf_vitest_no_cjs_esm_shim:5:30
  at node_modules/.pnpm/pg@8.22.0/node_modules/pg/lib/client.js?mf_vitest_no_cjs_esm_shim:10:20
  at node_modules/.pnpm/pg@8.22.0/node_modules/pg/lib/index.js
```

The failing line is the CommonJS package-name require:

```js
const { parse, serialize } = require("pg-protocol")
```

In this environment, that dependency can be resolved/transformed as an ES module, then loaded from a CJS wrapper.

### Environment

- `pg`: `8.22.0`
- `pg-protocol`: `1.15.0`
- package manager: `pnpm`
- test/bundler stack: `@cloudflare/vitest-pool-workers` + Vite + Miniflare/workerd
- worker compatibility flag: `nodejs_compat`

### Local workaround

We currently patch `pg` to force the CJS files to load the CJS dependency files directly from pnpm's sibling layout:

```diff
-const { parse, serialize } = require("pg-protocol")
+const { parse, serialize } = require("../../pg-protocol/dist/index.js")

-const Pool = require("pg-pool")
+const Pool = require("../../pg-pool/index.js")

-const parse = require("pg-connection-string").parse
+const parse = require("../../pg-connection-string/index.js").parse
```

That patch is not a good general solution because it depends on pnpm's installed package layout.

### Ask

Would `pg` consider an upstream change that makes the CommonJS entry robust in bundler/worker environments, for example by ensuring the CJS entry always resolves CJS-compatible dependency entrypoints for `pg-protocol`, `pg-pool`, and `pg-connection-string`?

Happy to provide more details or test a candidate fix.

zfaustk:
Fixes #3700.

## Problem

When the Cloudflare Vitest/Vite worker pipeline inlines both `pg` and `pg-protocol`, the CommonJS `pg` entry can resolve the bare `pg-protocol` import through its ESM condition. Loading that ESM wrapper from `pg`'s CommonJS files then fails before any tests run:

```text
SyntaxError: Cannot use import statement outside a module
at pg/lib/connection.js?mf_vitest_no_cjs_esm_shim:5:30
```

I reproduced this with the standalone fixture from #3700 using released `pg@8.23.0`, `pg-protocol@1.16.0`, `@cloudflare/vitest-pool-workers@0.16.18`, and Vitest 4.1.9.

## Fix

- load `pg-protocol/dist/index.js` explicitly from `pg`'s two CommonJS entry files;
- add a unit regression that verifies the subpath is exported and both call sites keep using it.

This remains package-manager independent: `pg-protocol` exposes `./dist/*.js` through its public `exports` map, and `pg` currently depends on `pg-protocol ^1.16.0`.

## Tests

- published standalone worker fixture: fails before the patch, passes 1/1 after it;
- `make test-unit` from `packages/pg`;
- `yarn lint`;
- `yarn build`;
- Prettier check for all changed files;
- `git diff --check`.

The full PostgreSQL/Node matrix was not run locally; it is left to CI. The repository's existing Cloudflare integration test also gets past module collection with the worker inline settings after the patch, then stops at this executor's unavailable PostgreSQL endpoint.

Implementation and test preparation were assisted by Codex. The reproduction and all reported checks above were run against this branch.

### DIFF
--- a/packages/pg/lib/connection.js
+++ b/packages/pg/lib/connection.js
@@ -2,7 +2,7 @@
 
 const EventEmitter = require('events').EventEmitter
 
-const { parse, serialize } = require('pg-protocol')
+const { parse, serialize } = require('pg-protocol/dist/index.js')
 const stream = require('./stream')
 const { getStream } = stream
 
--- a/packages/pg/lib/index.js
+++ b/packages/pg/lib/index.js
@@ -7,7 +7,7 @@ const Result = require('./result')
 const utils = require('./utils')
 const Pool = require('pg-pool')
 const TypeOverrides = require('./type-overrides')
-const { DatabaseError } = require('pg-protocol')
+const { DatabaseError } = require('pg-protocol/dist/index.js')
 const { escapeIdentifier, escapeLiteral } = require('./utils')
 
 const poolFactory = (Client) => {
--- a/packages/pg/test/unit/protocol-import-tests.js
+++ b/packages/pg/test/unit/protocol-import-tests.js
@@ -1,0 +1,16 @@
+'use strict'
+
+const assert = require('assert')
+const fs = require('fs')
+const helper = require('./test-helper')
+const suite = new helper.Suite()
+const test = suite.test.bind(suite)
+
+const connectionSource = fs.readFileSync(require.resolve('../../lib/connection'), 'utf8')
+const indexSource = fs.readFileSync(require.resolve('../../lib'), 'utf8')
+
+test('uses the exported pg-protocol CommonJS entrypoint', function () {
+  require.resolve('pg-protocol/dist/index.js')
+  assert.match(connectionSource, /require\('pg-protocol\/dist\/index\.js'\)/)
+  assert.match(indexSource, /require\('pg-protocol\/dist\/index\.js'\)/)
+})


charmander:
This seems like a build misconfiguration, not a pg issue that should be fixed by changing the imports in this way.

zfaustk:
Thanks — agreed. The reproduction showed the worker build selecting the ESM
condition for a CommonJS caller, but changing `pg`'s imports fixes that at the
wrong layer. I'll take this out of review and investigate the Vite/Workers
configuration instead.

PR #3752, opened 2026-08-17
fix(pg-cloudflare): prevent Client.end() from hanging

### LINKED ISSUE #3152: Cloudflare Client.end() hangs if socket did not successfully connect
There is a particular scenario where calling `Client.end` will result in a hanging promise when using Cloudflare's socket implementation and the socket does not successfully connect. I've found that this can be reproduced consistently by specifying an incorrect port in the connection string. This only occurs when using the Cloudflare runtime and `CloudflareSocket`.

```ts
// use a connection string with an unreachable port here
const client = new Client({ connectionString });
try {
  await client.connect();
  await client.query(`SELECT 1`);
} finally {
  // ⬇️ hangs here
  await client.end();
}
```

I believe I've found the root causes to it. The `Socket.closed` promise on the Cloudflare socket as used [here](https://github.com/brianc/node-postgres/blob/2a8efbee09a284be12748ed3962bc9b816965e36/packages/pg-cloudflare/src/index.ts#L137) does not resolve if the connection never successfully opened to begin with. `Connection` expects the event [here](https://github.com/brianc/node-postgres/blob/2a8efbee09a284be12748ed3962bc9b816965e36/packages/pg/lib/connection.js#L62:L64) to bubble out the `end` event to `Client` [here](https://github.com/brianc/node-postgres/blob/2a8efbee09a284be12748ed3962bc9b816965e36/packages/pg/lib/client.js#L618:L624) and resolve the promise or invoke the callback.

Additionally, the `stream.end` callback is also not invoked in `Connection.end`. That's from the `CloudflareSocket.write` function expecting the callback to always be in the third argument [here](https://github.com/brianc/node-postgres/blob/2a8efbee09a284be12748ed3962bc9b816965e36/packages/pg-cloudflare/src/index.ts#L86), where `Connection.end` passes it as the second argument [here](https://github.com/brianc/node-postgres/blob/2a8efbee09a284be12748ed3962bc9b816965e36/packages/pg/lib/connection.js#L197).

I made the following naive patches that resolve the issue in my case, though I'm not sure it's the best resolution. I'm expecting that successfully connected sockets would emit the `close` event twice with this patch.

```diff
diff --git a/node_modules/pg-cloudflare/dist/index.js b/node_modules/pg-cloudflare/dist/index.js
index e2db57e..f94b723 100644
--- a/node_modules/pg-cloudflare/dist/index.js
+++ b/node_modules/pg-cloudflare/dist/index.js
@@ -89,7 +89,8 @@ export class CloudflareSocket extends EventEmitter {
     end(data = Buffer.alloc(0), encoding = 'utf8', callback = () => { }) {
         log('ending CF socket');
         this.write(data, encoding, (err) => {
-            this._cfSocket.close();
+            // the closed Promise does not resolve if the connection was never open to begin with
+            this._cfSocket.close().then(() => this.emit('close'));
             if (callback)
                 callback(err);
         });
```

```diff
diff --git a/node_modules/pg/lib/connection.js b/node_modules/pg/lib/connection.js
index af4b8f1..af03e05 100644
--- a/node_modules/pg/lib/connection.js
+++ b/node_modules/pg/lib/connection.js
@@ -194,7 +194,7 @@ class Connection extends EventEmitter {
       this.stream.end()
       return
     }
-    return this.stream.write(endBuffer, () => {
+    return this.stream.write(endBuffer, undefined, () => {
       this.stream.end()
     })
   }
```

pratik-desgn:
Fixes https://github.com/brianc/node-postgres/issues/3152

## Summary

- complete Cloudflare shutdown when Socket.close() resolves but Socket.closed never settles
- emit close once, with an identity guard preventing a duplicate event from the normal closed handler
- add a regression test for a socket that never settles its closed promise

The callback overload used by Connection.end() is already present in the current base branch via https://github.com/brianc/node-postgres/pull/3747.

## Validation

- yarn lint
- yarn workspace pg-cloudflare build
- yarn workspace pg-esm-test test --test-name-pattern=pg-cloudflare (12 passed)
- GitHub CI: lint and all 11 Node/PostgreSQL matrix jobs passed

The PR is limited to pg-cloudflare shutdown behavior and its focused ESM test.

### DIFF
--- a/packages/pg-cloudflare/src/index.ts
+++ b/packages/pg-cloudflare/src/index.ts
@@ -107,7 +107,7 @@ export class CloudflareSocket extends EventEmitter {
   end(data = Buffer.alloc(0), encoding: BufferEncoding = 'utf8', callback: (...args: unknown[]) => void = () => {}) {
     log('ending CF socket')
     this.write(data, encoding, (err) => {
-      this._cfSocket!.close()
+      this._cfSocket?.close()
       if (callback) callback(err)
     })
     return this
--- a/packages/pg-cloudflare/src/index.ts
+++ b/packages/pg-cloudflare/src/index.ts
@@ -84,9 +84,13 @@ export class CloudflareSocket extends EventEmitter {
 
   write(
     data: Uint8Array | string,
-    encoding: BufferEncoding = 'utf8',
+    encoding: BufferEncoding | ((...args: unknown[]) => void) = 'utf8',
     callback: (...args: unknown[]) => void = () => {}
   ) {
+    if (typeof encoding === 'function') {
+      callback = encoding
+      encoding = 'utf8'
+    }
     if (data.length === 0) return callback()
     if (typeof data === 'string') data = Buffer.from(data, encoding)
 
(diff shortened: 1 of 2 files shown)


brianc:
ya this looks p self-contained. the promise-y conditional code is hard to read but it definitely ends up that way sometimes w/ networking code in node unfortunately.

PR #3753, opened 2026-08-18
perf: cache field parsers and prebuilt rows per query shape

Iaotle:
## What

`Result.addFields` currently calls `getTypeParser` per column and rebuilds the prebuilt empty row object on every query. Real workloads repeat a small set of result-set shapes, so this work is almost always identical to what the previous query already computed. In a production Node service we profiled (~700–800 queries/s through pg + an ORM), `addFields` + `getTypeParser` self-time added up to ~9.6% of the main thread's on-CPU time — pure per-query overhead, since most queries return a single row.

This PR caches the parser array and prebuilt row per query shape:

- Keyed on the types object in use (a `WeakMap`), then on a shape signature of `name`/`dataTypeID`/`format` per field (NUL-separated — Postgres identifiers cannot contain NUL).
- Invalidated by a version counter that `TypeOverrides.setTypeParser` bumps, so `client.setTypeParser(...)` between queries applies to later queries exactly as before (unit-tested).
- Only used for types objects whose registrations are versioned: `TypeOverrides` instances backed by the global pg-types module (the standard `Client` path — client.js injects its `TypeOverrides` into every `Result`), and the global module itself. User-supplied `types` objects (per-query `types`, or `TypeOverrides` wrapping custom `userTypes`) keep the existing uncached path, since their `getTypeParser` behavior can change without notice.
- Capped at 1000 shapes per types object (cleared wholesale beyond that, like a generation flip).

Microbenchmark (`new Result` + `addFields` + one `parseRow`, 20 columns, Node 26): 7.3µs → 2.9µs per query, ~2.6× faster result processing on repeated shapes.

## Global setTypeParser coverage

Registrations on the global module — `pg.types.setTypeParser(...)`, `defaults.parseInt8`, or a direct `require('pg-types')` — all mutate the same module object, so its `setTypeParser` is wrapped once (in result.js, at first pg load) to bump the same version counter. The `parseInt8` integration test, which toggles the global parser between same-shape queries, passes. The one remaining bypass is code that saved a reference to the unwrapped `setTypeParser` before pg was first required and calls it after queries have run; a version counter inside pg-types itself would close even that, and I'm happy to send that one-liner to brianc/node-pg-types as a follow-up if you want it.

## Tests

- `packages/pg/test/unit/client/field-metadata-cache-tests.js`: shape reuse, distinct shapes, name/oid/format keying, `setTypeParser` invalidation, per-instance isolation, custom-types bypass (both flavors), and a client-level test that `client.setTypeParser` between two same-shape queries applies to the second.
- New unit case covering the global-module path: `pg.types.setTypeParser` between two same-shape queries applies to the second.
- Full existing unit suite passes (two pre-existing failures on my machine — `getaddrinfo ENOTFOUND localhost` — fail identically on unmodified master).
- Wire-level smoke against a real Postgres 16: repeated shapes, multi-statement queries (two RowDescriptions in one action), `rowMode: 'array'`, named prepared statements, and live `setTypeParser` invalidation in both row modes.

### DIFF
--- a/packages/pg/lib/result.js
+++ b/packages/pg/lib/result.js
@@ -1,8 +1,31 @@
 'use strict'
 
 const types = require('pg-types')
+const TypeOverrides = require('./type-overrides')
 
 const matchRegexp = /^([A-Za-z]+)(?: (\d+))?(?: (\d+))?/
+
+// Real workloads repeat a small set of result-set shapes, so the parser array and
+// prebuilt empty row for a shape are cached per types object instead of rebuilt on
+// every query. Entries are only cached against types objects whose registrations are
+// versioned — TypeOverrides instances (see setTypeParser there) and the pg-types
+// module — so late setTypeParser calls invalidate; arbitrary user-supplied types
+// objects keep the uncached path since their behavior can change without notice.
+const fieldMetadataCache = new WeakMap()
+const FIELD_METADATA_CACHE_LIMIT = 1000
+
+// Registrations on the global module must invalidate cached field metadata no matter
+// how they arrive — pg.types.setTypeParser, defaults.parseInt8, or a direct pg-types
+// require all mutate the same module object — so its setTypeParser is wrapped once
+// to count them.
+if (!types.__setTypeParserWrapped) {
+  const setTypeParser = types.setTypeParser
+  types.setTypeParser = function () {
+    types.__typeParserVersion = (types.__typeParserVersion || 0) + 1
+    return setTypeParser.apply(types, arguments)
+  }
+  types.__setTypeParserWrapped = true
+}
 
 // result object returned from query
 // in the 'end' event and also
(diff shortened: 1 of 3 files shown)


fabiolmorato (approved):

brianc:
Thanks for the PR and sorry for the delay! Been out of town for a while dealing w/ some stuff...  I need to think about this one a bit more - it definitely makes sense in a lot of scenarios.  Would it be possible to run the benchmark suite that @nigrosimone wrote with these changes & see how it impacts?  I have a few very minor concerns like the seemingly reasonable but also seemingly arbitrary 1000 cache limit on number of queries. Overall on the surface it makes total sense. I'll sleep on it a bit and see if there are any weird end-cases throughout the years that could crop back up pop into my mind. FWIW A long time ago I used to build a constructor object with eval for each query which sped up subsequent rows but leaked some memory and had some hard to track down problems and the perf gains were minor. I always wary introducing something that might end up w/ more allocations in large production apps so...I'll take some caution here.

nigrosimone:
@Iaotle Brian talk about this benchmark https://github.com/nigrosimone/postgres-benchmarks you can fork and patch pg with your pr.

Iaotle:
I ran `nigrosimone/postgres-benchmarks` at `9164412f` against this PR (`2b4e8d78`) and its parent (`c9e57617`). I ran the baseline and PR in the same process, alternated their execution order for every query pair, pinned the container to one CPU, and collected 14 paired rounds.

| Workload | Throughput change |
| --- | ---: |
| Sequential LIMIT 1 | +0.7% |
| Sequential LIMIT 100 | +0.6% |
| Sequential LIMIT 500 | +2.6% |
| Pipelined LIMIT 1 | -0.7% |
| Pipelined LIMIT 100 | -1.9% |
| Pipelined LIMIT 500 | +1.4% |

The differences varied between rounds, so the result is effectively neutral across all six workloads. The benchmark selects five columns, which makes the saved `addFields` work small relative to the database, protocol, and row-processing work.

nigrosimone:
> The benchmark selects five columns

5 columns are too few to show such a performance improvement, try with 20

Iaotle:
I ran the requested benchmark against the PR head (`2b4e8d78`) and its parent (`c9e57617`).

I published the reproducible benchmark as a single commit in my fork:

- [method, full results, and run commands](https://github.com/Iaotle/postgres-benchmarks/blob/pr-3753-paired-benchmark/PAIRED_RESULTS_PR_3753.md)
- [paired runner](https://github.com/Iaotle/postgres-benchmarks/blob/pr-3753-paired-benchmark/paired-compare.mjs)
- [exact benchmark commit](https://github.com/Iaotle/postgres-benchmarks/commit/848f634d3a500b52e2bf55e1dd5081a61ab26687)

The parent and PR versions ran in the same Node process, their order alternated each round, PostgreSQL was pinned to CPU 0, and Node was pinned to CPU 1. These are geometric means of paired throughput ratios with 95% confidence intervals:

| Workload | 20 columns | 100 columns |
| --- | ---: | ---: |
| Sequential LIMIT 1 | **+5.6%** (+2.1% to +9.2%) | **+10.3%** (+5.8% to +15.0%) |
| Sequential LIMIT 100 | -1.5% (-7.2% to +4.6%) | +2.2% (-0.2% to +4.8%) |
| Sequential LIMIT 500 | -1.7% (-6.3% to +3.1%) | 0.0% (-2.6% to +2.5%) |
| Pipelined LIMIT 1 x10 | +1.7% (-3.4% to +7.2%) | **-9.4%** (-14.1% to -4.5%) |
| Pipelined LIMIT 100 x10 | +0.3% (-3.2% to +4.0%) | +2.2% (-0.6% to +5.1%) |
| Pipelined LIMIT 500 x10 | +1.9% (-1.7% to +5.6%) | -1.4% (-4.1% to +1.3%) |

An independent 20-pair confirmation of the 100-column `LIMIT 1` cases found **+12.4%** (+9.6% to +15.2%) sequentially and **-9.5%** (-12.2% to -6.7%) with 10 pipelined queries.

I'm not sure about these results because I was getting inconsistent results most of the time in earlier attempts, until this latest paired/pinned run. I discarded those earlier runs. The latest result did reproduce the sequential improvement and the pipelined regression at 100 columns, but I would still treat the numbers as directional given the earlier instability.

PR #3754, opened 2026-08-18
fix: add default error listeners to clients and pools

### LINKED ISSUE #3630: ## add default no-op `error` listener to pool and client

pratik-desgn:
## Summary

- add default no-op `error` listeners to `Pool` and both `Client` implementations
- prevent an emitted error from crashing an otherwise unconfigured client or pool
- add regression coverage for the pure JavaScript client and pool

Fixes #3630

### DIFF
--- a/packages/pg-pool/index.js
+++ b/packages/pg-pool/index.js
@@ -66,6 +66,7 @@ function makeIdleListener(pool, client) {
 class Pool extends EventEmitter {
   constructor(options, Client) {
     super()
+    this.on('error', NOOP)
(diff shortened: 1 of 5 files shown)


charmander:
https://github.com/brianc/node-postgres/pull/3748#issuecomment-5276047341

PR #3756, opened 2026-08-24
fix: Pass `host` to `tls.connect` for certificate validation

### LINKED ISSUE #2263: IP hosts are not validated correctly against certificate altnames

jhnns:
Continues #2273 by @RazerM (with tests, as requested in [this comment](https://github.com/brianc/node-postgres/pull/2273#issuecomment-2759620631)). Fixes #2263.

### The problem

When connecting with `ssl` enabled, pg hands an already-connected socket to `tls.connect()`, so [Node.js validates the server certificate against](https://github.com/nodejs/node/blob/v24.6.0/lib/_tls_wrap.js#L1635-L1640):

```
options.servername || options.host || 'localhost'
```

pg only sets `servername` for hostnames, since SNI must not contain IP addresses ([RFC 6066, section 3](https://datatracker.ietf.org/doc/html/rfc6066#section-3) — the reason for the check introduced in #1890). But `servername` was doing double duty as the name for certificate validation, and `host` was never set. So when connecting to an IP address, **the certificate was silently validated against `'localhost'`** — a certificate for an entirely different server would be accepted (see #2263).

### The fix

Also pass `host` to `tls.connect()`. Node.js then validates the certificate against the IP being connected to (via its IP SANs). Hostname connections are unaffected (`servername` still takes precedence), as are Unix domain sockets (`host` is `undefined` there, as before). Per Node's docs, `host` is [ignored except for certificate validation](https://nodejs.org/docs/latest/api/tls.html#tlsconnectoptions-callback) when a socket is supplied. Both sides are now documented with code comments in `upgradeToSSL`.

### Tests

Added `test/unit/connection/ssl-tests.js`, following the existing unit-test conventions:

1. Connecting to an IP address passes `host` to the secure stream and leaves `servername` unset.
2. Connecting to a hostname passes both `host` and `servername`.
3. An end-to-end test performing a real TLS handshake agar (using the existing `test/tls` certificates), assertingvia a `checkServerIdentity` override that Node.js verifies the certificate against the IP address. Without the fix, it receives `'localhost'` and the test fails.


charmander (approved):

brianc (commented):
This seems entirely reasonable to me & makes sense.  Thanks! sorry for the delay - out of town until sometime mid next-week ❤️

PR #3759, opened 2026-09-01
fix: sync rows query after error

### LINKED ISSUE #3707: Error during a `rows`-limited query permanently wedges the connection — `handleError` sends no Sync

foredstudio:
Fixes #3707.

When a query uses the `rows` option, `_getRows` sends `Flush` instead of pipelining `Sync`. If PostgreSQL returns an `ErrorResponse`, it ignores subsequent messages until a `Sync` arrives, so the client can remain permanently stuck waiting for `ReadyForQuery`.

This sends `Sync` from `handleError` only for rows-mode queries, matching the existing rows-mode handling in `handleCommandComplete` and `handleEmptyQuery`. Normal queries keep their existing behavior.

Tests added for both rows-mode and normal-query error paths.

Validation:
- `node test/unit/client/query-error-sync-tests.js`
- `make test-unit`
- `eslint packages/pg/lib/query.js packages/pg/test/unit/client/query-error-sync-tests.js`


foredstudio:
CI is green across the full matrix. This is ready for review when convenient; the change is limited to sending Sync after rows-mode query errors, with regression coverage for rows-mode and normal-query error paths.

brianc (commented):
Hey thanks for this! I'm out of town for another week so I'll be slow on responding but this looks reasonable. I have a question tho before merging this - is there a way to write an integration test that could also replicate the error?  I assume issuing an invalid query with `rows: 1` on the config could trigger the backend to send an error response and trip the previously offending code path? Unit tests are always nice, but in this case due to the interplay of backend and frontend I'd love to see an integration test covering the behavior. Is that something you'd be willing to include? 🙏

PR #3767, opened 2026-09-05
fix(pg): make prepared-statement caches prototypeless

### LINKED ISSUE #3625: Make `Connection`’s `parsedStatements` a prototypeless object

nirmal-shaji:
## Problem

A prepared statement whose name collides with an `Object.prototype` key — most obviously `constructor`, but also `hasOwnProperty`, `toString`, `__proto__`, etc. — is mishandled. The client believes the statement has already been prepared when it has not, so it skips the `Parse` message and sends a `Bind` for a statement the backend has never seen, producing a protocol error rather than running the query.

## Cause

`Connection` tracks which named statements have been prepared using plain object literals:

```js
this.parsedStatements = {}
this.submittedNamedStatements = {}
```

The native `Client` does the same with `this.namedQueries = {}`. These caches are keyed by user-supplied statement names, and `Query#hasBeenParsed` / `Query#submit` test membership with a simple truthy lookup:

```js
connection.parsedStatements[this.name] || connection.submittedNamedStatements[this.name]
```

Because the objects inherit from `Object.prototype`, `parsedStatements['constructor']` resolves to `Object.prototype.constructor` (truthy) on a brand-new connection, so `Parse` is never sent.

## Fix

Create the three caches with `Object.create(null)` so lookups only ever reflect statements that were actually prepared. This matches the request in #3625 (make `parsedStatements` prototypeless, alongside `namedQueries` on the native client); `submittedNamedStatements` is fixed too since it is consulted in the same `||` expression and shares the flaw. No API surface changes.

## Testing

Added a pure-Node unit test (no live Postgres) at `packages/pg/test/unit/connection/prepared-statement-name-tests.js`, mirroring the existing `test/unit/connection` tests and using the in-memory stream helper. It asserts that:

- a fresh connection reports no prepared statements for `Object.prototype` names;
- a statement named `constructor` sends a `Parse` before its `Bind`;
- `hasBeenParsed` is false for unprepared prototype-named statements.

The test fails on `master` and passes with this change. The full `packages/pg` unit suite (`make test-unit`, run file-by-file with `node`) remains green.

Targeting the default branch (`master`); the issue is tagged toward the `pg@9.0` milestone — happy to retarget if a release branch is preferred.

Fixes #3625


charmander:
https://github.com/brianc/node-postgres/pull/3732#issuecomment-5250525353

PR #3772, opened 2026-09-09
fix(pg): do not treat Sync as connection ending

### LINKED ISSUE #3769: `_ending` is set by every `sync()` and never reset, so connection errors are swallowed for the life of the client (and with `pipeline: true`, a query's promise can never settle)
- **pg:** 8.23.0
- **Runtime:** Cloudflare Workers (`workerd`) via `pg-cloudflare`, `nodejs_compat`
- **PostgreSQL:** Neon, reached through Cloudflare Hyperdrive (transaction-mode pooler)

## Summary

`Connection.prototype.sync()` sets `this._ending = true`, and nothing ever sets it back. Since `sync()` runs after **every** query in the extended protocol, `_ending` is permanently true from the first query onward, for a client that is not ending at all.

`reportStreamError` then silently drops `ECONNRESET` and `EPIPE` for the rest of that connection's life:

```js
// lib/connection.js:54-59
const reportStreamError = function (error) {
  // errors about disconnections should be ignored during disconnect
  if (self._ending && (error.code === 'ECONNRESET' || error.code === 'EPIPE')) {
    return
  }
  self.emit('error', error)
}
```

```js
// lib/connection.js:199-202
sync() {
  this._ending = true
  this._send(syncBuffer)
}
```

The comment says *during disconnect*, and the flag no longer means that. Recovery from a mid-request teardown therefore rests entirely on the socket's `close` -> `end` -> `_errorAllQueries` path.

## Where we hit it

With `pipeline: true` (8.22+) that recovery path races the pipelined bookkeeping, and an in-flight query's promise is sometimes **never settled**. Not rejected, not resolved: the awaiting code waits for ever.

In production this presented as a sign-in that hung indefinitely. The HTTP request never completed, so nothing was ever logged, while `pg_stat_activity` on the server showed the connection `idle` in `Client/ClientRead` with its last query answered, for fourteen hours. The server had replied; the reply never reached the promise.

## Reproduction

A Worker opening a fresh `Client` per round against a real Hyperdrive origin, running a sign-in-shaped chain of several sequential queries **with a transaction interleaved on the same client** (`begin` / `set_config` / ... / `commit`), 400 rounds per variant:

| variant | hung for ever |
|---|---|
| `pipeline: true` + interleaved transaction | **2 / 400** |
| `pipeline: false` + interleaved transaction | 0 / 400 |
| `pipeline: true`, no transaction | 0 / 400 |

Three things are needed together: pipelining on, more than one statement genuinely in flight (which the interleaved transaction guarantees), and a **real** origin teardown. A local proxy issuing RST/FIN rejected cleanly every time; only the real pooler, with its own connection recycling, produced the race at scale.

In both hangs the client-side state was:

```
events:         ["client.error:Connection terminated unexpectedly", "client.end"]
readyForQuery:  false
_activeQuery:   still set
_ended:         false
```

So the `end` handler had run while `_activeQuery` was still set: the queue never reached `_errorAllQueries` for that query.

## Why we think the two are the same story

`_ending` being permanently true means the ordinary error road is closed before the socket-close road is even reached, so everything depends on a path that, on this runtime, is delivered asynchronously by `cloudflare:sockets`' `.closed` and races `_pulsePipelinedQueryQueue`. We did not pin the lost rejection to a single line inside that race, and we would rather say so than guess. The `_ending` behaviour above, however, is plain in the source and looks wrong on its own terms regardless of runtime.

Possibly the same family as #3707, where a missing Sync leaves `ReadyForQuery` unarriving and wedges the connection.

## What we did

Turned `pipeline` off. The same sign-in step that never returned now answers in 88-115 ms, and the hang has not recurred.

Happy to run further instrumented variants of the probe if that would help; the environment is reproducible on our side.

devtechedge:
## Summary

Fixes #3769.

`Connection.sync()` was setting `_ending = true` on every extended-query Sync. That flag exists so `reportStreamError` can ignore `ECONNRESET` / `EPIPE` **during disconnect**. Sync is the protocol barrier after Parse/Bind/Execute, not a disconnect, so after the first parameterized query those socket errors were silently dropped for the rest of the connection lifetime.

With `pipeline: true` that interacts badly with unexpected pooler/socket teardowns: the normal error path is closed and recovery depends only on the async `close`/`end` path, which can leave an in-flight query promise unsettled.

## Changes

- Stop setting `_ending` in `sync()`; leave it set only in `end()` (Terminate) and the connect-timeout teardown path that already sets `con._ending = true` before destroying the stream.
- Add unit coverage that Sync leaves `_ending` false and that `ECONNRESET` after Sync still emits `error` (existing disconnect coverage still uses `end()`).

## Test plan

- [x] `node test/unit/connection/error-tests.js` (new cases green)
- [x] Full `packages/pg` unit suite (`find test/unit -name '*-tests.js' | xargs -n1 node`) - 284 pass

### DIFF
--- a/packages/pg/lib/connection.js
+++ b/packages/pg/lib/connection.js
@@ -195,7 +195,8 @@ class Connection extends EventEmitter {
   }
 
   sync() {
-    this._ending = true
+    // Sync is the extended-query protocol barrier, not a disconnect.
+    // Only end()/Terminate (and connect-timeout teardown) should set _ending.
     this._send(syncBuffer)
   }
 
--- a/packages/pg/test/integration/gh-issues/3772-tests.js
+++ b/packages/pg/test/integration/gh-issues/3772-tests.js
@@ -4,6 +4,13 @@ const assert = require('assert')
 const net = require('net')
 const suite = new helper.Suite()
 
+// These tests target the pure-JS connection state machine: `_ending` lives on
+// `Connection` and is only consulted by the JS `reportStreamError`. The native
+// (libpq) client has no such property, so there is nothing to assert here.
+if (helper.args.native) {
+  return
+}
+
 // https://github.com/brianc/node-postgres/issues/3769
 //
 // `Connection.prototype.sync()` used to set `_ending = true`. Sync is the
--- a/packages/pg/test/integration/gh-issues/3772-tests.js
+++ b/packages/pg/test/integration/gh-issues/3772-tests.js
@@ -1,0 +1,154 @@
+'use strict'
+const helper = require('../test-helper')
+const assert = require('assert')
+const net = require('net')
+const suite = new helper.Suite()
+
+// https://github.com/brianc/node-postgres/issues/3769
+//
+// `Connection.prototype.sync()` used to set `_ending = true`. Sync is the
+// extended-query protocol barrier sent after every parse/bind/execute, so the
+// flag was left true for the entire life of a healthy connection. Because
+// `reportStreamError` drops ECONNRESET/EPIPE while `_ending` is set, a genuine
+// mid-connection teardown was silently ignored from the first parameterized
+// query onwards: the socket error never reached the client, and recovery was
+// left to the asynchronous close -> end path.
+//
+// These tests use a real PostgreSQL backend, reached through a local TCP proxy
+// so the connection teardown can be triggered deterministically.
+
+const PG_PORT = Number(helper.config.port) || 5432
+
+// A TCP proxy in front of the real backend. Reports the port it listens on and
+// exposes the sockets so a test can reset the connection like a pooler would.
+function createProxy() {
+  const server = net.createServer((clientSocket) => {
+    const upstream = net.connect(PG_PORT, helper.config.host)
+    clientSocket.pipe(upstream)
+    upstream.pipe(clientSocket)
+    server.lastClientSocket = clientSocket
+    server.sockets = server.sockets || []
+    server.sockets.push(clientSocket, upstream)
+    clientSocket.on('error', () => {})
+    upstream.on('error', () => {})
+  })
+  return new Promise((resolve) => {
+    server.listen(0, '127.0.0.1', () => resolve({ server, port: server.address().port }))
+  })
+}
+
+function closeProxy(server) {
+  // Tear down every socket so the process is free to exit.
+  ;(server.sockets || []).forEach((sock) => {
+    try {
+      sock.destroy()
+    } catch (_) {
+      // the socket may already be torn down; nothing to clean up
+    }
+  })
+  return new Promise((resolve) => server.close(() => resolve()))
+}
+
+function connectThroughProxy(port, options) {
+  const client = new helper.pg.Client({
+    ...helper.config,
+    host: '127.0.0.1',
+    port,
+    ...options,
+  })
+  // A teardown is expected in these tests; collect client errors instead of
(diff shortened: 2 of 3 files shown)


brianc (commented):
Oh yeah I see this fix makes sense. What I want to see tho is an integration tests w/ an actual postgres backend to trigger the error path if possible. Unit tests are a pretty weak gaurentee of correctness compared to integration tests which actually test "This code works now against a backend and it didn't use to before the patch." Would you be able to include an integration test? (under `packages/pg/test/integration/gh-issues/3772-tests.js` if possible)

devtechedge:
Added the integration test you asked for at `packages/pg/test/integration/gh-issues/3772-tests.js`.

It runs both cases against a real Postgres backend:

1. `a real extended-protocol query does not mark the connection as ending` - a parameterized query really does write Sync, so `_ending` has to stay false for the life of a healthy connection.
2. `a mid-query connection reset is reported, not swallowed by Sync` - the client reaches Postgres through a local TCP proxy, the query is in flight, then the proxy socket is reset.

I verified the "it didn't use to" part by swapping only `packages/pg/lib/connection.js` between the patched and pre-patch versions, with everything else untouched:

- case 1: fails before the patch, passes after
- case 2: fails before the patch, passes after

Before the fix, case 2 only surfaces the generic `Connection terminated unexpectedly` from the close path. With the fix the underlying `ECONNRESET` reaches the client.

Local runs:

- `find test/integration/gh-issues -name '*-tests.js' | xargs -n1 node`
- `node test/unit/connection/error-tests.js`
- eslint and prettier clean on the new file

Two other integration files fail on my machine for environment reasons, not from this change: `130-tests.js` needs a `psql` binary on PATH, and the SSL tests need a server started with SSL. Both fail identically without the patch.

Happy to rework the teardown approach if you'd prefer it driven a different way.

devtechedge:
One follow-up on my last comment: the first push of the integration test turned CI red, and I've fixed it.

The integration suite runs twice, once against the JS implementation and once with a `native` argument that swaps in the libpq bindings. The native client has no `connection`, so my state-machine assertions threw a `TypeError` and aborted the run. I've guarded the file the same way `test/integration/client/pipeline-portal-tests.js` does, since the patch is to the JS `Connection` and there is no `_ending` to assert on under libpq.

CI is now green across all 12 checks (Node 16 through 26, PostgreSQL 13 through 18, plus lint).

brianc (approved):
nice! with the integration test I feel 💯 about merging this, thanks. I think its probably time I write an AGENTS.md in the repo to give instructions to folks who are using agents to make sure to include _integration tests_ when possible as unit tests are generally far less valuable.

brianc:
thank you for putting in the work here & the follow up! Appreciated!

devtechedge:
Thank you for merging, Brian!

This PR turned out to be a real learning experience with the extended query protocol.

I went in thinking Sync was just a message boundary, and came out with a proper understanding of why it exists as the barrier after Parse/Bind/Execute, and how `_ending` works with `reportStreamError` to keep intentional disconnects from surfacing as errors.

The `pipeline: true` angle made it more interesting than it first looked, since the dropped `ECONNRESET` only really showed up through pooler teardowns.

Thanks as well for the quick review, and for maintaining a codebase that made tracing this straightforward. Happy to help with anything else that comes up in the repo.

brianc:
❤️ ❤️ ❤️

PR #3776, opened 2026-09-13
fix(pg-protocol): copy the bytes a copyData message keeps out of the parser buffer

nigrosimone:
The parser hands `copyData` messages a `chunk` that is a view into its own buffer (`BufferReader.bytes` uses `slice`). When the next socket chunk arrives and what is left of the previous one fits in front, the parser moves those bytes to the start of the same buffer, so a chunk delivered earlier changes under whoever kept it. A `copyData` of 64 bytes followed by a `commandComplete` cut at its last byte is enough: the chunk then starts with `COPY 1\0`. `pg` itself ignores `copyData`, but a `Submittable` that keeps the chunk, for example to push it into a stream, gets silently corrupted data. The MD5 salt goes through the same method.

The fix returns a copy from `bytes()`.

The added test parses that sequence in two chunks and checks the chunk is intact: it fails on master and passes here.


charmander:
It’s a risky API design, but I think it was intentional that parsed messages can contain buffers that reference the parser’s working buffer. Is anything in pg actually broken as a result?

Then again, with binary mode being unimplemented, since `bytes` is only used by those two small messages, it’s not as if it’s much of a performance benefit.

PR #3777, opened 2026-09-13
fix(pg-protocol): keep the bytes of binary columns instead of decoding them as text

### LINKED ISSUE #3495: pg 8.16.2 in binary mode produces incorrect row values

nigrosimone:
Fixes #3495. With `binary: true` the parser decodes every `DataRow` field as utf8 text, and `Result` turns the string back into a Buffer before the type parser runs. Any byte utf8 cannot carry is replaced on the way, so `SELECT 1000::int4` comes back as 1007 and `200` as 239.

The row description says which columns are binary. The parser now remembers that and hands those fields over as bytes, the text columns as strings like before, so a query without binary columns pays one null check per row.

Two tests: the parser one feeds a binary row description and a row with the bytes of 1000, the pg one runs a binary query with an int4, a float8 and a non-ascii text. Both fail on master and pass here.

Built on #3776, which makes `BufferReader.bytes` return a copy; the first commit is that one.


charmander:
#3734

nigrosimone:
Oook 😅

PR #3778, opened 2026-09-13
fix(pg-native): report a missing row count or command as null

nigrosimone:
pg-native builds `rowCount` with `parseInt(pq.cmdTuples(), 10)` and `command` with `pq.cmdStatus().split(' ')[0]` without checking that libpq gave anything. A command with no row count, like `BEGIN`, comes back with `rowCount: NaN`, and an empty query with `command: ''`. pg reports `null` in both cases, and `rowCount` is typed as `number | null`, so code that checks `rowCount === null` behaves differently on the two clients.

The fix returns `null` when libpq gives an empty string.

The added test runs `BEGIN` and an empty query and checks both fields: it fails on master and passes here.


brianc (approved):
oh nice - I think there's a small argument to be made this is kind of a breaking change? But since the types already agree it should be `null` or `string | null` I'm incined to merge as semver patch as a bug. I'd like @charmander 's thoughts tho juuust in case. 😄

PR #3779, opened 2026-09-13
fix(pg-native): send Buffer parameters as bytea hex so no byte is lost

### LINKED ISSUE #980: Passing a Buffer to the native bindings is not binary safe
The pure-Javascript implementation will accept a Buffer object as a parameter to a query, and this works no matter what the contents of the Buffer. The native bindings, however, appear to expect a C-style string and truncate the data in the Buffer if it contains a zero byte. 

Test case:

```
var pg = require('pg').native
var conString = "postgres:///dev";

pg.connect(conString, function(err, client, done) {
  if(err) {
    return console.error('error fetching client from pool', err);
  }
  const buf1 = new Buffer([1, 2, 3]);
  const buf2 = new Buffer([0, 1, 2]);
  const buf3 = new Buffer([1, 2, 0]);
  const sql = "SELECT $1::bytea AS buf1, $2::bytea AS buf2, $3::bytea AS buf3";
  client.query(sql, [buf1, buf2, buf3], function(err, result) {
    done();
    pg.end();
    if(err) {
      return console.error('error running query', err);
    }
    console.log(result.rows[0].buf1.length, result.rows[0].buf1);
    console.log(result.rows[0].buf2.length, result.rows[0].buf2);
    console.log(result.rows[0].buf3.length, result.rows[0].buf3);
  });
});
```

With the pure-Javascript implementation, this produces the expected output:

```
3 <Buffer 01 02 03>
3 <Buffer 00 01 02>
3 <Buffer 01 02 00>
```

With native bindings, it produces incorrect output:

```
3 <Buffer 01 02 03>
0 <Buffer >
2 <Buffer 01 02>
```

I'm currently working around this problem by wrapping Client.query() and converting any parameters that are Buffers into hex-encoded strings, eg. "\\x000102". Not sure what the proper fix would be.

nigrosimone:
Fixes #980. libpq gets every parameter as a C string, so a Buffer is read as utf8 and cut at its first zero byte: `SELECT $1::bytea` with `dd 56 00 33 00` returns `ef bf bd 56`. pg sends the same Buffer in the binary format and returns it intact.

pg-native now turns a Buffer into the hex form bytea accepts, `\x...`, before handing the values to libpq, at every call that takes parameters: query, execute, the sync forms and the pipelined batch. Sending it in the binary format would need node-libpq to take a format per parameter, which it does not.

The added test sends a Buffer with a zero byte through a query, a prepared statement and the sync forms: it fails on master and passes here.


nigrosimone:
Closing: the hex form is right for a bytea column but wrong for a text one, and the test for #675 inserts a Buffer into a TEXT column and expects the text. Without a binary parameter format from node-libpq the two cannot both work, so the fix belongs in node-libpq.

PR #3780, opened 2026-09-13
fix(pg): expose detail and hint on errors from the native client

nigrosimone:
The native client copies the error fields libpq reports onto the error, renaming them to the names pg uses. Two are missing from that map: `messageDetail` and `messageHint`. So a duplicate key error has `err.detail` on pg and `err.messageDetail` on pg-native, and the same for `hint`.

The fix adds the two names to the map.

The parity test between the two clients now also compares a duplicate key error, which carries a detail, and a misspelt column, which carries a hint: it fails on master and passes here.


brianc (approved):
Ah nice - i wonder if way back in the day i didn't have these fields on the pg-native / libpq error object or something. Maybe it was just an oversight. Either way, looks good to me!

PR #3781, opened 2026-09-13
fix(pg): run a named statement with an empty text more than once

nigrosimone:
`client.query({ text: '', name: 'empty' })` works once and fails from the second run on with `prepared statement "empty" already exists`. On `ParseComplete` the client records the name with its text, `''`, and `hasBeenParsed` then tests that text for truth, so the statement is read as never parsed and is sent to Parse again. The native client keeps its own map of names and reads it the same way.

The fix checks both caches by presence rather than by the truth of the text.

The added test runs a named empty statement twice: it fails on master, on both clients, and passes here.


brianc (approved):
Nice - i never would have thought to even test for this. Looks solid to me. I have a nit though: it looks ilke maybe the comments are claude generated? The way it "talks" is overly clever. I prefer writing comments in a more straight-forward voice.  Maybe "check for undefined, not falsy. support empty query text." or no comment at all?

nigrosimone:
Because it's made with claude. I rewrite them.

charmander (approved):

brianc:
Woohoo thanks!

PR #3782, opened 2026-09-13
ci: fuzz the parser, the query modes and pg-native on a random seed

nigrosimone:
Three differential fuzzers in `fuzz/`, and a CI job that runs a few hundred rounds of each on a seed nobody chose, on every push and pull request. A divergence turns the job red and prints the seed that reproduces it, with the case shrunk to the lines worth pasting into a test.

- `wire.js`: random backend messages written from the protocol description, parsed in one buffer and cut at random points, against what was written
- `modes.js`: `client.query` against the extended protocol forced, a named statement run twice, rowMode array, the binary result format, pipeline mode, pg-cursor and pg-query-stream
- `native.js`: pg against pg-native, as a plain query, a named statement and rowMode array

They run from the command line too:

```bash
yarn fuzz:wire                              # the parser alone, no server needed
yarn fuzz:modes                             # needs a server, PG* variables say where
yarn fuzz:native                            # and pg-native built
node fuzz/modes.js --rounds 500             # longer, the default is 100
node fuzz/modes.js --seed 12345 --rounds 1  # replay what the CI printed
node fuzz/modes.js --keep-going             # do not stop at the first divergence
```

The first runs found #3776, #3777, #3778, #3780 and #3781, each fixed in its own PR with a test that fails on master. This branch carries those five commits so its CI is green; I will rebase it once they are merged. What it cannot fix is #980, a Buffer parameter reaching libpq as a C string, so the native arm draws none.


packages/pg-native/test/empty-query.js
```diff
@@ -13,4 +13,22 @@ describe('empty query', () => {
...
+      assert.strictEqual(res.command, 'BEGIN')
+      assert.strictEqual(res.rowCount, null)
+      client.query('', (err, rows, res) => {
+        assert(!err)
+        assert.strictEqual(res.command, null)
+        assert.strictEqual(res.rowCount, null)
+        client.end(done)
+      })
+    })
+  })
```
brianc: to be very pedantic about things, this is a breaking API change (unfortunately).

PR #3785, opened 2026-09-14
fix(pg-cursor): preserve cursor state after late responses

mwolting:
When a cursor closes before its first backend response arrives, late responses can reset its state and send a second Close/Sync pair. The extra ReadyForQuery can advance the client's query queue too early and corrupt the next query's result.

Preserve closed and failed states when handling late descriptions, and keep closed cursors closed when delivering rows. Fields and the active read callback still get delivered.

Add focused late-message tests and a PostgreSQL regression that closes before responses arrive, checks for a single Sync, and reuses the connection.


brianc (approved):
holy smokes I can only imagine this was difficult to diagnose the underlying issue in whatever production system you were running - corrupting the state of the client & query queue can result in some extremely confusing errors at runtime. Thanks for this! 100 lines of tests and only a few lines of implementation change is generally a nice type of PR to review! 😎

PR #3786, opened 2026-09-20
fix(pg-cursor): settle read() when rowCount exceeds remaining rows

### LINKED ISSUE #2949: cursor.read(rowCount) will hang if rowCount larger than total rows' size
Assume we have a table which contains 100 rows of data. And we query a select statement with cursor given rowCount = 1000:
```
cursor.read(1000, (err, rows) => {
  // do something
});
```
And the callback of read() will never be called.
If we set rowCount <= 100, then everything works as expected. 
But we can not always know how many rows inside a table. So I think a cursor should work fine when the total row count less than cursor rowCount parameter.

dyk1454683243-sudo:
Fixes #2949

## Problem

`cursor.read(rowCount)` never called its callback when `rowCount` was larger than the remaining (or total) rows. A table with 100 rows and `read(1000)` hung instead of returning the leftover rows.

PostgreSQL reports portal exhaustion with `CommandComplete`, not `PortalSuspended`. `pg-cursor` only delivered the current batch from `PortalSuspended` or `ReadyForQuery`. After `CommandComplete` the cursor closed the portal and sent `Sync`, then waited for `ReadyForQuery`. If that message was never dispatched to the cursor, `read()` never settled.

A late `RowDescription` could also reset a busy cursor to idle and start a queued execute, overwriting the in-flight callback. Queued `read()` calls were never drained when the portal was exhausted.

## Fix

In `packages/pg-cursor`:

- Deliver the in-flight `read()` as soon as `CommandComplete` arrives (shorter or empty batch).
- Resolve any queued reads with `[]` once the portal is exhausted.
- Do not start a queued execute from `RowDescription` / `noData` while a read is already in flight.
- Clear the current callback synchronously so a later `ReadyForQuery` cannot deliver the same batch twice.

## Tests

`packages/pg-cursor/test/read-exhaustion.js`:

- Unit: oversized `read()` settles on `CommandComplete` without `ReadyForQuery`.
- Unit: leftover rows after a partial page, then an oversized `read()`.
- Unit: queued reads settle with `[]` after exhaustion.
- Unit: late `RowDescription` does not overwrite an in-flight oversized read.
- Integration: 100 rows, `read(1000)` returns 100, then `[]`.
- Integration: `read(40)` then `read(1000)` returns the remaining 60.

## Claim check

Searched open and recently updated PRs on `brianc/node-postgres` and `dyk1454683243-sudo/node-postgres` for `#2949`, `cursor.read` hang, and `rowCount`. Issue #2949 is still open with no linked closer. No existing PR owns this fix.

## Verification

- `packages/pg-cursor`: 49 passing (including the new cases)
- `packages/pg-query-stream`: 40 passing
- `eslint` on the changed files: clean


dyk1454683243-sudo:
Withdrawing this PR while I clean up a high-volume open-PR backlog. Sorry for the noise — happy to come back later with a focused change if useful.

PR #3789, opened 2026-09-21
feat(pg-connection-string): support the sslpassword parameter

### LINKED ISSUE #3716: Feature request: support `sslpassphrase` parameter in connection string parsing
### Describe the solution you'd like

Parse the `sslpassphrase` query parameter and assign it to `config.ssl.passphrase`, similar to how other SSL-related parameters are handled.

### Example

Given the following connection string:
postgres://user:pass@host:5432/db?sslmode=verify-full&sslrootcert=/path/to/ca.crt&sslcert=/path/to/client.crt&sslkey=/path/to/client.key&sslpassphrase=MySecretPass

I'd be happy to submit a PR for this.

u9822914324-web:
Closes #3716.

## What

Parses `sslpassword` from the connection string and assigns it to `config.ssl.passphrase` — the Node TLS option used to decrypt an encrypted `ssl.key`.

```
postgres://user:pass@host:5432/db?sslcert=/path/client.crt&sslkey=/path/client.key&sslpassword=MySecretPass
```

## On the parameter name

The issue requested `sslpassphrase`, but @charmander pointed out in the thread that the libpq-compatible spelling is [`sslpassword`](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLPASSWORD). This implements `sslpassword`.

## Notes on the implementation

- Unlike `sslcert`, `sslkey` and `sslrootcert`, `sslpassword` is a **literal value rather than a path**, so it is deliberately not read from disk and does not pull in `fs`.
- It is added to the condition that initialises `config.ssl`, so it works on its own rather than throwing on an undefined `config.ssl`. That matches how the other `ssl*` parameters already behave.
- `toClientConfig` passes the whole ssl object through `toConnectionOptions`, so `passphrase` reaches `ClientConfig` without extra plumbing — confirmed by reading that function rather than assuming.
- `SSLConfig` in `index.d.ts` gains `passphrase?: string`.
- README documents the parameter alongside the other `ssl*` entries.

## Testing

Three tests added to `packages/pg-connection-string/test/parse.ts`: the parameter on its own, alongside `sslkey`, and URL-decoding of special characters (the last confirms the existing query parsing decodes the value, so passwords containing `/`, `?`, `&` or `=` survive).

```
83 passing (22ms)
```

Full `pg-connection-string` suite, 0 failing. `prettier --check` passes on `index.js`, `index.d.ts` and `test/parse.ts`.

The package README does not currently satisfy `prettier --check` on `main`, so I left its existing formatting alone rather than reformatting unrelated lines.

## One thing worth flagging

A password in a connection string can end up in logs, shell history, and process listings. If you would like a caution in the README similar to the existing `sslmode` one, I am happy to add it — I did not want to expand the scope unasked.

Prepared with AI assistance (Claude); behaviour verified by the tests above.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


u9822914324-web:
Apologies — my previous comment lost its formatting and came through with the code spans stripped. Reposting it properly.

Added one more commit: `docs/pages/features/ssl.mdx` states that including `sslcert`, `sslkey`, `sslrootcert` or `sslmode` in a connection string replaces an `ssl` object supplied alongside it. Since `sslpassword` now participates in the same condition that initialises `config.ssl`, it has that effect too, so it is listed there as well.

(That file does not satisfy `prettier --check` on `master` either, so I left its existing formatting alone.)

u9822914324-web:
Force-pushed to rebase onto `master`. This branch was accidentally cut from the branch behind #3788, so the PR was also showing that PR's `docs/pages/apis/errors.mdx` changes. Rebased to drop that commit — this PR is now just the `sslpassword` support plus the one-line `ssl.mdx` note, and is independent of #3788.

No change to the code under review.

